PR #5, opened 2019-12-20
changed signatures of our BasicApplication and prototypes.

sveneberth:
Removed the unused agrs and kwargs and added type annotations

### DIFF
--- a/prototypes/basic.py
+++ b/prototypes/basic.py
@@ -10,7 +10,6 @@ class BasicApplication(object):
 	:ivar kindName: Name of the kind of data entities that are managed by the application. \
 	This information is used to bind a specific :class:`server.skeleton.Skeleton`-class to the \
 	application. For more information, refer to the function :func:`_resolveSkel`.
-	:vartype kindName: str
 
 	:ivar render: will be set to the appropriate render instance at runtime
 	"""
@@ -20,7 +19,7 @@ class BasicApplication(object):
 	adminInfo = None
 	accessRights = None
 
-	def __init__(self, moduleName, modulePath, *args, **kwargs):
+	def __init__(self, moduleName: str, modulePath: str):
 		self.moduleName = moduleName
 		self.modulePath = modulePath
 		self.render = None
@@ -32,7 +31,7 @@ class BasicApplication(object):
 				if not rightName in conf["viur.accessRights"]:
 					conf["viur.accessRights"].append(rightName)
 
-	def _resolveSkelCls(self, *args, **kwargs):
+	def _resolveSkelCls(self):
 		"""
 		Retrieve the generally associated :class:`server.skeleton.Skeleton` that is used by
 		the application.
--- a/prototypes/hierarchy.py
+++ b/prototypes/hierarchy.py
@@ -1,10 +1,9 @@
 # -*- coding: utf-8 -*-
-import logging, sys
+import logging
 from datetime import datetime
 from time import time
 
-from viur.core import db, utils, errors, conf, request, securitykey
-from viur.core import forcePost, forceSSL, exposed, internalExposed
+from viur.core import conf, db, errors, exposed, forcePost, forceSSL, internalExposed, request, securitykey, utils
 from viur.core.bones import baseBone, keyBone, numericBone
 from viur.core.prototypes import BasicApplication
 from viur.core.skeleton import Skeleton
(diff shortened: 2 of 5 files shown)


prototypes/basic.py
```diff
@@ -20,7 +19,7 @@ class BasicApplication(object):
 	adminInfo = None
 	accessRights = None
 
-	def __init__(self, moduleName, modulePath, *args, **kwargs):
```
phorward: I'm quite unsure if its good to remove *args and **kwargs here, because enhanced classes may use extra information and this is the fallback for it, serving as a general function signature.

prototypes/basic.py
```diff
@@ -32,7 +31,7 @@ def __init__(self, moduleName, modulePath, *args, **kwargs):
 				if not rightName in conf["viur.accessRights"]:
 					conf["viur.accessRights"].append(rightName)
 
-	def _resolveSkelCls(self, *args, **kwargs):
```
phorward: same here

prototypes/hierarchy.py
```diff
@@ -48,10 +46,10 @@ def adminInfo(self):
 			"icon": "icons/modules/hierarchy.svg"  # Icon for this module
 		}
 
-	def __init__(self, moduleName, modulePath, *args, **kwargs):
```
phorward: same here

prototypes/hierarchy.py
```diff
@@ -48,10 +46,10 @@ def adminInfo(self):
...
-		super(Hierarchy, self).__init__(moduleName, modulePath, *args, **kwargs)
+	def __init__(self, moduleName: str, modulePath: str):
+		super(Hierarchy, self).__init__(moduleName, modulePath)
 
-	def viewSkel(self, *args, **kwargs):
```
phorward: same here

prototypes/hierarchy.py
```diff
@@ -63,9 +61,9 @@ def viewSkel(self, *args, **kwargs):
 		:return: Returns a Skeleton instance for viewing an entry.
 		:rtype: server.skeleton.Skeleton
 		"""
-		return self._resolveSkelCls(*args, **kwargs)()
```
phorward: same here

prototypes/hierarchy.py
```diff
@@ -63,9 +61,9 @@ def viewSkel(self, *args, **kwargs):
...
+		return self._resolveSkelCls()()
 
-	def addSkel(self, *args, **kwargs):
```
phorward: same here

prototypes/hierarchy.py
```diff
@@ -77,9 +75,9 @@ def addSkel(self, *args, **kwargs):
 		:return: Returns a Skeleton instance for adding an entry.
 		:rtype: server.skeleton.Skeleton
 		"""
-		return self._resolveSkelCls(*args, **kwargs)()
```
phorward: same here

prototypes/hierarchy.py
```diff
@@ -77,9 +75,9 @@ def addSkel(self, *args, **kwargs):
...
+		return self._resolveSkelCls()()
 
-	def editSkel(self, *args, **kwargs):
```
phorward: same here

prototypes/hierarchy.py
```diff
@@ -91,7 +89,7 @@ def editSkel(self, *args, **kwargs):
 		:return: Returns a Skeleton instance for editing an entry.
 		:rtype: server.skeleton.Skeleton
 		"""
-		return self._resolveSkelCls(*args, **kwargs)()
```
phorward: same here

prototypes/list.py
```diff
@@ -32,10 +30,10 @@ def adminInfo(self):
 			"icon": "icons/modules/list.svg"  # Icon for this module
 		}
 
-	def __init__(self, moduleName, modulePath, *args, **kwargs):
```
phorward: same here

prototypes/basic.py
phorward: this might also be useful for monkey patching ;-)

sveneberth: IMO it's much clearer to have an signature of a method which is reduced on the essentials. To allow kwargs could assume that something is done with them. But this is not the case, so we should omit them.
If you inherit one of these classes you still can change the signature according to your requirements.

PR #6, opened 2019-12-20
TreePrototype: fixed delete method and added generic methods

sveneberth:


tsteinruecken:
This is now obsolete as we'll continue with our unifiedTree module

PR #9, opened 2020-02-03
Setting version number

phorward:
Hello, please change the version number of viur-core like so.

@XeoN-GHMB and I are focusing the development of viur-vi to work both with ViUR 2 and ViUR 3, therefore a distinction of the server version is necessary in some cases.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -15,17 +15,17 @@
 
    I N F O R M A T I O N    S Y S T E M
 
- ViUR SERVER
- Copyright 2012-2019 by Mausbrand Informationssysteme GmbH
+ ViUR-core 3
+ Copyright 2012-2020 by Mausbrand Informationssysteme GmbH
 
  ViUR is a free software development framework for the Google App Engine™.
- More about ViUR can be found at https://www.viur.is/.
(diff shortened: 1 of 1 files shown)


tsteinruecken:
If we already want to set it we'll have to set it in full. The patch version is never exposed - so you'll have no chance to determine if its a full release or a development snapshot. Instead of setting everything to -99 we may set it to -3, -0, -0? So we'll always expose the last full mayor & minor it's based on an still signal it's not an official release?

phorward:
Hi @tsteinruecken,
I think for debugging and better testing, a "real" version number like 3.0.0 would be adequate. In ViUR-vi, I've added an appendix next to the version number, indicating it as development version. 

You could also think about a further flag (provided via config or so) to indicate if its a development version or not. Its on your's, and admin tools can rely on this and give apropriate warnings in case a development version is active.

In general, I think when there is a ViUR-core with a real version number available in a repo or branch that is explicitly flagged as "under development" or "unstable", this satisfies the condition of "it's a development version".

tsteinruecken:
Closing as outdated. We already advertise 3 as the mayor version

PR #15, opened 2020-02-21
translate bone params category and tooltip in json renderer

sveneberth:
Original PR: https://github.com/viur-framework/server/pull/180


tsteinruecken:
Lets wait with this one. I18N is still work in progess; the idea is that tr() will return an object instead of an translated string - so we'll have to resolve that object instead of calling translate again.

sveneberth:
> Lets wait with this one. I18N is still work in progess; the idea is that tr() will return an object instead of an translated string - so we'll have to resolve that object instead of calling translate again.

@tsteinruecken What about this PR now? How can we translate them?

sveneberth:
Replaced by #202

PR #20, opened 2020-05-13
SkeletonInstance.__iter__() and SkeletonInstance.__getValues__() incl. type annotations

sveneberth:
- Sometimes it's nice to simple access the stored values in a Skeleton, therefor I implemented the `getValues` method.
- I also implemented the `__iter__` with the usually python behavior to yield the keys.
- Finally I added some type annotations

### DIFF
--- a/skeleton.py
+++ b/skeleton.py
@@ -1,15 +1,20 @@
 # -*- coding: utf-8 -*-
 from __future__ import annotations
-from viur.core import db, utils, conf, errors
-from viur.core.bones import baseBone, keyBone, dateBone, selectBone, relationalBone, stringBone
-from viur.core.bones.bone import ReadFromClientError, ReadFromClientErrorSeverity, getSystemInitialized
-from viur.core.tasks import CallableTask, CallableTaskBase, callDeferred
-from collections import OrderedDict
-from time import time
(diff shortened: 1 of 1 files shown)


skeleton.py
```diff
@@ -175,6 +184,9 @@ def setEntity(self, entity: db.Entity):
 		self.accessedValues = {}
 		self.renderAccessedValues = {}
 
+	def getValues(self) -> Dict[str, Any]:
```
tsteinruecken, line 187: This should probably not be included. I'd had this discussion recently. getValues() was a hack to retrieve a stable dictionary of key<->values pairs for later usage (as the contents inside the skel could magically change when iterating over a skellist). This is guaranteed not to happen in Py3, so you should be able to use the skel itself in most cases - which is much faster as this expression forces the deserializion of all bones - even if they would not have been accessed otherwise.

sveneberth: Hmm... I think we should discuss this in the next ViUR round.
I'am using this method often for debugging. If your print/log the _Skeleton_ instance you get only a dictionary of the bone names and bone instances - but with _getValues_ you always get the unserialized values and can straightforward see all values.

PR #21, opened 2020-05-22
Two fixes for currently broken functions

phorward:


prototypes/hierarchy.py
```diff
@@ -157,9 +157,11 @@ def ensureOwnModuleRootNode(self):
 		:returns: The entity of the root-node.
 		:rtype: :class:`server.db.Entity`
 		"""
-		key = "rep_module_repo"
-		kindName = self.viewSkel().kindName + "_rootNode"
-		return db.GetOrInsert(key, kindName=kindName, creationdate=datetime.now(), rootNode=1)
+		return db.GetOrInsert(
+			db.Key(self.viewSkel().kindName + "_rootNode", "rep_module_repo"),
+			creationdate=datetime.now(),
+			rootNode=1
```
sveneberth, line 163: Looks okay for me.
But I think would be more suitable 
```suggestion
			rootNode=True
```

session.py
```diff
@@ -3,7 +3,7 @@
 from viur.core.tasks import PeriodicTask, callDeferred
 from viur.core import db, utils
 from viur.core.config import conf
-import logging
+import logging, pickle
```
sveneberth, line 6: Can you explain why we should use pickle here?
The datastore supports the primitive data types, Key- and Entity-objects, list and dict (key must be a str) as value.
What type of data you like to store which isn't supported?

tsteinruecken:
Pickle was the old method used to serialize the session (before embedded Entries where available). It's not used in Py3 anymore (he'd hit a bug with a plugin monkey-patching session.save which still forced the data to be serialized). The old hierarchy and tree module will also be gone soon.

PR #27, opened 2020-06-08
Check for userpassword authentication on auth method name

phorward:
This piece of code didn't work when a project used a modified userpassword authentication flow. This fixes it.

### DIFF
--- a/modules/user.py
+++ b/modules/user.py
@@ -746,7 +746,7 @@ def createNewUserIfNotExists():
 			and isinstance(userMod, User)
 			and "addSkel" in dir(userMod)
 			and "validAuthenticationMethods" in dir(userMod)  # Its our user module :)
-			and any([x[0] is UserPassword for x in userMod.validAuthenticationMethods])):  # It uses UserPassword login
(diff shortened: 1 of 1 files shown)


modules/user.py
```diff
@@ -746,7 +746,7 @@ def createNewUserIfNotExists():
 			and isinstance(userMod, User)
 			and "addSkel" in dir(userMod)
 			and "validAuthenticationMethods" in dir(userMod)  # Its our user module :)
-			and any([x[0] is UserPassword for x in userMod.validAuthenticationMethods])):  # It uses UserPassword login
+			and any([x[0].getAuthMethodName() == "X-VIUR-AUTH-User-Password" for x in userMod.validAuthenticationMethods])):  # It uses UserPassword login
```
sveneberth, line 749: I think this way would be more clearly and python-conform. Of course, works your implementation as well. 
```suggestion
			and any([issubclass(x[0], UserPassword) for x in userMod.validAuthenticationMethods])):  # It uses UserPassword login
```

PR #28, opened 2020-06-08
added move function for Uniform Tree

XeoN-GHMB:
Added a move function to move a leaf or node to another node.
Optionally its possible to update the sortindex.
The updateParentRepo function updates all children if the parentrepo has changed.
The move function uses the canMove function for access control.


prototypes/uniformtree.py
```diff
@@ -434,6 +469,105 @@ def deleteRecursive(self, nodeKey):
...
+		:returns: The rendered, edited object of the entry.
+
+		:raises: :exc:`server.errors.NotFound`, when no entry with the given *key* was found.
+		:raises: :exc:`server.errors.Unauthorized`, if the current user does not have the required permissions.
+		:raises: :exc:`server.errors.PreconditionFailed`, if the *skey* could not be verified.
+		"""
+		if "skey" in kwargs:
+			skey = kwargs["skey"]
+		else:
+			skey = ""
```
sveneberth, line 497:
```suggestion
		skey = kwargs.get("skey", "")
```
would be smarter ;)

prototypes/uniformtree.py
```diff
@@ -434,6 +469,105 @@ def deleteRecursive(self, nodeKey):
...
+		:param skelType: Defines the type of the entry that should be moved and may either be "node" or "leaf".
+		:type skelType: str
+		:param key: URL-safe key of the item to be moved.
+		:type key: str
+		:param destNode: URL-safe key of the destination node, which must be a node.
+		:type destNode: str
+
+		:returns: The rendered, edited object of the entry.
+
+		:raises: :exc:`server.errors.NotFound`, when no entry with the given *key* was found.
```
sveneberth, line 490: server ? it's the core!

prototypes/uniformtree.py
```diff
@@ -434,6 +469,105 @@ def deleteRecursive(self, nodeKey):
...
+		elif skelType == "leaf" and self.leafSkelCls:
+			skelType = TreeType.Leaf
+		else:
+			raise errors.NotAcceptable()
+		skel = self.addSkel(skelType) #srcSkel
+		parentNodeSkel = self.editSkel(TreeType.Node) #destSkel
+
+
+		if not self.canMove( skelType, key, parentNode ):
+			raise errors.Unauthorized()
```
sveneberth, line 509: I think we should raise here a Forbidden (403).
Unauthorized means you have submitted no or invalid credentials - But if you are logged in and have not the _move-/edit-access_ it's simply just forbidden to you.

tsteinruecken: Valid point. This is badly handled in ViUR. To be Standard-Compliant we'd have to check if getCurrentUser() is None and raise Unauthorized(); else we should use Forbidden(). But this would also inflate the code a lot - and has no real benefit as the client already knows whenever it has a valid session or not. So for now i'd suggest we continue to use simply one (regardless of the session state) and leave this for a later release.

sveneberth: Okay. I'am closing this thread and created an Issue for that: https://github.com/viur-framework/viur-core/issues/29

prototypes/uniformtree.py
```diff
@@ -434,6 +470,82 @@ def deleteRecursive(self, nodeKey):
...
+		:returns: The rendered, edited object of the entry.
+
+		:raises: :exc:`viur.core.errors.NotFound`, when no entry with the given *key* was found.
+		:raises: :exc:`viur.core.errors.Unauthorized`, if the current user does not have the required permissions.
+		:raises: :exc:`viur.core.errors.PreconditionFailed`, if the *skey* could not be verified.
+		"""
+		if skelType == "node":
+			skelType = TreeType.Node
+		elif skelType == "leaf" and self.leafSkelCls:
+			skelType = TreeType.Leaf
```
sveneberth, line 495: Here we are changing the skelType to the Enum-values.

prototypes/uniformtree.py
```diff
@@ -434,6 +470,82 @@ def deleteRecursive(self, nodeKey):
...
+			raise errors.NotAcceptable()
+
+		skel = self.addSkel(skelType)  # srcSkel - the skeleton to be moved
+		parentNodeSkel = self.editSkel(TreeType.Node)  # destSkel - the node it should be moved into
+
+		if not skel.fromDB(key) or not parentNodeSkel.fromDB(parentNode):
+			# Could not find one of the entities
+			raise errors.NotFound()
+
+		if not self.canMove(skelType, skel, parentNodeSkel):
```
sveneberth, line 506: But _canMove_ expects here the skelType as str "node" oder "leaf"

tsteinruecken: I think skelType should also be that enum

sveneberth: Okay, but than we should change the docstring + type annotations of `canMove` to avoid misunderstandings.

phorward:
> I found still a mistake in the implementation.

What mistake? Can you please try to fix it?

sveneberth:
> > I found still a mistake in the implementation.
> 
> What mistake? Can you please try to fix it?

See my comments above: https://github.com/viur-framework/viur-core/pull/28#discussion_r446165381
The passed arguments (int from enum) deviates from the expected type (str "leaf" or "node" - see docstring) and could cause unexpected behaviors.

phorward:
@XeoN-GHMB can you please provide an update for this pull request to speed up the merge? I need this function and a working implementation quickly.

tsteinruecken:
These functions have been added to uniformtree (now only tree) already - i think this one is outdated. Please check whenever this PR contains fixes that the current module is missing. @XeoN-GHMB

phorward:
> These functions have been added to uniformtree (now only tree) already - i think this one is outdated. Please check whenever this PR contains fixes that the current module is missing. @XeoN-GHMB

A diff between the uniformtree.py in this PR's branch and the current master of viur-core shows nearly the same diff from this PR. The move function does not exists in any of master's prototypes. Therefore, I made an updated pull request with the merged current master for a review.

PR #32, opened 2020-06-24
Fixed calling deferred tasks inside transactions

tsteinruecken:


sveneberth:
too fast ;)

In l.206 `db.Key` is also missing.

PR #34, opened 2020-06-26
fixed still more missing db.Key and db.Entity calls and caching

sveneberth:
- fixed still more missing db.Key and db.Entity calls on several placed
- fixed (and enabled) caching

### DIFF
--- a/cache.py
+++ b/cache.py
@@ -1,11 +1,12 @@
 # -*- coding: utf-8 -*-
-from viur.core import db, utils, tasks
-from viur.core.config import conf
-from hashlib import sha512
-from datetime import datetime, timedelta
 import logging
+from datetime import datetime, timedelta
 from functools import wraps
-from viur.core.utils import currentRequest, currentLanguage
+from hashlib import sha512
+
+from viur.core import db, tasks, utils
+from viur.core.config import conf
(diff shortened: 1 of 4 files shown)


cache.py
```diff
@@ -87,9 +88,7 @@ def keyFromArgs(f, userSensitive, languageSensitive, evaluatedArgs, path, args,
 	try:
 		appVersion = currentRequest.get().request.environ["CURRENT_VERSION_ID"].split('.')[0]
 	except:
-		appVersion = ""
-		# FIXME
-		#logging.error("Could not determine the current application id! Caching might produce unexpected results!")
+		appVersion = utils.projectID
```
tsteinruecken, line 91: This fix is invalid. appVersion is supposed to contain the version (like ts-dev) - not the projectID. Reason is to prevent leaking from development versions into the live site

phorward: what is your proposal then to fix this issue? `utils.projectID` looks for me as a fallback.

sveneberth: When we cannot determine the version it doesn't matter which constant we would use as fallback.
Therefore I let the empty string in #54. Maybe we will find a better way to get the version or have to differentiate btw. local devserver and cloud ...

sveneberth:
I replaced this PR with #54 #55

PR #35, opened 2020-06-26
some improvements/fixes to ensure expected behavior of tasks

sveneberth:
In ViUR 2.x is was not possible to have multiple classes with the same named method, which should run periodic, in the same module. Only the first one was executed, because they all had the same periodicTaskName.

Since python 3.3 it's possible to use `__qualname__` which also includes the class(es) where a bounded method belongs to (or only the name if it's an unbounded function). Therewith it's possible to run the same named method in multiple classes periodic.

In googles tasks v2 there is no longer a Exception like `PermanentTaskFailure`. Therefor I implemented this class to make the `noRetry`-decorator again functional.

[Here](https://github.com/viur-framework/viur-core/blob/70a6f3fbece537ad522d7c554354769df1c4d1b2/tasks.py#L98) We would skip methods with an leading underscore from periodic execution, but never get an error. I added an Exception in the according decorator.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -1,16 +1,17 @@
 # -*- coding: utf-8 -*-
-from datetime import datetime, timedelta
-from viur.core.update import checkUpdate
-from viur.core.config import conf
-from viur.core import errors, request, utils
-from viur.core import db
-from functools import wraps
 import json
 import logging
-import os, sys
+import os
+import sys
+from datetime import datetime, timedelta
+from functools import wraps
+from typing import Callable, Dict
+
 from google.cloud import tasks_v2
 from google.protobuf import timestamp_pb2
-from typing import Dict, List, Callable
+
+from viur.core import db, errors, utils
+from viur.core.config import conf
 from viur.core.utils import currentRequest, currentSession
 
 _gaeApp = os.environ.get("GAE_APPLICATION")
@@ -34,6 +35,10 @@ _deferedTasks = {}
 _startupTasks = []
 _periodicTaskID = 1  # Used to determine bound functions
 _appengineServiceIPs = {"10.0.0.1", "0.1.0.1", "0.1.0.2"}
+
+
+class PermanentTaskFailure(Exception):
+	"""Indicates that a task failed, and will never succeed."""
 
 
 class CallableTaskBase:
@@ -203,8 +208,7 @@ class TaskHandler:
 		for task, interval in _periodicTasks[cronName].items():  # Call all periodic tasks bound to that queue
 			periodicTaskName = "%s_%s" % (cronName, task.periodicTaskName)
 			if interval:  # Ensure this task doesn't get called to often
-				lastCall = db.Get(("viur-task-interval", periodicTaskName))
-				logging.error("Interval %s" % interval)
+				lastCall = db.Get(db.Key("viur-task-interval", periodicTaskName))
 				if lastCall and datetime.now() - lastCall["date"] < timedelta(minutes=interval):
 					logging.error(datetime.now())
 					logging.error(lastCall["date"])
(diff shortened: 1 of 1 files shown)


sveneberth:
An example of what I mean:

_myModule.py_

```python
class A:
	@PeriodicTask()
	def run(self):
		pass

class B:
	@PeriodicTask()
	def run(self):
		pass

class C:
	@PeriodicTask()
	def run(self):
		pass
```

In ViUR 2.x only one of them ran, because the _periodicTaskName_ is for all of them the same: `myModule_run`

Now it will be: `myModule.A.run`, `myModule.B.run`, `myModule.C.run` - And all will run

tsteinruecken:
Looks mostly okay,  but using qualname has unforeseen consequences: You cannot override any periodic task anymore. You will end up with two tasks (one from the base module like tree.task1 and one for your new modules mymodule.task1) - and booth will be called. That's probably also not what we want.

sveneberth:
> You cannot override any periodic task anymore. 

This is a valid point. But it's currently (and in server 2.x) also not possible.

**Case 1:** Super- and Sub-class are in two different modules:

_basicModules.py_
```python
class A:
	@PeriodicTask()
	def run(self):
		pass
```
_myModules.py_
```python
from basicModules import A
class B(A):
	@PeriodicTask()
	def run(self):
		pass
```

Task name of method in super-class would be _basicModules_run_
Task name of method in sub-class would be _myModules_run_
Different modules, same method-name -> different PeriodicTaskName - Both will run.
Without monkey-patching you are not able to stop execution _basicModules_run_


**Case 2:** Super- and Sub-class are in the same module:
_myModules.py_
```python
class A:
	@PeriodicTask()
	def run(self):
		pass

class B(A):
	@PeriodicTask()
	def run(self):
		pass
```

Task name of method in super-class would be _myModules_run_
Task name of method in sub-class would be _myModules_run_
Same modules, same method-name -> Same PeriodicTaskName 
1. The first defined and found task (with this name) will run (in our case `A.run`)
2. date of `db.Entity("viur-task-interval", "myModules_run")` will be set to now.
3. Found  `B.run`, get `db.Entity("viur-task-interval", "myModules_run")`, detect as recently run and skip

sveneberth:
To allow overwriting we need another, better solution. `__name__` and `__qualname__` only are not enough for this.

### Some proposals:
Introduce parameter `peridicTaskName` for our `task.PeriodicTask`-decorator: If this is set we will use this as name for periodic task.
**The advantages:** We can give multiple same-named methods in different class, in the same module an individual name - It doesn't matter if we use `__name__` and `__qualname__`, we have a own identifier.
**The disadvantage:** We will not forbid the super-class method to execute. Of course we can rename the sub-class method to the same PeriodicTaskName of super-method - even when they are in different modules - but we cannot be sure or have impact on the order of execution (first found will run, the second one will skip).

___

Introduce blacklist config to block execution of PeriodicTasks.
There you can list the PeriodicTaskName of the super-class and it will not run.
But, this rather a ugly workaround...

___

Introduce parameter `overwrite` for our `task.PeriodicTask`-decorator: If this is argument is set (with the method of our super-class) we will add the decorated method as usually in our `_periodicTasks`-dict and remove the overwrite-method from `_periodicTasks`.
This is btw my favorite. The implementation is straightforward:
```python
def PeriodicTask(interval=0, cronName="default", overwrite=None):
	def mkDecorator(fn):
		if overwrite is not None:
			del overwrite.periodicTaskID
			del overwrite.periodicTaskName
			del _periodicTasks[cronName][overwrite]
		# [...] (actually code here)
		return fn
	return mkDecorator
```

tsteinruecken:
I could not modify this pr as its from your private repository. Iv'e created a replacement pr for this to get the main issues fixed quickly, lets address the issue with __qualname__ on the next ViUR meeting.

sveneberth:
Replaced by #46 #56 #59

Issue #7, opened 2020-01-15
Standardize several notations


phorward:
There are several places in ViUR where the same functionality or behavior has different notations, but conceptually means the same. This issue should be first used as a collector for these notations, which finally should be renamed or equalized to the same notation when version 3.0 is finalized, to avoid confusions especially for newcomers.

# limit, amount

## Problem
[mergeExternalFilter()](https://github.com/viur-framework/viur-core/blob/master/db.py#L819) uses *amount* as GET-parameter, but sets the limit of the query. Functions like [run()](https://github.com/viur-framework/viur-core/blob/master/db.py#L1414) and [fetch()](https://github.com/viur-framework/viur-core/blob/master/db.py#L1516) are using *limit* instead.

## Proposal
Use `limit` everywhere.

# ~~getSkel(), getEntry()~~

## Problem
~~In the HTML-renderer, [getSkel()](https://github.com/viur-framework/viur-core/blob/master/render/html/env/viur.py#L324) returns the structure, and [getEntry()](https://github.com/viur-framework/viur-core/blob/master/render/html/env/viur.py#L128) returns and entry (which normally is a skel!). There is also a getSkel() function in db.Query which returns a Skeleton, and not the structure.~~

## Proposal
- ~~Rename `getSkel()` to `getStructure()`~~
- ~~Rename `getEntry()` to `getSkel()`~~

Edit: This has been fixed with #48 
~~
# ~~add, edit, delete => onItemAdded, onItemEdited, onItemDeleted => addItemSuccess(), editItemSuccess(), deleteSuccess()~~

## ~~Problem~~

~~Why can't we call them just "edit", "add" and "delete" same as the actions they do?~~

## ~~Proposal~~

~~My proposal is to rename this~~

    ~~edit => self.onItemEdited(skel) => self.render.editItemSuccess(skel)~~
    ~~add => self.onItemAdded(skel) => self.render.addItemSuccess(skel)~~
    ~~delete => self.onItemDeleted(skel) => self.render.deleteSuccess(skel)~~

~~into this:~~

    ~~edit => self.onEdit(skel) => self.render.editSuccess(skel)~~
    ~~add => self.onAdd(skel) => self.render.addSuccess(skel)~~
   ~~ delete => self.onDelete(skel) => self.render.deleteSuccess(skel)~~

Edit: This was fixed by #39 

# ~~Variants of relationalBone have misleading names~~

## ~~Problem~~
[...]

PR #39, opened 2020-07-06
Standardize several notations, part 1:  Renaming add/edit/delete-events and rendering functions

phorward:
This is one pull request for resolving the first issue in #7. I would like to integrate these changes stepwise.
I also started to integrate a simple replacement script that will be extended later on.
Please integrate it ASAP, so that further development can be performed, while still using the latest ViUR 3 master in a current customer project.


tools/viur-2to3.py
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/python3
```
tsteinruecken, line 1: This should not be part of the core (there is no need for it to be deployed). Also it's probably better to extend the existing 2to3 tool: http://python3porting.com/fixers.html

phorward: Tools should be put into one place, for a fast access and central maintenance. If you want to use 2to3, feel free to integrate the changes there, I don't have time for this right now.

sveneberth: I agree with both of you.
Generally I would prefer to have this (and other helping stuff like examples, unittests and co.) in the same repo.
But the core will not be build or shipped as pip-package, it's integrated as submodule. Therefore it's nice to keep it small as possible.

Maybe it's better placed in our [viur-tools repo](https://github.com/viur-framework/viur-tools).

phorward: OK well, I don't want to have any more discussion on such little things here anymore, I have to get things done.
I'll gonna put the file into the viur-tools repo and I'm done with it.

phorward:
OK the script is removed now, please continue with the reviewing process.

PR #41, opened 2020-07-06
Added a lockEmpty property to UniqueValue

tsteinruecken:
This allows the use of unique on properties that are not allways set (like uid in userSkel).
Currently, there can only be one user account that does not use oauth to authenicate as it locked the empty string

### DIFF
--- a/bones/bone.py
+++ b/bones/bone.py
@@ -47,6 +47,7 @@ class UniqueLockMethod(Enum):
 @dataclass
 class UniqueValue:
 	method: UniqueLockMethod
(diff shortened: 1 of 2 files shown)


bones/bone.py
```diff
@@ -47,6 +47,7 @@ class UniqueLockMethod(Enum):
 @dataclass
 class UniqueValue:
 	method: UniqueLockMethod
+	lockEmpty: bool
```
phorward, line 50: Please add a comment here how this variable is used.

PR #54, opened 2020-07-09
fixed request/site caching

sveneberth:
Replace #34

### DIFF
--- a/cache.py
+++ b/cache.py
@@ -112,7 +112,7 @@ def wrapCallable(f, urls: List[str], userSensitive: int, languageSensitive: bool
 	"""
 
 	@wraps(f)
-	def wrapF(self, *args, **kwargs) -> str:
+	def wrapF(self, *args, **kwargs) -> Union[str, bytes]:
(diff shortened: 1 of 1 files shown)


cache.py
```diff
@@ -87,73 +89,73 @@ def keyFromArgs(f, userSensitive, languageSensitive, evaluatedArgs, path, args,
...
+def wrapCallable(f, urls: List[str], userSensitive: int, languageSensitive: bool,
+				 evaluatedArgs: List[str], maxCacheTime: int):
 	"""
 		Does the actual work of wrapping a callable.
 		Use the decorator enableCache instead of calling this directly.
 	"""
 
 	@wraps(f)
-	def wrapF(self, *args, **kwargs):
+	def wrapF(self, *args, **kwargs) -> str:
```
tsteinruecken, line 115: This should be Union[str,byte] - it's possible to cache images/pdfs (that's why the content-type is also restored)

sveneberth: I think you mean `bytes` - I've corrected it :)

PR #55, opened 2020-07-09
fixed some problems with tasks and improved some loggings

sveneberth:
Fixed db.Key bug in l.228

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -331,7 +331,7 @@ def callDeferred(func):
 			if req:
 				req.pendingTasks.append(task)  # < This property will be only exist on development server!
 			else:
-				# Warmup request or something - we have to call it now as we can't deffer it :/
+				# Warmup request or something - we have to call it now as we can't defer it :/
 				task()
 
 			return  # Ensure no result gets passed back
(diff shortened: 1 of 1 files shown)


tasks.py
```diff
@@ -178,7 +179,7 @@ def deferred(self, *args, **kwargs):
 				raise errors.RequestTimeout()  # Task-API should retry
 		elif cmd == "unb":
 			if not funcPath in _deferedTasks:
-				logging.error("Ive missed a defered task! %s(%s,%s)" % (funcPath, str(args), str(kwargs)))
+				logging.error("Ive missed a deferred task! %s(%s,%s)", funcPath, args, kwargs)
```
phorward, line 182: Who is Ive? ;-)

tasks.py
```diff
@@ -331,7 +331,7 @@ def mkDefered(func, self=__undefinedFlag_, *args, **kwargs):
 			if req:
 				req.pendingTasks.append(task)  # < This property will be only exist on development server!
 			else:
-				# Warmup request or something - we have to call it now as we can't deferr it :/
+				# Warmup request or something - we have to call it now as we can't deffer it :/
```
phorward, line 334: defer => https://dict.leo.org/german-english/defer

tasks.py
sveneberth: ?

sveneberth: Idk. Maybe a friend of @tsteinruecken :)

PR #51, opened 2020-07-09
Fixed / unified cursors

tsteinruecken:

PR #57, opened 2020-07-09
Removed obsolte sendEmailToAdmins und _GAE_SendEmail functions

tsteinruecken:


dbtransfer.py
```diff
@@ -417,11 +417,8 @@ def exportItems(module, target, importKey, startCursor, endCursor):
 								headers={'Content-Type': 'application/x-www-form-urlencoded'})
 
 	if startCursor == endCursor:
```
phorward, line 419: Please check according to #51 if this test can fail now in ViUR3!

tsteinruecken:
Regarding sendEmailToAdmins: I think the best way is to utilize Stackdriver - configure an email/sms alert when something with critical is logged. This will then also work if no email service is configured.

The Cursor-Thing is a different issue :)

Regarding #24 please see comments there

sveneberth:
> Regarding #24 please see comments there

Which comments?

tsteinruecken:
Superseded by #195 and #200

PR #59, opened 2020-07-09
implemented parameter "override" for PeriodicTask

sveneberth:
This PR replace #35 to allow multiple classes and overwriting inheritance

Usage:
```python

class A(List):
	@tasks.PeriodicTask(60)
	def run(self):
		pass


class B(A):
	@tasks.PeriodicTask(60, override=A.run)
	def run(self):
		pass
```


tsteinruecken:
I've reworked the task-module a bit and added support for removing already queued tasks. See https://github.com/viur-framework/viur-core/pull/60

tsteinruecken:
This one has been superseeded.

PR #60, opened 2020-07-09
Added support for multiple tasks with the same name in the same file …

tsteinruecken:
…and support for unqueueing inherited tasks


tasks.py
```diff
@@ -429,18 +436,15 @@ def PeriodicTask(interval=0, cronName="default"):
...
 			raise RuntimeError("Periodic called methods cannot start with an underscore! "
 							   f"Please rename {fn.__name__!r}")
-		global _periodicTasks, _periodicTaskID
 		if not cronName in _periodicTasks:
 			_periodicTasks[cronName] = {}
 		_periodicTasks[cronName][fn] = interval
-		fn.periodicTaskID = _periodicTaskID
-		fn.periodicTaskName = "%s_%s" % (fn.__module__, fn.__name__)
-		_periodicTaskID += 1
+		fn.periodicTaskName = ("%s_%s" % (fn.__module__, fn.__qualname__)).replace(".", "_").lower()
```
sveneberth, line 447: I thought we've agreed to use format-strings :)
```suggestion
		fn.periodicTaskName = f"{fn.__module__}_{fn.__qualname__}".replace(".", "_").lower()
```

tsteinruecken: Nope

tsteinruecken:
This will not work. Assume the List-Prototype defines such an PeriodicTask. You will probably inherit that more than once. Then it's better to have one place (your main.py) to look wherever an overridden task has been canceled or not.

sveneberth:
> This will not work. Assume the List-Prototype defines such an PeriodicTask. You will probably inherit that more than once. Then it's better to have one place (your main.py) to look wherever an overridden task has been canceled or not.

Okay, you're right. Could cause some trouble.

tasks.py
sveneberth: It would be faster, but okay - it will works this way as well...

PR #63, opened 2020-07-13
Make unique session cookies

phorward:
This is a proposal to let ViUR/core save unique session cookies.

During testing, I recognized that I'm unable do a separate login in two instances running on the same machine, but different ports. This fix makes each session unique to the project id.

### DIFF
--- a/session.py
+++ b/session.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-import logging, os
+import logging
 from hmac import compare_digest
 from time import time
 
--- a/session.py
+++ b/session.py
@@ -34,10 +34,10 @@ class GaeSession:
 	def __init__(self):
 		super().__init__()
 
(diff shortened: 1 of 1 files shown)


session.py
```diff
@@ -25,15 +26,18 @@
...
-	sslCookieName = "viurSSLCookie"
+	"""Store Sessions inside the Big Table/Memcache"""
 	kindName = "viur-session"
 	sameSite = "lax"
 
-	"""Store Sessions inside the Big Table/Memcache"""
+	def __init__(self):
+		super().__init__()
+
+		ident = os.environ.get("GOOGLE_CLOUD_PROJECT", "")
```
tsteinruecken, line 37: Should use projectID from core.utils here, otherwise okay for me

phorward: you're right, I've just fixed it, using `utils.projectID`, and an underscore before.

PR #65, opened 2020-07-13
Fixing _tasks/list

phorward:
This simply fixes `_tasks/list`, but not well.
We just fake the skeleton system here and ignore it in the renderer... both bad solutions.

Any further suggestions on it?

### DIFF
--- a/render/json/default.py
+++ b/render/json/default.py
@@ -223,7 +223,9 @@ class DefaultRender(object):
 		if skellist:
 			for skel in skellist:
 				skels.append(self.renderSkelValues(skel))
-			res["structure"] = self.renderSkelStructure(skellist.baseSkel)
+
+			if skellist.baseSkel:
+				res["structure"] = self.renderSkelStructure(skellist.baseSkel)
 		else:
 			res["structure"] = None
 		res["skellist"] = skels
--- a/render/json/default.py
+++ b/render/json/default.py
@@ -224,8 +224,7 @@ class DefaultRender(object):
 			for skel in skellist:
 				skels.append(self.renderSkelValues(skel))
 
-			if skellist.baseSkel:
-				res["structure"] = self.renderSkelStructure(skellist.baseSkel)
+			res["structure"] = self.renderSkelStructure(skellist.baseSkel)
 		else:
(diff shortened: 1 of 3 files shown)


render/json/default.py
```diff
@@ -223,7 +223,9 @@ def list(self, skellist, action="list", params=None, **kwargs):
 		if skellist:
 			for skel in skellist:
 				skels.append(self.renderSkelValues(skel))
-			res["structure"] = self.renderSkelStructure(skellist.baseSkel)
+
+			if skellist.baseSkel:
+				res["structure"] = self.renderSkelStructure(skellist.baseSkel)
```
phorward, line 228: This is so ugly, I hate it... in any way, there must be an else case setting it to None...

tasks.py
```diff
@@ -251,15 +251,15 @@ def list(self, *args, **kwargs):
...
-		class extList(list):
-			pass
-
-		res = extList(
-			[{"key": x.key, "name": str(x.name), "descr": str(x.descr)} for x in _callableTasks.values() if
-			 x().canCall()])
-		res.cursor = None
-		res.baseSkel = {}
-		return (self.render.list(res))
+		tasks = db.SkelListRef(None)
```
tsteinruecken, line 254: Just replace this with `tasks = db.SkelListRef({)`  - then there is no need to mess with the render :)

xnopasaranx: I tried this solution but it causes another error, because task does not have a skelStructure. What can we do to mitigate this? I causes an error in the json renderer, so we are back to phorwards issue

tsteinruecken: Hm - can you provide a traceback? I've tested it locally and that seemed to fix the issue.

phorward: I can confirm @tsteinruecken's answer, it now works. 

Anyway, I have a proposal for another improvement made in just another commit 4cc94a29dd2fe018a786b27d31e1a266f622de9d. It makes SkelList's baseSkel default to an empty dict, if not provided.

@xnopasaranx can you please provide more details on your problem?

sveneberth: Works for me as well :)

PR #68, opened 2020-07-14
list has no getCursor() method

xnopasaranx:
fixed to provide this method and return None


xnopasaranx:
I just noticed, that there already is a PR dealing with this issue... my solution is probably a dirty hack and can be discarded.

PR #70, opened 2020-07-14
Fixed db.Query.clone()

tsteinruecken:
Needed for fulltext search


sveneberth:
@tsteinruecken  Why should we not clone the attribute `lastCursor`?

tsteinruecken:
This would be unnecessary effort. The reason to clone a query is to run it independently of the main query. This will always (re-)set lastcursor.

PR #72, opened 2020-07-20
Fixing performMagic on userBone

phorward:


tsteinruecken:
The fromClient function is garbarge here. setBoneValue is the new and recommended way - no fiddeling with the key here. I'll update this tomorrow.

phorward:
@tsteinruecken I merged the current master into this PR, I need a working version of this.

tsteinruecken:
I've pushed a new PR (#85 ) with updates to this (cannot update in your fork)

PR #73, opened 2020-07-20
Fixing subSkel behavior to be ViUR2 compliant

phorward:
- the subskel named "*" was always integrated in the past
- it was possible to create subskels without the "key" bone
- IMHO "creationdate" and "changedate" should also be taken always


phorward:
Ok I removed "creationdate" and "changedate" again, rest remains. Please use "squash & merge" to minimize number of commits.

PR #74, opened 2020-07-20
Fix/user login

tsteinruecken:
Revert a previous change to the login functionality which was broken.
Split render.login and render.loginChoices (the first one renders the form, the latter the choices how to login)
Removed now unsupported _() translation from server-templates (breaking the login form)


render/html/user.py
```diff
@@ -22,12 +22,18 @@ def login_disabled(self, authMethods, tpl=None, **kwargs):
...
-	def login(self, authMethods, tpl=None,  **kwargs):
+	def login(self, skel, tpl=None, **kwargs):
 		if "loginTemplate" in dir(self.parent):
 			tpl = tpl or self.parent.loginTemplate
 		else:
 			tpl = tpl or self.loginTemplate
+		return self.add(skel, tpl=tpl)
 
+	def loginChoices(self, authMethods, tpl=None, **kwargs):
+		if "loginTemplate" in dir(self.parent):
```
phorward, line 33: Is it correct to use the same template as for login here?

tsteinruecken: That would be the "backward compatible" way. Decide inside the template. If we want to change this we'll have to decide which has to be renamed (my suggestion: loginChoicesTemplate)

phorward: I think directly loginChoicesTemplate would be more straightforward.

PR #81, opened 2020-07-23
Fixed isInvalid/isEmpty logic in str and numericBone

tsteinruecken:


bones/numericBone.py
```diff
@@ -55,12 +55,13 @@ def singleValueFromClient(self, value, skel, name, origData):
 				value = int(rawValue)
 			else:
 				return self.getDefaultValue(), [ReadFromClientError(ReadFromClientErrorSeverity.Invalid, name, "Invalid Value")]
+		if not value:
```
sveneberth, line 58: This will also disallow `0` (zero) as value.

tsteinruecken: That's the expected behavior. Zero is the numerical equivalent of the empty string. If the field is required, you'll have to provide a non-zero value.

sveneberth: What? 
With this kind of validation we're will not able to store in a tournament software the scores of a game which ends 2:0 (except we remove the required flag, which is stupid for our data completeness). And we are not able to start a counter from zero (except we use not the validation/fromClient). In a online-shop-form you're not able to select "0" as amount for e.g. extra screws. In a accounting software you're not able to set the annual profit to 0 €.
Sorry, but this is not realistic and I see there no reason why we should handle zero as empty string, it's a valid and normal digit in this context! In a numeric-input you can submit a empty string as well - which will be of course invalid for a required-bone.

tsteinruecken: There where other objections as well. We'll defer this to the next ViUR Meeting. I think there is a issue with the meaning of required. It does not mean "enforce there is a value in the skeleton/datastore" - it now always is. For numeric-bones zero is the default (unless overwritten) and it will serve the add form prefilled with it. Maybe required is better understood as "filled" or "non-empty". Maybe we add an option of what value is treated as empty - so zero is valid and -999 isn't?

bones/bone.py
```diff
@@ -119,6 +119,27 @@ def setSystemInitialized(self):
...
+			This usually is the empty string, 0 or False.
+
+			Warning: isEmpty takes precedence over isInvalid! The empty value is always valid - unless the bone
+				is required. But even then the empty value will be reflected back to the client.
+
+			Warning: rawValue might be the string/object received from the user (untrusted input!) or the value
+				returned by get
+
+		"""
+		return not bool(rawVale)
```
sveneberth, line 141: could you please fix the typo?
```suggestion
	def isEmpty(self, rawVaule: Any) -> bool:
		"""
			Check if the given single value represents the "empty" value.
			This usually is the empty string, 0 or False.

			Warning: isEmpty takes precedence over isInvalid! The empty value is always valid - unless the bone
				is required. But even then the empty value will be reflected back to the client.

			Warning: rawValue might be the string/object received from the user (untrusted input!) or the value
				returned by get

		"""
		return not bool(rawValue)
```

bones/booleanBone.py
```diff
@@ -23,6 +23,14 @@ def singleValueFromClient(self, value, skel, name, origData):
 		else:
 			return False, None
 
+	def getEmptyValue(self):
+		return False
+
+	def isEmpty(self, rawVale: Any):
+		if rawVale is self.getEmptyValue():
+			return True
+		return bool(rawVale)
```
sveneberth, line 32: here as well
```suggestion
	def isEmpty(self, rawValue: Any):
		if rawVale is self.getEmptyValue():
			return True
		return bool(rawValue)
```

sveneberth, line 32: Is `return bool(rawValue)` correct?
If _rawValue_ is _True_ it would return `True` which means the field is empty.

bones/dateBone.py
```diff
@@ -79,9 +79,7 @@ def singleValueFromClient(self, value, skel, name, origData):
 			:returns: str or None
 		"""
 		rawValue = value
-		if not rawValue:
-			return None, [ReadFromClientError(ReadFromClientErrorSeverity.Empty, name, "No value selected")]
-		elif str(rawValue).replace("-", "", 1).replace(".", "", 1).isdigit():
+		if str(rawValue).replace("-", "", 1).replace(".", "", 1).isdigit():
```
sveneberth, line 82: I think you missed ":", or not?

bones/numericBone.py
```diff
@@ -41,11 +42,17 @@ def isInvalid(self, value):
 		if value != value:  # NaN
 			return "NaN not allowed"
 
+	def getEmptyValue(self):
+		return 0
+
+	def isEmpty(self, rawVale: Any):
+		return not (rawVale != self.getEmptyValue() or bool(rawVale))
```
sveneberth, line 49: here as well
```suggestion
	def isEmpty(self, rawValue: Any):
		return not (rawValue != self.getEmptyValue() or bool(rawValue))
```

bones/numericBone.py
sveneberth: This should be done with the `isEmpty` and `getEmptyValue` methods.

bones/dateBone.py
tsteinruecken: No, this is a timestamp. We'll just remove minus and the first decimal point to be able to check if every remaining character is a digit.

bones/booleanBone.py
tsteinruecken: Correct. Is Fixed.

tsteinruecken: Correct. Is Fixed.

PR #88, opened 2020-07-28
Fix/time zone handling

tsteinruecken:
Handling timezones was still somewhat broken in ViUR3 (due to datastore now returning/expecting tz-aware date objects).
This removed a lot of now obsolete code and desolves the ExtendedDateTimeClass

### DIFF
--- a/bones/dateBone.py
+++ b/bones/dateBone.py
@@ -19,35 +19,9 @@ datetime.now().strptime("2010%02d%02d" % (1, 1), "%Y%m%d")
 datetime.now().strftime("%Y%m%d")
 
 
-class ExtendedDateTime(datetime):
-	def totimestamp(self):
-		"""Converts this DateTime-Object back into Unixtime"""
-		return (int(round(mktime(self.timetuple()))))
-
-	def strftime(self, format):
-		"""
-		Provides correct localized names for directives like %a which dont get translated on GAE properly
-		This currently replaces %a, %A, %b, %B, %c, %x and %X.
-
-		:param format: String containing the Format to apply.
-		:type format: str
-		:returns: str
-		"""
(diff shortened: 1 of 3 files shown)


bones/dateBone.py
```diff
@@ -19,35 +19,9 @@
...
-		if "%a" in format:
-			format = format.replace("%a", translate("const_day_%s_short" % int(super(ExtendedDateTime, self).strftime("%w"))))
-		if "%A" in format:
-			format = format.replace("%A", translate("const_day_%s_long" % int(super(ExtendedDateTime, self).strftime("%w"))))
-		if "%b" in format:
-			format = format.replace("%b", translate("const_month_%s_short" % int(super(ExtendedDateTime, self).strftime("%m"))))
-		if "%B" in format:
-			format = format.replace("%B", translate("const_month_%s_long" % int(super(ExtendedDateTime, self).strftime("%m"))))
-		return super(ExtendedDateTime, self).strftime(format)
+def datetimeToTimestampp(datetimeObj: datetime) -> int:
```
sveneberth, line 22: Typo:
```suggestion
def datetimeToTimestamp(datetimeObj: datetime) -> int:
```

tsteinruecken:
It's not used inside the server. However one project i've checked uses this so i included this function so there is no loss in functionality. Maybe we could add a deprecation warning to this function and remove it once noone depends on it? Or just drop it now and the projects have to ship one themselves?

tsteinruecken:
@phorward Any preferences about datetimeToTimestamp? Keep or remove?

phorward:
> @phorward Any preferences about datetimeToTimestamp? Keep or remove?

Keeping is better.

PR #92, opened 2020-07-29
Let embedsvg load icons from poolfiles

phorward:
This implements the embedsvg feature for the HTML-renderer in an improved version, by using poolfiles for icons that can be shared both by server and client apps.

The SVG-images are normally generated into a JSON-file located somewhere in the /static folder, where both server and client can fetch the same icon set from one specific place. The default is /static/icons.json, this can be changed or extended to other files using the configuration variable `conf["viur.render.html.embedsvg.path"]`.

As a fallback, this function will also look into the /html/embedsvg folder for an SVG-file with the particular name. This fallback behavior was standard in ViUR2, the file-based approach described above is standard in ViUR3 now.

### DIFF
--- a/__init__.py
+++ b/__init__.py
@@ -16,7 +16,7 @@
    I N F O R M A T I O N    S Y S T E M
 
  ViUR SERVER
- Copyright 2012-2019 by Mausbrand Informationssysteme GmbH
+ Copyright 2012-2020 by Mausbrand Informationssysteme GmbH
 
  ViUR is a free software development framework for the Google App Engine™.
  More about ViUR can be found at https://www.viur.is/.
@@ -34,6 +34,7 @@ from viur.core import request
 from viur.core import languages as servertrans
 from viur.core.i18n import initializeTranslations
 from viur.core import logging as viurLogging  # Initialize request logging
+from viur.core.embedsvg import initializeEmbedSvgPool
 from viur.core.utils import currentRequest, currentSession, currentLanguage, currentRequestData
 from viur.core.session import GaeSession
 import logging
@@ -278,6 +279,8 @@ def setup(modules, render=None, default="html"):
 					uri.lower().startswith("https://") or uri.lower().startswith("http://"))
 	runStartupTasks()  # Add a deferred call to run all queued startup tasks
 	initializeTranslations()
+	initializeEmbedSvgPool()
+
 	assert conf["viur.file.hmacKey"], "You must set a secret and unique Application-Key to viur.file.hmacKey"
 	return app
 
--- a/config.py
+++ b/config.py
@@ -99,6 +99,12 @@ conf = {
 	# Allows the application to register a function that's called before the request gets routed
 	"viur.requestPreprocessor": None,
 
+	# Pathes to icon pool files (JSON-files) used to fill the icon pool
+	"viur.render.html.embedsvg.path": ["/static/icons.json"],
+
+	# This is the embedSvg icon pool
+	"viur.render.html.embedsvg.pool": None,
+
 	# Characters valid for the internal search functionality (all other chars are ignored)
 	"viur.searchValidChars": "abcdefghijklmnopqrstuvwxyz0123456789",
 
--- a/config.py
+++ b/config.py
@@ -100,7 +100,7 @@ conf = {
 	"viur.requestPreprocessor": None,
 
 	# Pathes to icon pool files (JSON-files) used to fill the icon pool
-	"viur.render.html.embedsvg.path": ["/static/icons.json"],
+	"viur.render.html.embedsvg.path": ["/static/embedsvg.json"],
 
 	# This is the embedSvg icon pool
 	"viur.render.html.embedsvg.pool": None,
--- a/embedsvg.py
+++ b/embedsvg.py
@@ -1,0 +1,28 @@
+import os, json, logging
+from viur.core import conf
+
+
+def initializeEmbedSvgPool():
+	if conf["viur.render.html.embedsvg.pool"] is None:
+		conf["viur.render.html.embedsvg.pool"] = {}
+
+		for path in conf["viur.render.html.embedsvg.path"]:
+			logging.debug("embedsvg caching path %r", path)
+
+			content = None
+			try:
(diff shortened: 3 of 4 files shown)


tsteinruecken:
The whole think looks wrong. I thought embedding SVGs will be phased out as it's worse for the client to get the same image inlined three times in the current page and 5 more on the next. Meanwhile styling SVGs with CSS is now possible when using an object tag. We could move the icons from html/ to static/ in case you need to to access them from outside.

phorward:
> The whole think looks wrong.

@tsteinruecken: NO! This pull requests implements a way that makes the entire embedsvg feature transparent to all parts of a ViUR application. The Vi will soon be powered by [flare](https://github.com/mausbrand/flare), same as other apps already do. flare fetches the same json-file with the embedded images, and it works like a charm. A JS-based app can fetch the icon pool and can work with it. The core fetches and uses it with this pull request.

This feature IS blocking. I'll set the blocking flag back.

phneutral:
How often should we explain that it is important to have a toolchain for pictograms that is easy to maintain and use?

This includes ...
... easily add existing icon collections to a project
... add project-specific icons to the existing collection without much effort
... CSS manipulability (color, shape, size, context reactivity)
... a corresponding interface in the project with which you can use quasi-native icons in any context

embedSVG offers all this in different versions in various projects distributed already now. We just want to put these different parts of the solution together and make them applicable for a large part of the projects.

And no: Object CSS manipulation is not even remotely comparable to SVG CSS manipulation. We have already solved the mentioned problem of multiple embedding in one project by using <svg><use href="#icon-name"></use></svg> and a central context sensitive icon store where each icon is actually loaded only once. 

We are happy to use an altvernative toolchain if anyone can present me with a similarly sophisticated solution. Until then, we'll use what we have. This one.

phorward:
I've updated to branch in compliance with @sveneberth to configure /static/embedsvg.json as the default poo file.
Again, I have  to call on importance: This feature is blocking us in straighten the icon toolchain for a current, very important project.

phorward:
fixed docstring in  bbc071a  @tsteinruecken  please use Squash+Merge feature from Git and GitHub to keep commits low.

tsteinruecken:
What's wrong about moving the svg folder from html/ to static/ (so you can reach them from outside)?

You can
... easily add existing icon collections to a project (Just drop them in the folder)
... add project-specific icons to the existing collection without much effort (No Toolchain to invoke, no nothing)
... CSS manipulability (color, shape, size, context reactivity) - As said - this works. We use this already. I've added 3 Testscripts (have to be served from a webserver - try it yourself)
... a corresponding interface in the project with which you can use quasi-native icons in any context (Just keep embedsvg for now and made it yield an object tag to /static)

Please name what exctly is not possible with an object tag - See Samples:
Example 1) Simply embed the CSS-Link into the SVG - no nothing on the client side, but you'll have to modify the SVG
Example 2) Keep the SVG unmodified - Add CSS-Link by JS
Example 3) Keep the SVG unmodified, use JS to replace the object tag with inline-svg client side - can even keep existing css - fully context aware

Inlining them Serverside has dozens of drawbacks
- You'll send an Image multiple times to the client (maybe even multiple times in one page!)
- You prevent any sort of caching (Google provides these Edge-Caches for a reason)
- You prevent the client from loading ressources in the best order (important things like CSS first)
- You forcefully push all (probably 100+) icons to flare - even if the application only uses a tiny subset of it

I've checked the startpage of one of our projects (the big red S). Replacing Inline-Icons with Object-Tags decreases the HTML transferred by over 15%. Which also decreases Server-Cost for that request by the same >15%. And that's the cached result. Building pages from scratch will be worse!

Is that worth it? I highly doubt it. Theres also a small race-condition in that code (as we use threads now). That initialization of viur.render.html.embedsvg.pool has to move to instance-startup (core.setup or the like). (I can't fix that).

If you're unable to adapt your code now my suggestion would be: Merge this temporarily, get the current project up, re-evaluate this in November in a Live-Project (A:B-Test) and drop that function before final release of ViUR3 if we see adverse effects.
[svg-test.zip](https://github.com/viur-framework/viur-core/files/5002119/svg-test.zip)

XeoN-GHMB:
The main problem with the object tag is that the svg is no longer part of the actual DOM. This means that css cascades like .wrapper .button:hover .icon can no longer be used.

Since all projects based on Ignite are affected, I recommend to merge the PR for now. 

A re-evaluation with the whole Ignite - Team should be planned for later.

tsteinruecken:
In Example 3 it *is* part of the dom. There is *no* difference from having it inlined Serverside. Zero.
Yes - it is just another way of providing icons. (That's why it is obviously impossible to have svgs in the old location and the json in the new one which leads to this beeing "blocking" :). Problem is that the concept is bad. 

"Can be bypassed with the use-tag": Nope. Cannot be. You can (with enough effort) prevent it from being embedded twice in the same page. But as soon the page reloads - or the user loads another Page you'll get the icons again. Just see big red S - Menu and Footer...

"the embedsvg.json is cached as well?" - Yes. But i was referring to the ones embedded server side into html. No Chance here.

""The Big S" has the ability to preload only the icons that are necessary": Again mixing both sides. Flare will have to fetch the entire json with all the unnecessary icons (i don't see any chance to be picky here); the ones you embed have totally different issues.

Again: Someone fix that racecondition, we'll merge that temporarily and re-evaluate this later

XeoN-GHMB:
Last comment from me on this. Example3 uses img tags... 
- Img tags do not allow re-coloring on css basis.
- Object tags can include a css file but are not part of the DOM.

In my opinion the use case is not understood here.

phorward:
> Again: Someone fix that racecondition, we'll merge that temporarily and re-evaluate this later

Hopefully db6b264c319b676a614fedfb6aed7ef5bcedd3a3 does this now.

tsteinruecken:
Can this one be closed? I think it has been resolved otherwise. @phorward

sveneberth:
> Can this one be closed? I think it has been resolved otherwise. @phorward

Because there's no response from @phorward within the last 23 days and IMO this is solved with #102 I close this PR now.

PR #94, opened 2020-08-04
Use skeleton.setBoneValue() for defaultValues

phorward:
This feature is needed to set a relationalBone's defaultValue via a callback.

### DIFF
--- a/skeleton.py
+++ b/skeleton.py
@@ -139,7 +139,12 @@ class SkeletonInstance:
 				if self.dbEntity is not None:
 					boneInstance.unserialize(self, key)
 				else:
-					self.accessedValues[key] = boneInstance.getDefaultValue()
+					defaultValue = boneInstance.getDefaultValue()
+					if defaultValue:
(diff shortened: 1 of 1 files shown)


tsteinruecken:
That line you uncommented is there for a reason (despite the misleading fixme). Skeleton-Values are loaded lazy - they get de-serialized from the underlying db.Entity only on first access. That line accesses that entry your're trying to set (forcing it to get loaded if it has not yet been accessed). Without it f.e. appending to an relation that you have not accessed before will just overwrite whatever is in the datastore instead of appending to it as the current content never got deserialized. Therefore we cannot use setBoneValue in getDefaultValue as it will create a loop (trying to read a value -> retrieving its default value if called on a new skeleton -> calling setBoneValue -> trying to read a value). I'll put a new PR that will make it easier to construct valid defaultValues for relations.

phorward:
>  I'll put a new PR that will make it easier to construct valid defaultValues for relations.

That's OK for me! :+1:

PR #106, opened 2020-08-11
Made the session-module TLS-Only 

tsteinruecken:
- No more support for cookies over unencrypted http
- Made the cookie SessionOnly (no max-age or expire time set) by default (it will vanish when the browser is closed)

### DIFF
--- a/session.py
+++ b/session.py
@@ -29,13 +29,9 @@ from viur.core.config import conf
 class GaeSession:
 	"""Store Sessions inside the Big Table/Memcache"""
 	kindName = "viur-session"
-	sameSite = "lax"
-
-	def __init__(self):
-		super().__init__()
-
-		self.plainCookieName = f"viurHttpCookie_{utils.projectID}"
-		self.sslCookieName = f"viurSSLCookie_{utils.projectID}"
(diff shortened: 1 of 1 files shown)


session.py
```diff
@@ -29,13 +29,9 @@
...
-	sameSite = "lax"
-
-	def __init__(self):
-		super().__init__()
-
-		self.plainCookieName = f"viurHttpCookie_{utils.projectID}"
-		self.sslCookieName = f"viurSSLCookie_{utils.projectID}"
+	sameSite = "lax"  # Either None (dont issue sameSite header), "none", "lax" or "strict"
+	sessionCookie = True  # If True, issue the cookie without a lifeTime (will disappear on browser close)
+	cookieName = f"viurCookie_{utils.projectID}"
```
phorward, line 34: For above 3 lines, is there a way or possibility to customize or override the default values with minor effort?

tsteinruecken: That's why there left as class variables instead of being hard-coded into the one line that uses them. Shall we move them to conf? (We might want to decide if we put everything in conf - or if me move stuff that's rarely and only once used to the class it belongs to)

phorward: hmm okay then it should satisfy this. I'll approve for now.

phorward:
For testing, I've merged the current master into this branch.

PR #107, opened 2020-08-11
Provide a dedicated encodeKey function

tsteinruecken:


tsteinruecken:
@sveneberth We have the keyHelper function for that. It also validates that the kind encoded in that key is the one you expect

PR #108, opened 2020-08-11
Implemented the QueryIter class.

tsteinruecken:
This allows to iterator over a infinite amount of entries (you get a callback for each entry matched)

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -11,6 +11,7 @@ from google.protobuf import timestamp_pb2
 from viur.core import db, errors, utils
 from viur.core.config import conf
 from viur.core.utils import currentRequest, currentSession
+from time import sleep
 
 
 class JsonKeyEncoder(json.JSONEncoder):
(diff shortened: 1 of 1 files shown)


tasks.py
```diff
@@ -497,44 +515,132 @@ def runStartupTasks():
...
-"""
+			qryIter = qry.run(5)
+		for item in qryIter:
+			try:
+				cls.handleEntry(item, qryDict["customData"])
+			except:
+				sleep(5)
+				try:
+					cls.handleEntry(item, qryDict["customData"])
+				except Exception as e:
```
sveneberth, line 619: It could be helpful to have a parameter/option `breakOnException` to stop the whole iteration after an error occurred and call an method like `onError(item, customData, exc)`.

tasks.py
```diff
@@ -497,44 +515,132 @@ def runStartupTasks():
...
+	@classmethod
+	def handleEntry(cls, entry, customData):
+		"""
+			Overridable hook to process one entry. "entry" will be either an db.Entity or an
+			SkeletonInstance (if that query has been created by skel.all())
+
+			Warning: If your query has an sortOrder other than __key__ and you modify that property here
+			it is possible to encounter that object later one *again* (as it may jump behind the current cursor).
+		"""
+		logging.debug("handleEntry called on %s with %s." % (str(cls), entry))
```
sveneberth, line 639: You could save the `str`-cast. The formatstring will already do this.
```suggestion
		logging.debug("handleEntry called on %s with %s.", cls, entry)
```

tasks.py
```diff
@@ -497,44 +515,132 @@ def runStartupTasks():
...
+
+	@classmethod
+	def handleFinish(cls, totalCount: int, customData):
+		"""
+				Overridable hook that indicates the current run has been finished.
+		"""
+		logging.debug("handleFinish called on %s with %s total Entries processed" % (str(cls), totalCount))
```
sveneberth, line 646: here as well
```suggestion
		logging.debug("handleFinish called on %s with %s total Entries processed", cls, totalCount)
```

sveneberth (dismissed):

sveneberth (dismissed):

sveneberth:
> This looks fine for me as a first step.
> 
> As an reference implementation, I would like to see the rebuildSearchIndex standard task using this technique, for a better understanding.

See #160 for this 😃

sveneberth (approved):

PR #114, opened 2020-08-13
keyBone is broken

phorward:
Hi, the keyBone seems to be broken, especially when configured as `keyBone(multiple=True)`. It is stored correctly, but loading failed. This just makes it load and store correctly, but I'm unsure if it's necessary to implement `singleValueSerialize`, `singleValueUnserialize` or `singleValueFromClient` for it. What do you think?


tsteinruecken:
Should have been resolved by #121

Issue #115, opened 2020-08-13
Limit multiple-bones to maximum amount of entries


phorward:
It might be useful to configure bones as multiple, but limit the amount of entries allowed. Especially when working with subSkels, this would be essential: maybe one kind of the subSkel allowes for only one entry, where others allow for more. Configuring the bone in the addSkel or editSkel differently might run into problems, as there might be entites with either lists or single entries.

An implementation might look like this:
```python
unlimited = stringBone(multiple=True)
limitedTo3 = stringBone(multiple=3)
```

This is just an idea that should be discussed at some time.

phorward:
slightly related to #79

PR #122, opened 2020-08-20
First draft of a multiple constraint

### LINKED ISSUE #79: relationalBone(multiple=True) but with unique relations per destination
Hi there,

I'm having a requirement for some kind of unique-Feature for relationalBone(multiple=True) bones, where multiple relations to different destination can be added, but only one relation for each target key is allowed.
In the current implementation, I can create multiple relations to the same destination entity.

How could this be implemented?
How could it be named? (I think `unique` is not the right wording, because it is used differently).

tsteinruecken:
This allows to set min/max length of a multiple bone and whenever duplicates are allowed.
This will currently only work with string/numericBones, for relations we'd have to decide what counts as duplicate.
This should resolve #79 and #115

### DIFF
--- a/bones/__init__.py
+++ b/bones/__init__.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 
-from viur.core.bones.bone import baseBone, UniqueValue, UniqueLockMethod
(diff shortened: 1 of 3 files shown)


sveneberth (approved):
Looks fine

PR #123, opened 2020-08-28
Started to generialize setBoneValue to handle multiple and languages

tsteinruecken:


bones/relationalBone.py
```diff
@@ -822,6 +822,7 @@ def setBoneValue(self, skel, boneName, value, append, *args, **kwargs):
 			:return: Wherever that operation succeeded or not.
 			:rtype: bool
 		"""
+		assert not language, "Language not yet supported"  # FIXME!
```
sveneberth, line 825: IMO it's more appropriate to raise a `NotImplementedError` here.

bones/bone.py
```diff
@@ -685,32 +685,44 @@ def mergeFrom(self, valuesCache, boneName, otherSkel):
...
 			:type boneName: object
 			:param append: If true, the given value is appended to the values of that bone instead of
 				replacing it. Only supported on bones with multiple=True
 			:type append: bool
+			:param language: Set/append which language
+			:type language: str or None
 			:return: Wherever that operation succeeded or not.
 			:rtype: bool
 
 		"""
```
sveneberth, line 707: i'am not really satisfied with this types here.
The type annotations and docstring-types are sometimes different (must boneName be a str or object) or they're missing (Skel[e]tonInstance).
I think we don't need them redudant and should remove one kind of them.
I suggest to eliminate the types in the docstring and use only the type annotations in the signature.

sveneberth (commented):
.

sveneberth (changes requested):
.

bones/bone.py
```diff
@@ -715,32 +715,39 @@ def mergeFrom(self, valuesCache, boneName, otherSkel):
 			return
 		valuesCache[boneName] = copy.deepcopy(otherSkel.valuesCache.get(boneName, None))
 
-	def setBoneValue(self, skel, boneName, value, append, *args, **kwargs):
+	def setBoneValue(self, skel: 'SkeltonInstance', boneName: str, value: Any, append: bool,
```
sveneberth, line 718:
```suggestion
	def setBoneValue(self, skel: 'SkeletonInstance', boneName: str, value: Any, append: bool,
```

As already noted in the [last review](https://github.com/viur-framework/viur-core/pull/123#discussion_r479362994), there is a typo.

sveneberth (changes requested):

sveneberth (approved):

PR #124, opened 2020-08-29
Extend relationalBone to hook-able getEntity function

phorward:
I need this to override the way how an entity is retrieved: It may also be referenced by an unique îdentifier that needs to be looked-up. Here's an example implementation:

```python
class appBone(relationalBone):
	"""
	appBones is a relationalBone that accepts also the app identifier as valid key...
	"""

	def __init__(self, descr="App", kind="app", indexed=True, *args, **kwargs):
		super().__init__(kind=kind, descr=descr, indexed=indexed, *args, **kwargs)

	def getEntity(self, key):
		entity = super().getEntity(key)

		if not entity:
			entity = db.Query(self.kind).filter("identifier", key).getEntry()

		return entity
```

### DIFF
--- a/bones/relationalBone.py
+++ b/bones/relationalBone.py
@@ -384,8 +384,7 @@ class relationalBone(baseBone):
 			dbKey = None
 			errors = []
 			try:
-				dbKey = db.keyHelper(key, self.kind)
-				entry = db.Get(dbKey)
+				entry = self.getEntity(key)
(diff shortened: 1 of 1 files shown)


tsteinruecken:
Can we push this to the next ViUR meeting? There no comments yet and i'm also quite unsure if that's a good idea...

PR #126, opened 2020-09-10
Update editform_bone_bool.html

xnopasaranx:
value of checkbox is not accepted by server, needs to be capitalized!

### DIFF
--- a/template/editform_bone_bool.html
+++ b/template/editform_bone_bool.html
@@ -7,7 +7,7 @@
 			id="ignt-id-{{ boneName }}"
 			type="checkbox"
 			name="{{ boneName }}"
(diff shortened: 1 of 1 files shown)


sveneberth (approved):
IMO we should rather (or additionally) add support for lowercase in the booleanBone instead of adjust the template.

Lines like https://github.com/viur-framework/viur-core/blob/8f0843bfa62079691138e4474dc7883fe2a6ee8b/bones/booleanBone.py#L21 should be:
```python
if str(value).lower() in self.trueStrs:
```
(and `trueStrs` contains the boolean-representations in lowercase)

<hr>

But of course is this a valid solution for the bug.

PR #129, opened 2020-09-21
Strip username from userpassword/login parameters

phorward:
It already happened twice in two ViUR projects, that users reported a login problem which was obviously heavy to find. The reason is: There was a space character behind the user's name. In one case, this was inserted automatically by the operating system which was Android and an android keyboard app. To avoid this problem in future, a strip() function is applied onto the username to make it clean.

### DIFF
--- a/modules/user.py
+++ b/modules/user.py
@@ -142,8 +142,10 @@ class UserPassword(object):
 		if not name or not password or not securitykey.validate(skey, useSessionKey=True):
 			return self.userModule.render.login(self.loginSkel())
 
+		name = name.lower().strip()
+
 		query = db.Query(self.userModule.viewSkel().kindName)
-		res = query.filter("name.idx >=", name.lower()).getEntry()
+		res = query.filter("name.idx >=", name).getEntry()
 
 		if res is None:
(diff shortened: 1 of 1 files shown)


modules/user.py
```diff
@@ -142,8 +142,10 @@ def login(self, name=None, password=None, skey="", *args, **kwargs):
 		if not name or not password or not securitykey.validate(skey, useSessionKey=True):
 			return self.userModule.render.login(self.loginSkel())
 
+		name = name.lower().strip()
```
sveneberth, line 145: Due to this assignment `fromClient` in L. 182 will not get the original value. 
Im not sure whether it breaks something, but maybe this could be an unwanted side effect.

sveneberth (changes requested):

tsteinruecken:
It's okay for me. We should probably remove that skel.fromClient({"name": name, "nomissing": "1"}) Line all together. If we don't reflect any values back we can't have an injection even if our sanitation failes and we don't give the user the (maybe) false hint that the password was incorrect (maybe the username/email did not exist).

tsteinruecken:
Replaced by #133

PR #130, opened 2020-09-25
added backwards compatibility to recordBone singleValueUnserialize method

xnopasaranx:
Same as in relationalBone, when working with existing viur2 datastore entries we will need this. Don't know if it has already been provided by someone else in some other branch though...

### DIFF
--- a/bones/recordBone.py
+++ b/bones/recordBone.py
@@ -67,7 +67,7 @@ class recordBone(baseBone):
 							res.append(tag)
 			return res
 
-		value = values.get(key)
+		value = values[key]
 		res = []
 
 		if not value:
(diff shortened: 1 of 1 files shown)


sveneberth:
I think this shouldn't be part of the core.
If you migrate from ViUR 2 to ViUR 3 it's a breaking change. So you have to adjust your code and data anyway. You have to rewrite this data manually with a script one time.

tsteinruecken (dismissed):
Looks good to me.. We have this already for relations; we hopefully can migrate a project by just migrating the blobs and rebuild the datastore inplace

xnopasaranx:
@sveneberth I am unsure if it is that easy. Yes, we would need to rewrite a bunch of existing data but thinking about some of our bigger customer projects where we are dealing with huge sets of data that are mission critical, I think we need backwards compatibility.

tsteinruecken (approved):

PR #137, opened 2020-10-07
Push /file/download requests into the ede caches …

tsteinruecken:
…if it does not expire; require exiration explictly set in jinja2 render


sveneberth (approved):

xnopasaranx:
Not sure if this has been tested thoroughly but it just broke my project. All templates that execute downloadUrlFor explode with this change. Do I have to specify an expiration as an argument now?

PR #147, opened 2020-10-14
Feature/hooks before actions

phorward:
Additionally to the **onAdded**, **onEdited**, **onDeleted** hooks, this pull request introduces further hooks called before the specific database operation is performed, using **onAdd**, **onEdit** and **onDelete**.
Furthermore **onViewed** has been renamed to **onView** because the hook is performed before the operation takes place.

### DIFF
--- a/prototypes/list.py
+++ b/prototypes/list.py
@@ -180,7 +180,7 @@ class List(BasicApplication):
 		or as the first parameter in *args*. The function performs several access control checks
 		on the requested entity before it is modified.
 
-		.. seealso:: :func:`editSkel`, :func:`onEdited`, :func:`canEdit`
+		.. seealso:: :func:`editSkel`, :func:`onEdit`, :func:`onEdited`, :func:`canEdit`
 
 		:returns: The rendered, edited object of the entry, eventually with error hints.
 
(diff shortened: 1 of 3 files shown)


prototypes/singleton.py
```diff
@@ -286,26 +300,26 @@ def onEdited(self, skel):
...
 		The default is doing nothing.
 
-		:param skel: The Skeleton that is viewed.
+		:param skel: The Skeleton that is being viewed.
 		:type skel: :class:`server.skeleton.Skeleton`
 
 		.. seealso:: :func:`view`
 		"""
-		pass
+		logging.debug("onView")
```
sveneberth, line 322: I think this will spam the logs :/

prototypes/list.py
```diff
@@ -546,7 +578,20 @@ def onViewed(self, skel):
 
 		.. seealso:: :func:`view`
 		"""
-		pass
+		logging.debug("onView")
```
sveneberth, line 581: I think this will spam the logs :/

prototypes/singleton.py
sveneberth, line 322: this as well

prototypes/tree.py
```diff
@@ -754,7 +783,20 @@ def onViewed(self, skel):
 
 		.. seealso:: :func:`view`
 		"""
-		pass
+		logging.debug("onView")
```
sveneberth, line 786: this as well

prototypes/list.py
sveneberth, line 581: I think this will spam the logs :/

sveneberth (changes requested):

tsteinruecken:
I don't mind. For view it's probably better to remove the logging; but add/edit/deletion isn't that common...

phorward:
All loggings had been replaced by passes.

sveneberth (approved):

PR #151, opened 2020-10-19
fixed shortKey to work with new Key Format

Grashalmbeisser:


phorward (approved):
Thats good!

I'm not sure if we also might handle the case when val is already a db.Key instance? @tsteinruecken

tsteinruecken:
Should not happen. All keys are encoded as string before being passed to the template

PR #152, opened 2020-10-22
Add rawValueBone

XeoN-GHMB:
The codebone is a typed basebone for easier access in the frontend

### DIFF
--- a/bones/codeBone.py
+++ b/bones/codeBone.py
@@ -1,0 +1,10 @@
+# -*- coding: utf-8 -*-
+from viur.core.bones import baseBone
+
+class codeBone(baseBone):
+	'''
+		The codebone does not check and excaped the entered content.
+		This can lead to security problems.
+		Therefore, like the basebone, use this bone with caution.
+	'''
+	type = "code"
--- a/bones/codeBone.py
+++ b/bones/codeBone.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 from viur.core.bones import baseBone
 
 class codeBone(baseBone):
--- a/bones/codeBone.py
+++ b/bones/codeBone.py
@@ -1,9 +1,0 @@
-from viur.core.bones import baseBone
-
-class codeBone(baseBone):
-	'''
-		The codebone does not check and excaped the entered content.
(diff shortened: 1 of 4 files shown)


sveneberth (dismissed):

bones/codeBone.py
```diff
@@ -0,0 +1,10 @@
+# -*- coding: utf-8 -*-
+from viur.core.bones import baseBone
+
+class codeBone(baseBone):
+	'''
+		The codebone does not check and excaped the entered content.
+		This can lead to security problems.
+		Therefore, like the basebone, use this bone with caution.
```
phorward, line 8: Please change this to:
> Likeweise the baseBone, the codeBone does not check or escape the provided content.
> This might lead to security problems, but is required under some circumstances.
> codeBone just specializes its content more specifically so it can be easily adapted.
> Use the dot-notation like "code.markdown" or similar to describe subsequent types.

bones/codeBone.py
```diff
@@ -0,0 +1,10 @@
+# -*- coding: utf-8 -*-
```
phorward, line 1: Please remove this line, as it is not required anymore (see [here](https://stackoverflow.com/questions/14083111/should-i-use-encoding-declaration-in-python-3))

phorward (dismissed):
Please apply the two provided comments.

tsteinruecken:
CodeBone may be a bad name. Can we come up with a name that explicitly states that unlike all other bones, no sanitation is performed on this bone?

sveneberth:
> CodeBone may be a bad name. Can we come up with a name that explicitly states that unlike all other bones, no sanitation is performed on this bone?

I thought we had already agreed on this name together in a ViUR round... For me is the name a low hanging fruit, `rawBone` or something like this would be okay for me as well.

sveneberth (changes requested):
@XeoN-GHMB Can you please apply @phorward's improved docstring?

tsteinruecken:
I've now renamed it to rawValueBone, updated the docstring and ensured that the baseBone will not read a value in fromClient.

sveneberth (approved):

PR #153, opened 2020-10-23
Revert to old server-behavior: bones are not indexed by default

sveneberth:


xnopasaranx (approved):

xnopasaranx:
this should be the other way around by default in my opinion. otherwise people will definitely forget to set indexed=False on bones and cause massive writeops

tsteinruecken:
I would suggest keeping the default indexed (but set textbones to not indexed to avoid issues with text longer than 1,5kb). There is no additional cost involved using more indexed fields - and i've not seen an increased write latency anymore. On the other hand we have often issues with queries failing due to fields not being indexed.

phorward (approved):

tsteinruecken:
As discussed, we'll keep the new indexed=True default behavior and just set it on textbones to false

sveneberth:
> As discussed, we'll keep the new indexed=True default behavior and just set it on textbones to false

It would have been nice if you had provided an alternative PR than just closing that.

PR #174, opened 2020-12-09
respecting defaultValue setting of selectBone in renderEditForm template

skoegl:
Now we can again see the defaultValue of
 boneParams in rendered html select options selected.

Disclaimer: not tested yet the multiple code path...

### DIFF
--- a/template/editform_bone_select.html
+++ b/template/editform_bone_select.html
@@ -8,7 +8,7 @@
                     name="{{ boneName }}"
                     value="{{ key }}"
                     type="checkbox"
-                    {{ "checked" if key in boneValue["key"] }}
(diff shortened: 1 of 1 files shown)


template/editform_bone_select.html
```diff
@@ -8,7 +8,7 @@
                     name="{{ boneName }}"
                     value="{{ key }}"
                     type="checkbox"
-                    {{ "checked" if key in boneValue }}
+                    {{ "checked" if key in boneValue["key"] }}
```
tsteinruecken, line 11: @skoegl Did you test this case (multiple=True)?  boneValue should be a dict[(key: str), (descr: str)] in this case..

tsteinruecken (approved):

PR #179, opened 2020-12-14
fixed a mistake in numericBone that was casting numerical input to str ...

xnopasaranx:
… on accident and thus evaluating the bone as having an empty value.

### DIFF
--- a/bones/numericBone.py
+++ b/bones/numericBone.py
@@ -48,14 +48,16 @@ class numericBone(baseBone):
 			return 0
 
 	def isEmpty(self, rawValue: Any):
-		if isinstance(rawValue, str):
-			try:
-				if self.precision:
-					rawValue = float(rawValue.replace(",", ".", 1))
-				else:
-					rawValue = int(rawValue)
-			except:
-				return True
+		# try to cast rawValue to a float and make it comparable, then if bone does not have precision flag set,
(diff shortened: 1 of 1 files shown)


bones/numericBone.py
```diff
@@ -48,15 +48,14 @@ def getEmptyValue(self):
...
-			else:
-				rawValue = int(rawValue.replace(",", ".", 1))
-		except:
-			return True
+		if isinstance(rawValue, str):
+			try:
+				if self.precision:
+					rawValue = float(rawValue.replace(",", ".", 1))
+				else:
+					rawValue = int(rawValue)
```
sveneberth, line 56: I think we should also accept here strings like `"123,89"` and round them down to a int. This would be the analogous behavior if the precision is a positiv number and you supplied more fractional digits than allowed (see [l. 69](https://github.com/viur-framework/viur-core/blob/e7ec7bc67776250f70d22bc5e5e67204aa294b95/bones/numericBone.py#L69) for this).

```suggestion
					rawValue = int(float(rawValue.replace(",", ".", 1)))
```

sveneberth (changes requested):

sveneberth (approved):

tsteinruecken:
Superseeded by #182

PR #182, opened 2021-01-05
Fixed .replace() being called on non-string values in numericBone:isEmpty

tsteinruecken:
Replacement for #179


xnopasaranx (approved):
works on my side with a testcase that would fail otherwise. used the same scenario as when testing https://github.com/viur-framework/viur-core/pull/179

PR #187, opened 2021-01-18
create fromClient method in datebone

klaus-thorres:
Hey guys!

I found out the dateBone failed saving data from Jinja-generated templates.
Through some testing, I found out that the singleValueFromClient method was not called by trying to save data from outside of vi. Instead the fromClient method was called. So I used the code from the original singleValueFromClient method to design a new fromClient method in the dateBone.

Currently adding from the jinja templates and adding + editing from vi is working. Further tests have not been done yet.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 *.pyc
 *.bak
 .idea/
+tags
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,4 @@
 *.pyc
 *.bak
 .idea/
-tags
+
--- a/bones/dateBone.py
+++ b/bones/dateBone.py
@@ -118,7 +118,7 @@ class dateBone(baseBone):
 					month = time_value_raw.minute
 					day = time_value_raw.second
 					value = datetime(year=year, month=month, day=day, hour=hour, minute=minute, second=second)
-				if str(time_value_raw).count(":") == 2:
+				elif str(time_value_raw).count(":") == 2:
 					(hour, minute, second) = [int(x.strip()) for x in str(time_value_raw).split(":")]
 					time_value = datetime(year=1970, month=1, day=1, hour=hour, minute=minute, second=second)
 				elif str(time_value_raw).count(":") == 1:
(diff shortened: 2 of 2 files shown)


tsteinruecken:
Hi and welcome to the project :)

Can you tell what exactly you're trying to achieve? Are you trying to modify data from within the template? (Which is not supposed to happen in MVC). The datebone inherits baseBone, which has the fromClient() function, which will in turn call said singleValueFromClient(). That fromClient function handles most of the multiple/languages wrapping for the bone, which we would have to duplicate if the dateBone get it's own fromClient() function.

klaus-thorres:
Hi @tsteinruecken 

I must confess that I do not yet fully understand the system. So this should be seen as a basis for discussion.

As I stated above, the main achievement of my version is that saving datetimeBones from the jinja templates containing date and time information is possible. Since these information were not processed by the original singleValueFromClient method and other bones also utilize the fromClient method by simultaneously changing data. I did not realize that these proceeding is not wanted.

phorward:
Hello there,

as additional information, the problem comes from the default implementation of editform_bone_date.html used by the HTML-renderer's renderForm-function. This template outputs in case `timestamp = dateBone(time=True, date=True)` is defined, two input fields, which ca named `timestamp_date` and `timestamp_time` (and a hidden one then called just `timestamp` that seems to be filled by JavaScript, but isn't). This Pull Requests shall both handle the case when one input field with both values exists (as the Vi handles it) or the values are separated into two input fields, with a date and time postfix.

I'm of the opinion that the singleValueFromClient()-function should implement this, but the notation should be separated by dot, so it looks either for a `timestamp` with both values, or for a `timestamp.date` and `timestamp.time`. This must be changed in both the bone and in the templates.

tsteinruecken:
If that's the issue then this javascript has to be fixed. The datebone expects them as a single string. Parsing two fields would require parseSubFields to be enabled (which is slower to parse and then also requires data to be transmitted as subfields - ie. it will break the current api). Patching fromClient as suggested would break support for multiple=True and/or multi-language support.

tsteinruecken:
Has to be fixed in our form.js

PR #189, opened 2021-01-20
create relationalBone template

klaus-thorres:
Hi :)

For our project we need a jinja template for the relational bone since it was not available in viur-core. So we created one and I like to share it with you.

The goal of is that it should be possible to select a dataset from the linked skeleton. As a starting point, I chose the code of the select bone while name value of the subordinated skeleton represents the whole dataset.
To overcome the limitation of 100 datasets per request a combination of two nested for loops is used so that the final limit of relational Bone is set to 1000. This value is selected  to prevent overloading the client or a disruptive delay while loading the form.

Also I would later implement an input element in combination with predefined options in a datalist element while it is possible to choose between the two options using a parameter in the boneParams.params dict.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,3 @@
 *.pyc
 *.bak
 .idea/
-tags
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 *.pyc
 *.bak
 .idea/
+tags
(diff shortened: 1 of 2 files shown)


phorward (approved):
this looks good and is an out-of-the box working solution for most cases.

phneutral (approved):

tsteinruecken:
As there has been no progress in the last 90 days and the form logic will be moved to base anyway im closing this pr.

tsteinruecken (commented):
For integration into viur-base

tsteinruecken:
My review-comments where not published; they are now - for reference for integration into base

PR #194, opened 2021-02-08
Ensure a valid limit is set in queries containing an IN filter

tsteinruecken:


db.py
```diff
@@ -823,7 +823,7 @@ def run(self, limit=-1, **kwargs):
 			res = []
 			# We run all queries first (preventing multiple round-trips to the server)
 			for singleQuery in self.queries:
-				res.append(self._runSingleFilterQuery(singleQuery, limit))
+				res.append(self._runSingleFilterQuery(singleQuery, limit if limit != -1 else singleQuery.limit))
```
sveneberth, line 826: Isn't 
```suggestion
				res.append(self._runSingleFilterQuery(singleQuery, limit if limit > 0 else singleQuery.limit))
```
more appropriate? A non-positive limit is nonsense for me.

sveneberth (changes requested):

db.py
tsteinruecken: That's correct. But -1 is the only negative limit we set (and it's never read from the outside). So if we encounter any other negative limit other than -1, the developer as deliberately set it - it's then probably better to pass it along.

PR #202, opened 2021-02-25
Made the json-render aware of our translation-class

tsteinruecken:
This allows translations to be used anywhere (bone-descriptions; it's params; in values for selectBones etc).
Replaces #15


sveneberth (approved):
Works for me with the changes of #208.

PR #204, opened 2021-02-25
Prevent add/delete etc being called on a skeleton if an outputrender is attached

tsteinruecken:
Fixes #187


sveneberth (approved):
Makes sense to me. But I don't understand why this should fix #187.

PR #206, opened 2021-02-26
Moved the emailtransport function and it's delivery success callback into a class

tsteinruecken:


email.py
```diff
@@ -37,6 +38,50 @@ def cleanOldEmailsFromLog(*args, **kwargs):
...
+			This function can be used to pre-validate the queue entity before it's deferred into the queue.
+			Must raise an exception if the email cannot be send (f.e. if it contains an invalid attachment)
+			:param entity: The entity to validate
+		"""
+		return
+
+	@staticmethod
+	def transportSuccessfulCallback(entity: db.Entity):
+		"""
+			This callback can be overriden by the project to execute additional tasks after an email
```
sveneberth, line 78:
```suggestion
			This callback can be overridden by the project to execute additional tasks after an email
```

email.py
```diff
@@ -7,14 +7,15 @@
...
 """
 	This module implements an email delivery system for ViUR. Emails will be queued so that we don't overwhelm
 	the email service. As the Appengine does not provide an email-api anymore, you'll have to use a 3rd party service
 	to actually deliver the email. A sample implementation for Send in Blue (https://sendinblue.com/) is provided.
-	To enable Send in Blue,	set conf["viur.email.transportFunction"] to transportSendInBlue and add your API-Key to
+	To enable Send in Blue,	set conf["viur.email.transportClass"] to EmailTransportSendInBlue and add your API-Key to
 	conf["viur.email.sendInBlue.apiKey"]. To send via another service, you'll have to implement a different transport
-	function (and point conf["viur.email.transportFunction"] to that function). This module needs a custom queue
+	class (and point conf["viur.email.transportClass"] to that function). This module needs a custom queue
 	(viur-emails) with a larger backoff value (sothat we don't try to deliver the same email multiple times within a
```
sveneberth, line 19:
```suggestion
	(viur-emails) with a larger backoff value (so that we don't try to deliver the same email multiple times within a
```

sveneberth (changes requested):
I miss the context-parameter in the sendEmail-method. And found some typos.

And can you please rearrange the imports (some of them are unused and the order is mixed up) and remove the first line (for py3 it's enough that the file is encoded in utf8).

sveneberth (approved):

PR #211, opened 2021-03-01
Prevent running with an insecure script-src policy

tsteinruecken:
Prevent running a ViUR3 Instance with insecure script-src CSP-Rules


tsteinruecken:
This has been rejected by the team; the removal of the default unsafe-inline will be another PR

Issue #212, opened 2021-03-02
baseBone-fallback for a recordBone failing with "'NoneType' object is not iterable" when None


phorward:
In this case, the recordBone usage `additional_addresses = recordBone()` produces this error when a fresh, empty skeleton shall be saved.
```
Traceback (most recent call last):
  File "/workspace/viur/core/request.py", line 221, in processRequest
    self.findAndCall(path)
  File "/workspace/viur/core/request.py", line 395, in findAndCall
    res = caller(*self.args, **self.kwargs)
  File "/workspace/modules/user.py", line 162, in login
    assert skel.toDB()
  File "/workspace/skeletons/user.py", line 349, in toDB
    return super().toDB(skel, clearUpdateTag)
  File "/workspace/viur/core/skeleton.py", line 883, in toDB
    key, dbObj, skel, changeList = db.RunInTransaction(txnUpdate, key, skelValues, clearUpdateTag)
  File "/workspace/viur/core/db.py", line 1007, in RunInTransaction
    res = callee(*args, **kwargs)
  File "/workspace/viur/core/skeleton.py", line 695, in txnUpdate
    bone.serialize(skel, key, True)
  File "/workspace/viur/core/bones/bone.py", line 408, in serialize
    for singleValue in newVal:
TypeError: 'NoneType' object is not iterable
```

tsteinruecken:
I thing the error is rather that the defaultValue for recordBones is None (despite it being multiple). We initialize that value in https://github.com/viur-framework/viur-core/blob/master/skeleton.py#L145 . Probably that bug is not limited to recordBones.

PR #213, opened 2021-03-02
Improved baseBone.serialize()

phorward:
This is a fix draft for #212

### DIFF
--- a/bones/bone.py
+++ b/bones/bone.py
@@ -388,8 +388,12 @@ class baseBone(object):  # One Bone:
 			:type name: str
 			:returns: dict
 		"""
-		if name in skel.accessedValues:
-			newVal = skel.accessedValues[name]
+		if name not in skel.accessedValues:
+			return False
+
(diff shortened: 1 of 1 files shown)


bones/bone.py
```diff
@@ -409,15 +413,19 @@ def serialize(self, skel: 'SkeletonInstance', name: str, parentIndexed: bool) ->
...
-			if indexed and name in skel.dbEntity.exclude_from_indexes:
-				skel.dbEntity.exclude_from_indexes.discard(name)
-			elif not indexed and name not in skel.dbEntity.exclude_from_indexes:
-				skel.dbEntity.exclude_from_indexes.add(name)
-			return True
-		return False
+		else:
+			res = None
+
+		skel.dbEntity[name] = res
```
phorward, line 419: It maybe is better to just leave `skel.dbEntity[name]` unset in case the value is none, what do you think @tsteinruecken

PR #214, opened 2021-03-02
First draft for evaluating typehints in exposed functions

tsteinruecken:
This is the base for discussion wherever we want to support type-conversion in exposed functions using the new type-annotations in python or not.


XeoN-GHMB:
This changes the behavior of a Python own function. I like it, because it makes the json interface more controllable. But I can imagine that there are cases where I don't care or don't know how the parameters are passed into the function. With this solution I can only use annotations in a way that all parameters always have to be sent via the kwargs. 

Like with the tasks I think a hidden parametr _disable could avoid the problem.

What exactly is a use case here?

tsteinruecken:
This is the result of a discussion between @phorward and me regarding how and where type-conversions should be happen. And the easiest way is of course having the core handling these for you. As far as i can tell this is completely backwards compatible as it's entirely optional. If there are no type-annotations that code is entirely bypassed. And if there are, only the parameters having an annotation will be converted; all others will be left unchanged (so everything collected by *args or **kwargs is never modified).

PR #215, opened 2021-03-03
Initialize the default-value to a list if defaultValue is None and the bone is multiple

### LINKED ISSUE #212: baseBone-fallback for a recordBone failing with "'NoneType' object is not iterable" when None

tsteinruecken:
Fixes #212
Replaces #213

### DIFF
--- a/bones/bone.py
+++ b/bones/bone.py
@@ -106,7 +106,7 @@ class baseBone(object):  # One Bone:
 		self.params = params or {}
 		self.multiple = multiple
 		self.indexed = indexed
-		self.defaultValue = defaultValue
+		self.defaultValue = [] if defaultValue is None and self.multiple else defaultValue
 		self.languages = languages
 		self.searchable = searchable
 		if vfunc:
--- a/bones/bone.py
+++ b/bones/bone.py
@@ -106,7 +106,13 @@ class baseBone(object):  # One Bone:
 		self.params = params or {}
 		self.multiple = multiple
 		self.indexed = indexed
-		self.defaultValue = [] if defaultValue is None and self.multiple else defaultValue
+		# Convert a None default-value to the empty container that's expected if the bone is multiple or has languages
+		if defaultValue is None and self.languages:
+			self.defaultValue = {}
+		elif defaultValue is None and self.multiple:
+			self.defaultValue = []
+		else:
+			self.defaultValue = defaultValue
 		self.languages = languages
 		self.searchable = searchable
 		if vfunc:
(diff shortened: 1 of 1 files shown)


bones/bone.py
```diff
@@ -106,7 +106,7 @@ def __init__(self, descr="", defaultValue=None, required=False, params=None, mul
 		self.params = params or {}
 		self.multiple = multiple
 		self.indexed = indexed
-		self.defaultValue = defaultValue
+		self.defaultValue = [] if defaultValue is None and self.multiple else defaultValue
```
phorward, line 109: And what about the cases [here](https://github.com/viur-framework/viur-core/blob/master/bones/bone.py#L397) and [here](https://github.com/viur-framework/viur-core/blob/master/bones/bone.py#L404)? My solution from #213 fixes all of them, and also cleans up the code.

phorward (changes requested):

bones/bone.py
tsteinruecken: Thanks for pointing that out. Ive pushed a new commit that also handles the multi-language case.
With the new skeleton system in ViUR3, it should be guaranteed that a bone has always the correct type expected by the bone (even if being empty) - that's why stringBones initializes to an empty string, numericBones to 0 etc. This bug prevented this on bones with multiple or languages, as these will change the type that's stored in the skeleton.

XeoN-GHMB (approved):

phorward (approved):

sveneberth (approved):

PR #220, opened 2021-03-05
Add support to use a task queue emulator

sveneberth:
This changes allows to run tasks (with retries) in a separate request even on local development server.

# Usage

## Install emulator
I've used https://pypi.org/project/gcloud-tasks-emulator/
```sh
pip install gcloud-tasks-emulator
```

## Start the emulator
```sh
gcloud-tasks-emulator start -p=9999 -t=9090 --queue-yaml=deploy/queue.yaml --queue-yaml-project=$YOUR_PROJECT_NAME --queue-yaml-location=local -r 50
```
Take a look at `gcloud-tasks-emulator start -h` for the meaning of the commands

## Start ViUR
with the environment variable `TASKS_EMULATOR` pointing to the emulator-server.
```sh
dev_appserver.py -A $YOUR_PROJECT_NAME --port=9090 --env_var TASKS_EMULATOR=localhost:9999 deploy
```
(Alternatively you can add this in your app.yaml. However, be careful that you not deploy this, otherwise it will use this server in the cloud as well and will probably fail)

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -58,11 +58,11 @@ if _gaeApp:
 	regionPrefix = _gaeApp.split("~")[0]
 	queueRegion = regionMap.get(regionPrefix)
 
-if not queueRegion and os.getenv("TASKS_EMULATOR") is None:
+if not queueRegion and utils.isLocalDevelopmentServer and os.getenv("TASKS_EMULATOR") is None:
 	# Probably local development server
 	logging.warning("Taskqueue disabled, tasks will run inline!")
 
-if os.getenv("TASKS_EMULATOR") is None:
+if not utils.isLocalDevelopmentServer or os.getenv("TASKS_EMULATOR") is None:
 	taskClient = tasks_v2.CloudTasksClient()
 else:
 	taskClient = tasks_v2.CloudTasksClient(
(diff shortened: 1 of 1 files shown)


tasks.py
```diff
@@ -56,11 +58,17 @@ def jsonDecodeObjectHook(obj):
 	regionPrefix = _gaeApp.split("~")[0]
 	queueRegion = regionMap.get(regionPrefix)
 
-if not queueRegion:
+if not queueRegion and os.getenv("TASKS_EMULATOR") is None:
 	# Probably local development server
 	logging.warning("Taskqueue disabled, tasks will run inline!")
 
-taskClient = tasks_v2.CloudTasksClient()
+if os.getenv("TASKS_EMULATOR") is None:
```
tsteinruecken, line 65: Can you change this condition to 
`if not utils.isLocalDevelopmentServer or os.getenv("TASKS_EMULATOR") is None: `?
This provides an additional safety check if the project is accidentally deployed with the TASKS_EMULATOR env var set in app.yaml

tasks.py
```diff
@@ -188,7 +196,8 @@ def deferred(self, *args, **kwargs):
 		if 'X-AppEngine-TaskName' not in req.headers:
 			logging.critical('Detected an attempted XSRF attack. The header "X-AppEngine-Taskname" was not set.')
 			raise errors.Forbidden()
-		if req.environ.get("HTTP_X_APPENGINE_USER_IP") not in _appengineServiceIPs:
+		if req.environ.get("HTTP_X_APPENGINE_USER_IP") not in _appengineServiceIPs \
```
tsteinruecken, line 199: Same here - don't trust TASKS_EMULATOR alone, check also utils.isLocalDevelopmentServer

PR #216, opened 2021-03-03
Fixed retreiving skeleton structure via view/structure

tsteinruecken:
Fixes #207

PR #223, opened 2021-03-09
Split view/structure into a seperate function

tsteinruecken:
Replaces #216


sveneberth (changes requested):
Looks fine to me. But for final approval, can you please resolve the conflicting files first?

PR #224, opened 2021-03-10
Removed obsolete functions on db.Query

tsteinruecken:


sveneberth (changes requested):
Will there be no replacement?

tsteinruecken:
No. You'll have to inspect query.queries instead (as multiqueries can have different kinds, sort-orders, filters, cursors etc)

PR #228, opened 2021-03-11
Added a RequestValidator to request.py

tsteinruecken:
These validators can be used to abort a request early (before it's being dispatched within ViUR).
The provided FetchMetaDataValidator evaluates the fetch-metadata as recommended by https://web.dev/fetch-metadata/


sveneberth (changes requested):
Can you please resolve the conflicting files first?

PR #247, opened 2021-03-29
Fixed sending the password recovery email

tsteinruecken:


core/modules/user.py
```diff
@@ -233,7 +233,7 @@ def pwrecover(self, authtoken=None, skey=None, *args, **kwargs):
 				60 * 60 * 24,
 				userKey=utils.normalizeKey(user.key),
 				password=skel["password"])
-			utils.sendEMail([userSkel["name"]], self.userModule.passwordRecoveryMail, userSkel)
+			email.sendEMail([userSkel["name"]], self.userModule.passwordRecoveryMail, userSkel)
```
sveneberth, line 236: The parameters are keyword-only, so your fix is still incomplete.
```suggestion
			email.sendEMail(dests=userSkel["name"], tpl=self.userModule.passwordRecoveryMail, skel=userSkel)
```

sveneberth (changes requested):
Can you fix the calls https://github.com/viur-framework/viur-core/blob/21de5399437ec86ad07a14c8f30195f8fcdc4190/core/skeleton.py#L1366 and https://github.com/viur-framework/viur-core/blob/21de5399437ec86ad07a14c8f30195f8fcdc4190/core/skeleton.py#L1279
as well?

tsteinruecken:
Thanks, the first one went somehow lost in my stash. skeleton.py is also fixed.

sveneberth (approved):

PR #256, opened 2021-03-31
Implemented a new password recovery process

tsteinruecken:
Instead of asking for the username, a new password and then sending out a link which will activate the new password,
we'll now ask for the username, send a key in the background if such a user exists, and then have a form bound to the session
that will ask for the recovery code and a new password.


sveneberth (changes requested):
Hey, in general these changes looks fine to me. But I think we can improve the user experience, when we hide the recoveryKey. Therefore I created the PR #275 - See the details there.

Furthermore please add the necessary new and changed templates in the base-project - So people will be able to use this great feature.

sveneberth (approved):
After https://github.com/viur-framework/viur-base/pull/47 is merged, this should work fine.

PR #265, opened 2021-04-08
Updating icon identifiers for the new Vi

phorward:


tsteinruecken:
What is/was the rationale behind changing the icon logic there (dropping paths and file extensions)?
We traditionally had these 3 ways of referencing icons:

- URLs starting with http:// or https:// (embedding from somewhere in the web)
- URLs starting with / (being fetched relative to the current domain (eg /static/icons/xxx)
- Everything else being relative the the icon directory of the Vi/Admin

The first one was (afaik) never used and we can probably drop support for it as CSP-Rules will probably prevent these anyway, but the other two ones are widely used in projects.

tsteinruecken (approved):
Has been discussed on the last ViUR round - will adopt the new icon logic from the VI and port it to the projects/admin

PR #268, opened 2021-04-14
Fixed thumbnail deriver, changed output filename

tsteinruecken:
We now reinitialize the output BytesIO for each file (so the new thumbnail does not get appended to the previous file)
For thumbnails that are scaled to a specific width (instead of a excact resolution) the height is removed from the filename so it can be used without guessing the scaled height.


xnopasaranx (approved):
This is a needed improvement! The filename needs to be predictable in order to use these resized pictures in templates.

PR #271, opened 2021-04-20
WIP: Feature/record bone(multiple=false)

phorward:
This pull requests adds support for recordBones with multiple=False configuration and related Jinja2 form rendering.


xnopasaranx:
I am using this in a customer project where it makes handling billing and shipping address real nice. Please merge!

xnopasaranx (approved):

core/render/html/default.py
```diff
@@ -164,13 +164,18 @@ def renderBoneStructure(self, bone):
 
 		if bone.type == "relational" or bone.type.startswith("relational."):
 			ret.update({
-				"type": bone.type,
```
phorward, line 167: This is already done in the ret = {} dict above, it can be removed.

core/bones/recordBone.py
```diff
@@ -55,7 +55,11 @@ def parseSubfieldsFromClient(self) -> bool:
 
 	def singleValueFromClient(self, value, skel, name, origData):
 		usingSkel = self.using()
-		usingSkel.fromClient(value)
+
```
tsteinruecken, line 58: Nope. Always return all errors. This will suppress warnings and "invalidates other" errors. Error-Handling is done outside (in fromClient).

core/render/html/env/viur.py
```diff
@@ -502,7 +502,7 @@ def shortKey(render, val):
 
 
 @jinjaGlobalFunction
-def renderEditBone(render, skel, boneName):
+def renderEditBone(render, skel, boneName, prefix=None):
```
tsteinruecken, line 505: Also no. For what reason do you need a prefix here? The entire skeleton system has been carefully redesigned to avoid exactly this context-mess (each step/bone now only handles it's own, single value and hasn't have to care about being multiple, translated or inside a relation/recordBone. Please don't re-introduce it.

tsteinruecken (changes requested):
Please don't reintroduce the context-awareness of bones and don't suppress any errors.

core/render/html/env/viur.py
phorward: @tsteinruecken Please look carefully where this prefixing is done: The reason for this is that the renderEditForm/renderEditBone-functions render an edit form recursively. Therefore it is necessary to pull the prefix through the recursion so that correct bone names for ViUR's fromClient()-function are generated. Here's a working example:

This skeleton-compound
```python
class pilotSkel(RelSkel):
	firstname = stringBone(
		descr="Firstname",
		required=True,
		searchable=True
	)

	lastname = stringBone(
		descr="Lastname",
		required=True,
		searchable=True
	)

class flightSkel(Skeleton):
	comment = stringBone(
		descr="Comment",
		searchable=True
	)

	pilot = recordBone(
		descr="Pilot",
		using=pilotSkel,
		format="$(firstname) $(lastname)",
		searchable=True,
		required=True
	)

	copilot = recordBone(
		descr="Copilot",
		using=pilotSkel,
		format="$(firstname) $(lastname)",
		searchable=True,
		required=True
	)
```
results in this an HTML-form (simplified)
```html
<form action="/order/add" method="POST">
	<input name="comment">
	<input name="pilot.firstname">
	<input name="pilot.lastname">
	<input name="copilot.firstname">
	<input name="copilot.lastname">
</form>
```
This is the only position where prefix is used. A version for multiple=True recordBones might also be implemented this way.

phorward: @tsteinruecken if you're not fine with this proposal, please provide a better solution for this in your opinion. Thanks.

tsteinruecken: Hm.. okay, i see your point. The best way would probably to render an special html container element (with classes/data-attributes) around the bones from these record bones and "fix" the name of these input fields with javascript:

- You'll need javascript anyway (eg. for datebones)
- You can support recordBones with multiple=True
- You can correctly handle the required flag (so it's not enforced when you don't fill that bone)

tsteinruecken:
I'd guess it would be better to start with a fresh PR on this? Several non-related issues have been merged into this one; it still contains changes to the deprecated renderEditBone which we'll drop at least in 3.1 and the issue with errors being suppressed is still not addressed?

phorward:
Replaced by #293

PR #275, opened 2021-04-30
Improve password recovery process (UX & naming consistency)

sveneberth:
* Use emailBone for email-inputs (we do this already in all
  login-forms, this can be rendered with input-type=email)
* recoveryKey is now named everywhere recoveryKey
* Use a hidden input for recoveryKey (visible=False), so the user
  doesn't have to care about the recoveryKey
* Load first the values fromClient and then check for POST, this is
  neccessary to load the recoveryKey from the GET-params (when the user
  opens the link .../pwrecovery?recoveryKey=XXX from the mail) into the
  hidden input

This PR is built on #256

### DIFF
--- a/core/modules/user.py
+++ b/core/modules/user.py
@@ -160,11 +160,11 @@ class UserPassword(object):
 		password = passwordBone(descr="Password", indexed=True, params={"justinput": True}, required=True)
 
 	class lostPasswordStep1Skel(RelSkel):
-		name = stringBone(descr="Username", required=True)
+		name = emailBone(descr="Username", required=True)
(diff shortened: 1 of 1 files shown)


tsteinruecken:
emailBone and recoverKey are okay for me; the hidden key and Get/Post-Swap not. The process is bound to the users session - so the idea is to send him the key and he have to copy&paste (or type it) from the received email. We shouldn't make any assumptions about the device the user receives the email on (eg on his smartphone while the recovery process started on his desktop) or any browser boundary (incognito tab, different browsers etc)

tsteinruecken (approved):

PR #276, opened 2021-04-30
Add support to derive Index-module from BasicApplication

### LINKED ISSUE #99: Make index to a BasicApplication or not (Reminder to evaluate)
Currently the index-module mostly inherits from object (and the renderer is assigned as attribute) or from the defaultRenderer.
IMO it's much more proper to inherit the index-module from the BasicApplication and handle it like any other module.
This caused also to handle the index differently than a module. See here: https://github.com/viur-framework/server/pull/191

We have to test was is currently possible on a index-module, and what isn't (deferred-tasks, periodicTasks, ...).
Maybe is was only a problem in ViUR 2...

sveneberth:
When all modules are derived from BasicApplication this a consistent pattern. To be backward compatible, the old style is still supported.

Old style for modules/index.py:
```py
class Index(object):
  def __init__(self, *args, **kwargs):
    self.render = default_render(self)
```

New style:
```py
class Index(BasicApplication):
  pass
```

Additionally I simplify some parts, correct some ugly indentation/breaks, add type annotation and fix some docstrings.

This resolves #99 

Btw: Deferred and period-tasks works now quiet well in the index-module.

### DIFF
--- a/core/__init__.py
+++ b/core/__init__.py
@@ -234,6 +234,8 @@ def setup(modules: Union[object, ModuleType], render: Union[ModuleType, Dict] = 
 			(=> /user instead of /html/user)
 	"""
 	from viur.core.bones import bone
+	# noinspection PyUnresolvedReferences
+	import skeletons  # This import is not used here but _must_ remain to ensure that the
 	# application's data models are explicitly imported at some place!
 	assert projectID in conf["viur.validApplicationIDs"], \
 		"Refusing to start, applicationID %s is not in conf['viur.validApplicationIDs']" % projectID
--- a/core/__init__.py
+++ b/core/__init__.py
@@ -189,7 +189,7 @@ def buildApp(modules: Union[ModuleType, object], renderers: Union[ModuleType, Di
 				# Apply Renderers postProcess Filters
 				if "_postProcessAppObj" in render:
 					render["_postProcessAppObj"](targetResolverRender)
-		if hasattr(moduleName, "seoLanguageMap"):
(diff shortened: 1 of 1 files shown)


core/__init__.py
```diff
@@ -132,71 +133,66 @@ def buildApp(config, renderers, default=None, *args, **kwargs):
...
+					setattr(getattr(root, renderName), moduleName, moduleInstance)
+					targetResolverRender = resolverDict.setdefault(renderName, {})
+				mapModule(moduleInstance, moduleName, targetResolverRender)
 				# Apply Renderers postProcess Filters
-				if "_postProcessAppObj" in rendlist[renderName]:
-					rendlist[renderName]["_postProcessAppObj"](targetResoveRender)
-		if "seoLanguageMap" in dir(moduleClass):
+				if "_postProcessAppObj" in render:
+					render["_postProcessAppObj"](targetResolverRender)
+		if hasattr(moduleName, "seoLanguageMap"):
```
tsteinruecken, line 192: This has to be moduleClass

tsteinruecken (changes requested):
This is close to un-reviewable. Instead of 4 Lines adding the support there are variable renaming, expression rewriting and completely unrelated changes. It took me 1 1/2 Hours to step through this and im still unsure whenever it introduces any subtle bugs in here. Please break PRs like this into multiple PRs in the future (adding functionality / streamlining expressions / variable renaming / unrelated stuff). With the one fix in L192 we can probably merge & test this.

tsteinruecken (approved):

PR #278, opened 2021-05-06
Fix NoneType error when more than one bcc has been pushed into bcc list.

xnopasaranx:
BCC can be a list of several addresses which lead to an error... probably due to my very inelegant fix of the former problem where bcc and cc did not work at all.


xnopasaranx:
this was not an error in the core, closing this PR

PR #284, opened 2021-05-20
Added first support for the dbaccelerator (faster get_multi)

tsteinruecken:
in response to https://github.com/googleapis/python-datastore/issues/150
The original implementation of google.cloud.datastore currently has a major performance regression. We could work around it by adding support for the experimental datastore REST-API wrapper from https://github.com/viur-framework/dbaccelerator and use it for db.get and db.queries if not run inside a transaction


xnopasaranx (dismissed):
This would be great to have as soon as possible. Performance right now is just not excusable and since google won't fix it... I have seen the speed gains this brings on a testing setup.

skoegl (dismissed):

PR #286, opened 2021-06-02
Temp. disable the changelist evaluation in updateRelations

tsteinruecken:
This causes the query to become a list of queries (due to the IN filter) - which currently don't support cursors.
We'll loose a minor optimization (we'll process more entities in updateRelations than we've to) so we should re-enable this as soon we support (internal) cursors on queries with a IN filter.


sveneberth (approved):

sveneberth:
@tsteinruecken Can you please create an issue for this problem so we don't forget it and it is listed as a known issue.

PR #292, opened 2021-06-16
Allow setting a default expiration for downloadUrlFor, added docstrings

tsteinruecken:


phorward (approved):

core/utils.py
```diff
@@ -123,6 +123,16 @@ def hmacVerify(data: Any, signature: str) -> bool:
 
 def downloadUrlFor(folder: str, fileName: str, derived: bool = False,
 				   expires: Union[timedelta, None] = timedelta(hours=1)) -> str:
+	"""
+		Utility function that creates a signed download-url for the given folder/filename combination
+
+		:param folder: The GCS-Folder (= the download-key) for that file
+		:param fileName: The name of that file. Either the original filename as uploaded or the name of a dervived file
+		:param derived: True, if it points to a derived file, False if it points to the original uploaded file
+		:param expires: 0/None if the file is supposed to be public (which causes it to be cached on the google ede
```
sveneberth, line 132: `0` wouldn't be a valid type according to the type annotation `Union[timedelta, None]`. The docstring and type annotations should be conflict-free.

sveneberth (changes requested):

core/utils.py
tsteinruecken: Good point. I've clarified the type-hint (and added some for the jinja2 wrapper for this function which has a slightly different API).

sveneberth (approved):

PR #293, opened 2021-06-17
recordBone(multiple=False) - 3rd attempt

phorward:
This is the third attempt to enable recordBones(multiple=False) again in ViUR3.
It replaces #271.
Please see my comments below.

### DIFF
--- a/core/bones/recordBone.py
+++ b/core/bones/recordBone.py
@@ -56,10 +56,6 @@ class recordBone(baseBone):
 	def singleValueFromClient(self, value, skel, name, origData):
 		usingSkel = self.using()
 		usingSkel.fromClient(value)
-
-		# Discard any errors from non-required bones
-		usingSkel.errors = [err for err in usingSkel.errors if getattr(usingSkel, err.fieldPath[0]).required]
-
 		return usingSkel, usingSkel.errors
 
 	def getSearchTags(self, values, key):
--- a/core/bones/recordBone.py
+++ b/core/bones/recordBone.py
@@ -14,12 +14,12 @@ except ImportError:
 class recordBone(baseBone):
 	type = "record"
 
-	def __init__(self, using, format=None, multiple=True, indexed=False, *args, **kwargs):
-		super(recordBone, self).__init__(multiple=multiple, *args, **kwargs)
+	def __init__(self, using, format=None, indexed=False, *args, **kwargs):
+		super(recordBone, self).__init__(*args, **kwargs)
 		self.using = using
 		self.format = format
-		if not format or indexed or not multiple:
-			raise NotImplementedError("A recordBone must not be indexed, must be multiple and must have a format set")
+		if not format or indexed:
+			raise NotImplementedError("A recordBone must not be indexed and must have a format set")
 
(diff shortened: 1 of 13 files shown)


core/bones/recordBone.py
```diff
@@ -56,6 +56,10 @@ def parseSubfieldsFromClient(self) -> bool:
 	def singleValueFromClient(self, value, skel, name, origData):
 		usingSkel = self.using()
 		usingSkel.fromClient(value)
+
+		# Discard any errors from non-required bones
```
phorward, line 60: When this is not done, ViUR does not accept any edit/add action containing a recordBone that contain an empty, non-required bone.

core/render/html/default.py
```diff
@@ -164,13 +164,18 @@ def renderBoneStructure(self, bone):
 
 		if bone.type == "relational" or bone.type.startswith("relational."):
 			ret.update({
-				"type": bone.type,
```
phorward, line 167: This line is obsolete because type is generally written to ret in line 157.

core/render/html/default.py
```diff
@@ -358,7 +368,7 @@ def add(self, skel, tpl=None, params=None, *args, **kwargs):
 		skel["skey"] = securitykey.create()
 		if currentRequest.get().kwargs.get("nomissing") == "1":
 			if isinstance(skel, SkeletonInstance):
-				super(SkeletonInstance, skel).__setattr__("errors", {})
+				super(SkeletonInstance, skel).__setattr__("errors", [])
```
phorward, line 371: Errors now contain a list, but here it was set to a dict.

core/render/html/default.py
```diff
@@ -398,7 +408,7 @@ def edit(self, skel, tpl=None, params=None, **kwargs):
 
 		if currentRequest.get().kwargs.get("nomissing") == "1":
 			if isinstance(skel, SkeletonInstance):
-				super(SkeletonInstance, skel).__setattr__("errors", {})
+				super(SkeletonInstance, skel).__setattr__("errors", [])
```
phorward, line 411: Errors now contain a list, but here it was set to a dict.

core/bones/recordBone.py
tsteinruecken: I was able to reproduce this. Suppressing the errors in this way would f.e allow you to store any value in an email bone as long it's not required. We'll have to fix this in a different way (probably making bone.fromClient the errors as well as a overall success/failure marker)

phorward: Nope, because the emailBone doesn't take the value provided. It is thrown away in this case. I'm a little confused why there is still an error message for bones which are allowed to be empty. IMHO this is the confusing part, and if this won't be available, I can remove this line and take all real errors which are errors as they are.

tsteinruecken: Yes, pls remove these lines; I'll address the issue with empty non-required fields in a required container next week in a separate pr. The rest of this PR should be resolved.

phorward: OK I removed the discarding from the PR.

tsteinruecken (approved):

PR #298, opened 2021-06-24
Fixed parsing of invalid data supplied from the frontend

tsteinruecken:


core/bones/bone.py
```diff
@@ -257,8 +257,8 @@ def collectRawClientData(self, name, data, multiple, languages, collectSubfields
 							continue
 						fieldSubmitted = True
 						partKey = key.replace(prefix, "")
-						firstKey, remainingKey = partKey.split(".", maxsplit=1)
 						try:
+							firstKey, remainingKey = partKey.split(".", maxsplit=1)
 							firstKey = int(firstKey)
 						except:
```
sveneberth, line 263: Please [do not use bare exception](https://docs.python.org/3.1/howto/doanddont.html#except).

sveneberth (changes requested):

core/bones/bone.py
tsteinruecken: That's fine. Whatever exception raised in there should just skip parsing that row.

sveneberth: No it is not, this is an anti-pattern. Please read the linked section again. It is not recommended even in the official python docs. Just because something is technically possible doesn't mean you should do it.
There is no need to catch a `NameError`, `SyntaxError` or even `SystemExit` here. To catch `ValueError, TypeError, AttributeError` should be enough here. 
If you really mean to catch all exceptions, then at least use `except Exception`, that's a little less bad.

PR #299, opened 2021-06-24
Rename rawValueBone to rawBone. Fixes #259

### LINKED ISSUE #259: Rename rawValueBone to just rawBone
We have a stringBone, a relationalBone, a fileBone... and now a rawValueBone? Why isn't this just be called rawBone, to comply with the already existing naming convention?  Or, let's rename stringBone, relationalBone, fileBone ... into stringValueBone, relationalValueBone and fileValueBone...

The type should be changed to "raw" instread of "rawvalue" as well.

tsteinruecken:


phorward (approved):
:-*

PR #303, opened 2021-06-25
Larger set of documentation updates; Enabled HSTS by default; Reserved "limit" as bonename

tsteinruecken:
This PR is mostly documentation update & additions. It contains two minor non-documentation changes:
- HTST (http strict transport security) is now enabled by default for a full year (we don't support non-tls connections anymore)
- As we changed "amount" to "limit", "limit" is now a reserved bonename (it cannot be used as a bonename anymore)


core/prototypes/basic.py
```diff
@@ -1,29 +1,119 @@
...
 from viur.core import conf
-from viur.core.skeleton import skeletonByKind
+from viur.core.skeleton import skeletonByKind, Skeleton
+from typing import Dict, List, Any, Union, Callable
 
 
 class BasicApplication(object):
 	"""
-	BasicApplication is a generic class serving as the base for the four BasicApplications.
+		BasicApplication is a generic class serving as the base for the four BasicApplications.
```
sveneberth, line 9: The indent was already fine. Please use the recommended docstring format from [PEP-257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings)
> The entire docstring is indented the same as the quotes at its first line

[Here](https://github.com/viur-framework/viur-core/blob/9f8fb2179a61214db04c8564bc5f676ab5c81f52/core/prototypes/list.py#L33-L48) is the indent okay (which means that it's currently totally inconsistent), but the summary is too long (should be one line). Please try to format the docstring you changed properly.

core/skeleton.py
```diff
@@ -31,7 +31,7 @@ class MetaBaseSkel(type):
 	_skelCache = {}  # Mapping kindName -> SkelCls
 	_allSkelClasses = set()  # list of all known skeleton classes (including Ref and Mail-Skels)
 
-	__reservedKeywords_ = {"self", "cursor", "amount", "orderby", "orderdir",
+	__reservedKeywords_ = {"self", "cursor", "amount", "orderby", "orderdir", "limit"
```
sveneberth, line 34: I think this should better replace `"amount"` instead of be a new element?

core/prototypes/list.py
```diff
@@ -37,45 +29,57 @@ def adminInfo(self):
 	def __init__(self, moduleName, modulePath, *args, **kwargs):
 		super(List, self).__init__(moduleName, modulePath, *args, **kwargs)
 
-	def viewSkel(self, *args, **kwargs):
+	def viewSkel(self, *args, **kwargs) -> SkeletonInstance:
 		"""
-		Retrieve a new instance of a :class:`server.skeleton.Skeleton` that is used by the application
+		Retrieve a new instance of a :class:`viur.core.skeleton.Skeleton` that is used by the application
```
sveneberth, line 34: The return type-annotations says `SkeletonInstance`, the docstring says (instance of) `viur.core.skeleton.Skeleton`. Which one is correct?

sveneberth (changes requested):
Some first annotations.

core/prototypes/basic.py
tsteinruecken: That's how it should be indented in core (i don't mind PEP-257) and how the majority of docstrings are indented (just see above in that file)

core/skeleton.py
tsteinruecken: Good point. I think we should unblock amount later on (3.1?) - for now let's leave it there to ensure all projects migrate correctly.

core/prototypes/list.py
tsteinruecken: Fixed. SkeletonInstance is correct (instancing a skeleton-class now gives you an instance of SkeletonInstance and never an instance of that skeleton-class anymore)

core/skeleton.py
sveneberth: I see no reason to keep it blocked. "amount" could not be used before, if you unblock it now nothing will break.
The critical part is that "limit" is now blocked. But that is a project related problem...

core/prototypes/basic.py
sveneberth: It seems you are right that the majority of docstrings are indented. However, the amount that is not indented is not insignificant. Almost all docstrings in the prototypes (current master) are not indented. And in many modules/files it's alternating, for instance: email, ratelimit, utils, requests.

So as we have to reindent many docstrings anyway to have a consistent layout (no matter if with or without indent), I would prefer to remove the indent and follow the PEP.

PR #304, opened 2021-06-25
Ensure the SEO-Key being parsed lower-case

tsteinruecken:


sveneberth (commented):
Looks okay to me. But please take a look at the other side:

I think we should store the `id_or_name` of a key in lowercase as well.
https://github.com/viur-framework/viur-core/blob/d4af8c2c9047a288a6cce88b28680b4c7173c9ce/core/skeleton.py#L814
In case of a `name` this might be mixed-cases, because afaik keys are case-senstive.

For all other cases this should be already stored in lowercase due to this line:
https://github.com/viur-framework/viur-core/blob/d4af8c2c9047a288a6cce88b28680b4c7173c9ce/core/skeleton.py#L770

Correct me, if I'm wrong...

tsteinruecken:
@sveneberth That's correct. However as we made the seo-keys case-IN-sensitive there would be no way to handle entities with mixed cases (eg afaik there could be"test" and "Test" as separate entities) . So the current solution would "correctly" index both - but only the all-lowercase key will be accessible

sveneberth (approved):

PR #305, opened 2021-06-28
Fixed viur-tags search with multiple words, fixed it's sort-order

tsteinruecken:


xnopasaranx:
fixes my search problems! tested this with two word phrases which would not work before, but are correctly queried now.

xnopasaranx (approved):
oops... see my comment above. works on my side.

PR #312, opened 2021-07-16
Added documentation for the db.RunInTransaction function

tsteinruecken:


sveneberth (changes requested):
Can you please mention this changes in the changelog. A general entry like "Improve docstrings/documentation" would be fine for me for this and future changes.

tsteinruecken:
As this is not a technical change i'd suggest to keep it out of the changelog to keep this small & readable (otherwise it's more or less a mirror of git's commit log). But we can vote on this on the next ViUR-Round.

PR #318, opened 2021-08-03
Fixed adminInfo for tree modules without a leafSkel

tsteinruecken:


sveneberth:
Can you please short explain what caused here the `ValueError`? An invalid skelType, raised by this line?
https://github.com/viur-framework/viur-core/blob/ff7d46666160c151aad605c05efa51f2afeaa16a/core/prototypes/tree.py#L92

tsteinruecken:
Correct - trying to resolve a leaf-skel on a tree module that only defines nodes will raise said ValueError

sveneberth (approved):

PR #368, opened 2022-03-02
Implement verbose error page mode

sveneberth:
Display several meta information of the current request, session and environment on the error page for faster debugging.
Can be enabled by setting `conf[viur.debug.verboseErrorHandler"]` to `True`.

I changed the error.css a bit to support larger content and the new content. Because afaik a new design is already in the pipeline, this is only very rudimentary.

Screenshot

### DIFF
--- a/core/request.py
+++ b/core/request.py
@@ -1,18 +1,22 @@
-import traceback, os, unicodedata, typing
-from viur.core.config import conf
-from urllib import parse
-from string import Template
-from io import StringIO
-import webob
-from viur.core import errors
-from urllib.parse import urljoin, urlparse, unquote
-from viur.core.logging import requestLogger, client as loggingClient, requestLoggingRessource
-from viur.core import utils, db
-from viur.core.utils import currentSession, currentLanguage
 import logging
-from time import time
+import os
+import sys
+import traceback
+import typing
+import unicodedata
 from abc import ABC, abstractmethod
+from io import StringIO
+from string import Template
+from time import time
+from urllib import parse
+from urllib.parse import unquote, urljoin, urlparse
 
+import webob
+
+from viur.core import db, errors, utils
+from viur.core.config import conf
+from viur.core.logging import client as loggingClient, requestLogger, requestLoggingRessource
+from viur.core.utils import currentLanguage, currentSession
 
 """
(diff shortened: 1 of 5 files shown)


core/request.py
```diff
@@ -310,48 +315,81 @@ def processRequest(self):
...
-					descr = descr.replace("<", "&lt;").replace(">", "&gt;").replace(" ", "&nbsp;").replace("\n",
-																										   "<br />")
-				res = tpl.safe_substitute(
-					{"error_code": "500", "error_name": "Internal Server Error", "error_descr": descr})
+				with open(conf["viur.errorTemplate"], "r") as fh:
+					tpl = Template(fh.read())
+				debugInformation = ""
+				# We're running on development server or verbose mode is on
+				if self.isDevServer or conf["viur.debug.verboseErrorHandler"]:
+					if not isinstance(e, errors.HTTPException):
```
sveneberth, line 350: @viur-framework/developers What do your think about this. Shall we print a traceback on `HTTPException`s or not? Might be helpful to find the right case, if there are multiple reasons.

tsteinruecken:
I'm undecided about this one. I generally like the idea of a better error page, however i see a few problems with this approach.
1) Remove that config switch. Make this the default for local development and keep it disabled if deployed. Someone will forget to turn this switch back of when deploying - causing probably sensitive information to leak.
2) None of the information added has any additional value for the developer. You really know the state of your session, the parameters supplied, your python version and so on (or can very easily retrieve these information from your browser / environment etc). But what we could do is collecting all the logs that have been generated by that request that lead to the exception (at least the ViUR ones) - because this information is usually swamped out by all the logs the dev_appserver / google api calls generated.
3) Ignore httpErrors - unless traceExceptions is set. That flag is explicitly for debugging these.

phorward:
Hello @sveneberth, I'm in the opinion this feature is useful and should be integrated somehow. Can you please update your pull request to current main, or provide a similar solution in viur-base?

phorward (changes requested):
Please see previous comment.

phorward:
As there is no more interaction here, the ticket will be closed.
It cannot be merged in this way, but a verbose and nice error reporting would be great.
Please see my [comments](https://github.com/viur-framework/viur-core/pull/368#issuecomment-1110809499) on this in here and create a different approach based on the work done here. Thanks.

PR #375, opened 2022-03-15
added PyPI packaging

XeoN-GHMB:
- adds a pipenv to build releases with `pipenv run python -m build`
- adds coreBasePath to utils
- allow submodule and package useage
- inlined style and script for error_template and google_login and secured with extendCsp
- added github action for autobuild and publish on pypi, test.pypi and as github release
- updated gitignore
- removed all assets in resources


tsteinruecken (commented):
I would recommend dropping the build&publish workflow.
We went great lengths in removing pypi from the list of entities we need to trust to run our application securely by providing secure hashes for each and every dependency fetched from there. Now providing github with the credentials to publish new packages there doesn't seem like a good idea to me.
Otherwise fine.

PR #377, opened 2022-03-17
Cure/rel skel.unserialize()

phorward:
This is also stated in the docstring before, but didn't work as subsequent calls access properties not implemented in dict.


tsteinruecken:
The better solution would be probably to fix the docstring instead (as it would be a cleaner api and you cannot set skel["key"] without supplying a db.Entity)

tsteinruecken (approved):
As discussed today in the ViUR Meeting

PR #378, opened 2022-03-17
Implement baseSkel for all module prototypes

phorward:
`baseSkel()` returns an instance of an unmodified base skeleton for any module. It is used by more-specific skeleton retrival functions like `addSkel()`, `editSkel()` and `viewSkel()`.


tsteinruecken:
In my opinion, this is superfluous. Yes, you have to call __resolveSkelCls("node")() or the like, but giving you back an class is intentional. There are class-level functions (like subSkel) which (in my view) should be reachable without creating an instance first (which would get immediately discarded)

phorward:
@tsteinruecken to reduce the massiv amount of stack usage caused by the baseSkel function, I've updated this PR in [d19673f](https://github.com/viur-framework/viur-core/pull/378/commits/d19673ffec8f259768f514a61f56907135d9dccb) so that addSkel, editSkel and viewSkel are using _resolveSkelCls again, but baseSkel can be used or overridden as well.

phorward:
@tsteinruecken I've now pushed the changes back to the previous version, as discussed privately. I've also fixed some docstrings.

tsteinruecken (approved):

PR #382, opened 2022-03-17
Fix file downloads with a provided filename parameter

phorward:


tsteinruecken:
Alternative is in #394

tsteinruecken:
As discussed, we'll go with #394 for now

PR #383, opened 2022-03-17
Set format string to established notation

phorward:


tsteinruecken (changes requested):
This is scheduled to be discussed between 28.03 -> 01.04

phorward:
Hello, this pull request is a quick fix for current use-cases. The entire format string topic was already brainstormed in the ViUR week two weeks ago, but more ideas came up where a task force will be established.

phorward:
Substituted by #436

PR #385, opened 2022-03-17
Replace CustomJsonEncoder by viur.json module

phorward:
In custom project implementations, any calls to json.dumps() must be replaced by CustomJsonEncoder from viur.core.render.json.default to use the CustromJsonEncoder.

This change just requires to import viur.json instead of just json to provide a simular behavior and reduce amount of code changes in custom projects.

### DIFF
--- a/core/json.py
+++ b/core/json.py
@@ -1,0 +1,32 @@
+"""
+JSON module replacement for ViUR
+"""
+import json, typing, datetime
+
+from viur.core import db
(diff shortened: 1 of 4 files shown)


tsteinruecken (changes requested):
Please be aware that there is no such thing as *the* json module here.
For example, tasks.py has it's own encode/decode hooks.

PR #387, opened 2022-03-17
Raise exception when invalid objects are assigned to multiple=True bones

phorward:
This...
```python
b = stringBone(multiple=True)

skel["b"] = "test"
```
...will write `['t', 'e', 's', 't']` into the bone in the datastore.

This fix will write `["test"]`, as wanted.


phorward:
Hello @tsteinruecken, as requested I've updated the pull request to not correct values, but throw an Exception in case invalid data was supplied.

tsteinruecken (approved):

PR #388, opened 2022-03-17
Fix buildDBFilter rawFilter condition

phorward:
In case `rawFilter[name].lower() == "none"` and rawFilter[name] is something else than str, ViUR will crash here.


tsteinruecken:
Okay - i've checked this, this pr is fine. It's only the special case for filtering for unset relations - a list does not make sense here.

tsteinruecken (approved):

PR #390, opened 2022-03-17
selectBone() values accept for list, tuple or callable

phorward:
- list and tuple are converted to dict
- callable allows to obtain values on demand by a callback


tsteinruecken (changes requested):
If values are given as a list, we might want a naming-scheme and create the description for each key-value pair on the fly (especial useful if we translate these). This way we can just supply a list of keys, and have the human readable description be generated automatically.

phorward:
Hello @tsteinruecken, please re-check the PR as I updated the automatic translation of values.

phorward:
Hello @tsteinruecken, as requested I've updated the pull request to not correct values, but throw an Exception in case invalid data was supplied.

tsteinruecken:
Thanks for the update. I would recommend introducing a naming-scheme for these values (maybe prepending kindname-bonename before the value?) However, we would hook/adapt setSystemInitialized for this.

phorward:
> Thanks for the update. I would recommend introducing a naming-scheme for these values (maybe prepending kindname-bonename before the value?) However, we would hook/adapt setSystemInitialized for this.

Can you please provide a proposal for this? IMHO this feature is used for rapid prototyping data models. For example, in some cases, you only have tags in form of a selectBone, where the key is equal to the value, and no other or special meaning for the value is required or available, as the key stands on its own.

If you have further ideas on this, please make a proposal, probably in a separate pull request.

phorward:
I just made a commit on this PR solving a bug in the default userSkel's `access`-bone that came in with the new modification of the selectBone. It makes the use of extendAccessRights in the user module obsolete.

PR #401, opened 2022-03-23
Fixed unique=True on multiple=True relationalBones

tsteinruecken:


core/bones/relationalBone.py
```diff
@@ -112,7 +112,7 @@ def __init__(self, kind: str = None, module: Optional[str] = None, refKeys: Opti
 					CascadeDeletion set, and B references C also with CascadeDeletion; if C gets deleted, both B and A
 					will be deleted as well.
 		"""
-		baseBone.__init__(self, *args, **kwargs)
+		baseBone.__init__(self, multiple=multiple, *args, **kwargs)
 		self.multiple = multiple
```
phorward, line 116: I think `self.multiple = multiple` is obsolete here,  as this is set [in baseBone.__init__() already](https://github.com/viur-framework/viur-core/blob/master/core/bones/bone.py#L109) for self.

phorward (changes requested):

sveneberth (approved):

PR #404, opened 2022-03-30
Improve SEO url integration

sveneberth:
* Correct the type annotation of the `@exposed` decorator
* Eliminate duplicate by calling `list()` instead
* Add docstrings
* Redirect to the current seo-key if the site was opened with an old one

### DIFF
--- a/core/__init__.py
+++ b/core/__init__.py
@@ -308,7 +308,7 @@ def forcePost(f: Callable) -> Callable:
 	return f
 
 
-def exposed(f: Callable) -> Callable:
+def exposed(f: Union[Callable, dict]) -> Callable:
 	"""
 		Decorator, which marks an function as exposed.
 
--- a/core/prototypes/list.py
+++ b/core/prototypes/list.py
@@ -3,7 +3,7 @@ from viur.core import utils, errors, conf, securitykey, db
 from viur.core import forcePost, forceSSL, exposed, internalExposed
 from viur.core.skeleton import SkeletonInstance
 from viur.core.prototypes import BasicApplication
-from viur.core.utils import currentRequest
+from viur.core.utils import currentRequest, currentLanguage
 from viur.core.cache import flushCache
 
 import logging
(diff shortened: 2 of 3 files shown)


core/prototypes/list.py
```diff
@@ -313,29 +313,35 @@ def index(self, *args, **kwargs):
...
 				if not self.canView(skel):
 					raise errors.Forbidden()
+				lang = currentLanguage.get()
+				# Check whether this is the current seo-key, otherwise redirect to it
+				if (
+					skel["viurCurrentSeoKeys"]
+					and (currentSeoKey := skel["viurCurrentSeoKeys"].get(lang))
+					and seoKey != currentSeoKey
+				):
+					modulePath = self.modulePath
```
tsteinruecken, line 335: You should use the function seoUrlToEntry from https://github.com/viur-framework/viur-core/blob/master/core/utils.py#L153

sveneberth: I didn't know this method, good to know. Thank you for this info.

tsteinruecken (approved):

core/__init__.py
```diff
@@ -308,7 +308,7 @@ def forcePost(f: Callable) -> Callable:
 	return f
 
 
-def exposed(f: Callable) -> Callable:
+def exposed(f: Union[Callable, dict]) -> Callable:
```
phorward, line 311: @sveneberth can you please shortly describe this change?

phorward (approved):

core/__init__.py
sveneberth: The `exposed` method is usable in the common way as decorator (which will be called with your method as `Callable` argument). Or it's callable with a `dict` as argument to provide a translated/seo-friendly name for this method (and generates a new method (`exposeWithTranslations`) to wrap your method).

Example: 
```py
class MyModule(List):
	@exposed({
		"de": "aktuelle-Uhrzeit",
		"en": "current-time",
	})
	def currenttime(self):
		return f"It's {datetime.datetime.now().isoformat(' ', 'seconds')}"

# On the German site:
# GET /de/mymodule/aktuelle-Uhrzeit

# On the English site:
# GET /en/mymodule/current-time
```

I want to write a detailed site/example in our docs about the SEO-features in the core - Maybe on the next OpenSourceFriday :)

sveneberth: Yes, it would be nice to have more parameters, e.g. to restrict access for this method by access levels (as far as i remember, we want a logic for this which is non-public by default (only root-users) and additional users have to be explicitly permitted. 

But this doesn't really have to do with the PR... Maybe we should create an issue to collect some ideas or discuss this. :)

phorward: > Maybe we should create an issue to collect some ideas or discuss this. :)

I've opened #412 on this, please add further suggestions there.

PR #406, opened 2022-04-07
Fix DbAccessLog in Cache Module

ArneGudermann:
Hello, this pull request fixes DbAccessLog in the cache module.


tsteinruecken:
This fix is invalid. It will break nesting of cache-entries. Furhtermore, this is still broken in viur-datastore; i'll take care of it.

PR #409, opened 2022-04-08
Logical mistake in an assert expression 

co-dev10:


core/bones/bone.py
```diff
@@ -418,7 +418,7 @@ def serialize(self, skel: 'SkeletonInstance', name: str, parentIndexed: bool) ->
 			elif self.multiple:
 				res = []
 
-				assert newVal is None or isinstance(newVal, (list, tuple)), \
+				assert newVal is None or not isinstance(newVal, (list, tuple)), \
```
sveneberth, line 421: Hey @co-dev10, can you please explain your intention behind the change? I think this assertion was already correct.
With your negation you accept any object except `list` and `tuple`. So even non-iterable objects, which would fail in the next line https://github.com/viur-framework/viur-core/blob/98d5b3716ebd2e923acd7fdfa96c13f8a01a610e/core/bones/bone.py#L424

PR #424, opened 2022-04-22
File.write() to write file from buffer

phorward:
This function is intended to be used by server-side code to write a file from any bytes-like object or buffer.

### DIFF
--- a/core/modules/file.py
+++ b/core/modules/file.py
@@ -9,7 +9,7 @@ from base64 import urlsafe_b64decode
 from datetime import datetime, timedelta
 from io import BytesIO
 from quopri import decodestring
-from typing import Dict, Tuple, Union, List
+from typing import Any, Dict, Tuple, Union, List
 from urllib.request import urlopen
 from viur.core.config import conf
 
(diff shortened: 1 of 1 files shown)


core/modules/file.py
```diff
@@ -239,6 +239,35 @@ class File(Tree):
...
+		blob.upload_from_file(BytesIO(content), content_type=mimetype)
+
+		skel = self.addSkel("leaf")
+		skel["name"] = filename
+		skel["size"] = blob.size
+		skel["mimetype"] = mimetype
+		skel["dlkey"] = dl_key
+		skel["parentdir"] = None
+		skel["parententry"] = None
+		skel["pending"] = False
```
sveneberth, line 264: These are already the `defaultValue`s and can be omitted to save the setitem-calls.
```suggestion
```

core/modules/file.py
```diff
@@ -239,6 +239,35 @@ class File(Tree):
...
+		skel["weak"] = True
+		skel["width"] = 0
+		skel["height"] = 0
```
sveneberth, line 267: This is wrong. If you set this to `0` it means "it's has a value" (which is 0), but this isn't determined correctly according to the content here. 
Height and width are currently only used for images in the core. But you can implement this for different file-types as well. For instance the height of a text-file could be the amount of lines and the width the amount of columns, for a pdf you can use the dimension of the page(s). Therefor this attributes should be flexible and always contain a reliable int (or nothing).

My suggestion: Extend the signature with kwargs like `height: Optional[int] = None, width: Optional[int] = None`. Then you don't have a value by default (= `None`) and you can pass a correct value with the `File.write()` call.


```suggestion
		skel["width"] = width   # from the signature 
		skel["height"] = height   # from the signature
```

core/modules/file.py
```diff
@@ -239,6 +239,35 @@ class File(Tree):
 
 	blobCacheTime = 60 * 60 * 24  # Requests to file/download will be served with cache-control: public, max-age=blobCacheTime if set
 
+	def write(self, filename: str, content: Any, mimetype: str = "text/plain"):
```
sveneberth, line 242: For the sake of completeness add the return type.
```suggestion
	def write(self, filename: str, content: Any, mimetype: str = "text/plain") -> db.Key:
```

sveneberth (changes requested):

core/modules/file.py
phorward: You're right, but there are plenty of other locations where this is done likely. I think this has to be done in a separate pull request.

core/modules/file.py
phorward:
```suggestion
	def write(self, filename: str, content: Any, mimetype: str = "text/plain", width: int = 0, height: int = 0) -> db.Key:
```

core/modules/file.py
```diff
@@ -239,6 +239,34 @@ class File(Tree):
 
 	blobCacheTime = 60 * 60 * 24  # Requests to file/download will be served with cache-control: public, max-age=blobCacheTime if set
 
+	def write(self, filename: str, content: Any, mimetype: str = "text/plain", width: int = 0, height: int = 0):
```
sveneberth, line 242: Please use `None` as explained in [this comment](https://github.com/viur-framework/viur-core/pull/424#discussion_r856369056) and add the return type.

```suggestion
	def write(self, filename: str, content: Any, mimetype: str = "text/plain", width: int = None, height: int = None) -> db.Key:
```

sveneberth (changes requested):

tsteinruecken (approved):

PR #426, opened 2022-04-26
Refactoring all bone-related __init__ functions

phorward:
- Don't allow for *args anymore
- Error-report any unknown kwargs not intercepted before
- Sorted parameters alphabetically (baseBone)

This includes an improvement likewise #421 for textBone as well.

### DIFF
--- a/core/bones/bone.py
+++ b/core/bones/bone.py
@@ -6,7 +6,7 @@ import hashlib
 import copy
 from enum import Enum
 from dataclasses import dataclass, field
-from typing import Union, List, Set, Any
+from typing import Union, Dict, List, Set, Any
 
 __systemIsIntitialized_ = False
 
@@ -60,13 +60,26 @@ class MultipleConstraints:  # Used to define constraints on multiple bones
 
 
 class baseBone(object):  # One Bone:
-	hasDBField = True
 	type = "hidden"
 	isClonedInstance = False
 
-	def __init__(self, descr="", defaultValue=None, required=False, params=None, multiple=False, indexed=True,
-				 languages=None, searchable=False, vfunc=None, readOnly=False, visible=True, unique=False,
-				 isEmptyFunc=None, getEmtpyValueFunc=None, **kwargs):
+	def __init__(
+		self,
+		defaultValue: Any = None,
+		descr: str = "",
+		getEmtpyValueFunc=None,
+		indexed: bool = True,
+		isEmptyFunc=None,
+		languages: Union[None, List[str]] = None,
+		multiple: bool = False,
+		params: Dict = None,
+		readOnly: bool = False,
+		required: bool = False,
+		searchable: bool = False,
+		unique: Union[None, UniqueValue] = None,
+		vfunc: callable = None,  # todo: rename to verify_function?
+		visible: bool = True,
+	):
 		"""
 			Initializes a new Bone.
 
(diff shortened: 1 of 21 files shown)


core/bones/bone.py
```diff
@@ -60,13 +60,25 @@ class MultipleConstraints:  # Used to define constraints on multiple bones
 
 
 class baseBone(object):  # One Bone:
-	hasDBField = True
```
phorward, line 63: This is unused.

core/bones/bone.py
```diff
@@ -60,13 +60,25 @@ class MultipleConstraints:  # Used to define constraints on multiple bones
...
 	type = "hidden"
 	isClonedInstance = False
 
-	def __init__(self, descr="", defaultValue=None, required=False, params=None, multiple=False, indexed=True,
-				 languages=None, searchable=False, vfunc=None, readOnly=False, visible=True, unique=False,
-				 isEmptyFunc=None, getEmtpyValueFunc=None, **kwargs):
```
phorward, line 69: "getEmtpyValueFunc" :-D unused... it was too emtpy...

core/prototypes/tree.py
```diff
@@ -17,7 +17,7 @@
 class TreeSkel(Skeleton):
 	parententry = keyBone(descr="Parent", visible=False, indexed=True, readOnly=True)
 	parentrepo = keyBone(descr="BaseRepo", visible=False, indexed=True, readOnly=True)
-	sortindex = numericBone(descr="SortIndex", mode="float", visible=False, indexed=True, readOnly=True, max=pow(2, 30))
```
phorward, line 20: This is the reason why we need a SortIndexBone.... GRRRR

tsteinruecken:
Would be okay for me. However, be aware that this will break things. At least i relay in my projects on descr being the first args parameter. These would have to be adapted.

core/bones/keyBone.py
```diff
@@ -8,8 +8,8 @@
 class keyBone(baseBone):
 	type = "key"
 
-	def __init__(self, descr="Key", readOnly=True, visible=False, **kwargs):
-		super(keyBone, self).__init__(descr=descr, readOnly=True, visible=visible, defaultValue=None, **kwargs)
+	def __init__(self, *, descr="Key", readOnly=True, visible=False, **kwargs):
+		super(keyBone, self).__init__(descr=descr, readOnly=readOnly, visible=visible, defaultValue=None, **kwargs)
```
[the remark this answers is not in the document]
phorward: @tsteinruecken OK thanks. I think that only very few projects use args-notation for bones. Not even my older projects use this feature, so it might be a "small breaking change", not requiring a new minor release.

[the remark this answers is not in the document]
phorward: Hello @tsteinruecken, this is correct, but might be a wanted behavior. According to https://github.com/viur-framework/viur-core/blob/main/core/skeleton.py#L580, a Skeleton's standard key is predefined as readOnly=True, and this is also the case with the keyBone's defaultValue.

Anyway, developers should not be limited to explicitly allow setting the readOnly behavior of a keyBone on their own risk.

phorward:
@tsteinruecken OK thanks. I think that only very few projects use args-notation for bones. Not even my older projects use this feature, so it might be a "small breaking change", not requiring a new minor release.

I've fixed numericBone to allow for mode-parameter again, as deprecation warning (this will be breaking to some projects).
With the keyBone issue, I'm undetermined. The behavior was this way in ViUR2, and was changed in commit 2871460e4fdcad1d48d74b5da564b5f7dc8484bf, which some features in a project I'm currently porting from ViUR2 to ViUR3 as well. Surely I can use a rawBone for this, but this would miss out the automatic conversion and rendering of keys in a keyBone.

tsteinruecken:
Yes - ive missed that in the transition from ViuR1->ViUR2. In V1, the skeleton had a hidden __currentDbKey_ assigning to .id would not have done much. I can see valid use-cases for this - just be aware that if you ever set the key on skeleton/relskels to writeable, bad things will happen.

PR #427, opened 2022-04-26
Removing any generageSearchWidget functions

phorward:
I think we "generaged" enough of these widgets since VIUR 1.0 now...


tsteinruecken (approved):
Not sure if any project still depends on these, however they're probably broken in the front-ends anyway. However, we should improve the Vi so you can filter by arbitrary fields (if supported by available indexes)

PR #432, opened 2022-05-11
Implement unittests

sveneberth:
* Implement the first unittests which are runable in a mocked enviroment.
* Add a GitHub-Action workflow to automate the test suite execution and
  deploy the coverage html report to a branches for GitHub-Pages


phorward (approved):
Hello, thanks for this pull request.
I'm fine with this first approach.

phorward:
#434 fixes the getEmptyValueFunc issue, can you please re-enable it in your tests?

XeoN-GHMB (approved):

tsteinruecken:
I think this is the wrong approach. With mocking everything and then randomly calling ViUR-functions in a faked environment we might be able to test some simple utility functions (like isEmpty) - but never the important bits (Does a bone loose it's value when saved? Does Unique-Locking work? Are Access-Checks performed?). I think, we'll need to fully initialize the ViUR System (including access to a datastore/cloud task Project or Emulator) and start from there. This will also ensure that some internal machinery (like setSystemInitialized or translations) is in a valid state.

phorward:
> I think this is the wrong approach.

No, this is a valid beginning to finally bring any form of test suite into viur-core.

> I think, we'll need to fully initialize the ViUR System (including access to a datastore/cloud task Project or Emulator) and start from there. This will also ensure that some internal machinery (like setSystemInitialized or translations) is in a valid state.

Sure, please open a separate pull request with the desired functionality and a set of test cases. Of course, it would be nice to simply extend this attempt and cherish a single code base.

xnopasaranx (approved):

sveneberth:
@tsteinruecken There're different types of testing. These tests are currently only unit (component) tests. They test a small piece of code, initially a few bone methods. 
Of course it doesn't makes sense to test the whole core only on this level. Later, there should also be integration tests ("Do our bones work inside a Skeleton?") and system tests (cf. [Wikipedia/Software testing
/Testing_levels](https://en.wikipedia.org/wiki/Software_testing#Testing_levels)). For the latter and be able to do realistic tests, we need a gcloud environment, as you mentioned correctly. But this requires some more time to create a suitable test suite.
Therefore, let's start simple. And anything is better than nothing :)

sveneberth:
@phorward I just saw that you changed the base branch. Please update the _develop_ from the _main_ before merging. I have already synced my branch. Otherwise you squash the other changes in this PR and destroy the history.

phorward:
> @phorward I just saw that you changed the base branch. Please update the _develop_ from the _main_ before merging. I have already synced my branch. Otherwise you squash the other changes in this PR and destroy the history. 

I've done this, thanks for pointing this out.

sveneberth:
> > @phorward I just saw that you changed the base branch. Please update the _develop_ from the _main_ before merging. I have already synced my branch. Otherwise you squash the other changes in this PR and destroy the history.
> 
> I've done this, thanks for pointing this out. 

GitHub didn't get this... The commit message 5e8f873 is wrong...

PR #435, opened 2022-05-12
PEP8-compliant naming of Bone classes

phorward:
This commit

- renamed all bones to follow a PEP-8 compliant naming convention
- renames the source files to shorter and PEP-8 conforming name convention
- fixes and optimizes all bone imports
- generates a class emitting a deprecation warning for any old naming

This change is breaking to most VIUR projects, at least some imports need to be touched.


phorward:
Hello, I've tested this branch with two major projects in ViUR3 without any problems. Needed to fix two imports in one case.

tsteinruecken (approved):
As said - this is fine for me. However, i'd still suggest to do all breaking changes for cosmetic reasons (PEP8 etc) now, instead of spreading them over multiple releases and forcing projects to constantly adopt.

PR #438, opened 2022-05-13
Remove the unused cursor parameter in doClearSKeys

sveneberth:


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ This file documents any relevant changes done to ViUR-core since version 3.0.0.
 ### Changed
 
 ### Fixed
+- Set the cursor of the next chunk on the query in `securitykey.doClearSKeys()`
```
phorward, line 13: Hello, thanks for doing this, but in future we gonna generate Changelogs from the git history, which is nice due the squashed commits.

core/securitykey.py
```diff
@@ -99,8 +99,10 @@ def startClearSKeys():
 
 
 @callDeferred
-def doClearSKeys(timeStamp, cursor):
+def doClearSKeys(timeStamp: str, cursor: Optional[str]) -> None:
 	query = db.Query(securityKeyKindName).filter("until <", datetime.strptime(timeStamp, "%d.%m.%Y %H:%M:%S"))
+	if cursor is not None:
```
phorward, line 104: you can call setCursor with a None-value as will, with the same effect.

phorward (changes requested):
Hello @sveneberth, thanks for this PR. Can you please check the indentation, it seems there is a mess-up with tabs and spaces. Turn it into tabs first, we convert them later in a separate PR.

tsteinruecken:
I'd suggest removing the cursor parameter altogether. We delete the entities we hit - so there's no need to advance the query with a cursor.

phorward:
> I'd suggest removing the cursor parameter altogether. We delete the entities we hit - so there's no need to advance the query with a cursor.

I'm complying with this suggestion, please remove the cursor entirely.

sveneberth:
> Hello @sveneberth, thanks for this PR. Can you please check the indentation, it seems there is a mess-up with tabs and spaces. Turn it into tabs first, we convert them later in a separate PR.

I don't see any indentation issues?!

sveneberth:
> I'd suggest removing the cursor parameter altogether. We delete the entities we hit - so there's no need to advance the query with a cursor.

Done

core/securitykey.py
```diff
@@ -99,10 +99,9 @@ def startClearSKeys():
 
 
 @callDeferred
-def doClearSKeys(timeStamp, cursor):
+def doClearSKeys(timeStamp: str) -> None:
 	query = db.Query(securityKeyKindName).filter("until <", datetime.strptime(timeStamp, "%d.%m.%Y %H:%M:%S"))
 	for oldKey in query.run(100, keysOnly=True):
```
phorward, line 104: keysOnly=True doesn't have any effect in viur-datastore, and is just thrown away by kwargs.
```suggestion
	for oldKey in query.run(100):
```

phorward (changes requested):
I still found one more little issue. I think we generally should fix this in viur-datastore and remove all keysOnly-parameters which don't have any effect (there are some places). I've opened issue https://github.com/viur-framework/viur-datastore/pull/6 to fix this as well.

phorward (approved):

PR #446, opened 2022-05-19
Providing a SortIndexBone

phorward:
The SortIndexBone is a pre-defined NumericBone with a precision of 8, max value of pow(2, 30) and a defaultValue of the current timestamp.

### DIFF
--- a/core/bones/sortindex.py
+++ b/core/bones/sortindex.py
@@ -1,0 +1,23 @@
+from viur.core.bones.numeric import NumericBone
+import time
+
+
+class SortIndexBone(NumericBone):
+	type = "numeric.sortindex"
+
+	def __init__(
+		self,
+	    *,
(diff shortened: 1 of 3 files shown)


core/bones/sortindex.py
```diff
@@ -0,0 +1,23 @@
...
+class SortIndexBone(NumericBone):
+	type = "numeric.sortindex"
+
+	def __init__(
+		self,
+	    *,
+		defaultValue=lambda: time.time(),
+		descr="SortIndex",
+		max=pow(2, 30),
+		precision=8,
```
sveneberth, line 14: Type annotations would be great.

sveneberth (approved):

core/bones/sortindex.py
phorward: Indeed.

PR #449, opened 2022-05-20
Remove keysOnly argument and unused cursor

sveneberth:
Remove keysOnly argument from some queries (related to https://github.com/viur-framework/viur-datastore/pull/6) and remove unused cursor (like in #438)


phorward (approved):
Thanks for cleaning this up!

PR #450, opened 2022-05-23
Update viur.py downloadUrlFor wrapper method. fileName fix

xnopasaranx:
Untested, have to test this myself first and will report. Since right now it is not possible to pass a custom filename for a downloadable file, I would think this should work. also the filename parameter is apparently set by default, but it's not working for me. have to investigate why.


xnopasaranx:
now tested and should theoretically make setting a custom fileName possible. But the name is mangled by the downloadUrlFor in utils... not sure how to get around that.

xnopasaranx:
PR can be ignored, the solution has to be done on a different level and I will make another PR for that. Closing this one now

PR #452, opened 2022-05-24
Flip imports in modules/user as the email-bone would now shadow the email module

tsteinruecken:


XeoN-GHMB (approved):

phorward (approved):

sveneberth:
I don't think this is a good solution. If you reformat the code with _optimize imports_, the order becomes wrong again due to the lexicographic order.

We use `email` in [l. 454](https://github.com/viur-framework/viur-core/blob/2dcc7e3fc4e8add4b060f6b4de19d0526067c23c/core/modules/user.py#L454) also as variable name and use only two methods from the `email` module. 
Therefore I recommend to replace `from viur.core import email` with `from viur.core.email import sendEMail, sendEMailToAdmins` and adjust the method calls.

PR #454, opened 2022-05-30
Fixing file module

phorward:
Due the renaming of bones.stringBone into bones.string, the stdlib `import string` and then `from viur.core.bones import *` afterwards causes the assert with string.ascii_digits etc. to fail.

This fixes it entirely and cleans up Skeleton definition of the file module generally (formatting, index=True removal)

### DIFF
--- a/core/bones/__init__.py
+++ b/core/bones/__init__.py
@@ -1,30 +1,37 @@
-from viur.core.bones.base import BaseBone, ReadFromClientError, UniqueValue, UniqueLockMethod, MultipleConstraints
-from viur.core.bones.boolean import BooleanBone
-from viur.core.bones.captcha import CaptchaBone
-from viur.core.bones.color import ColorBone
-from viur.core.bones.credential import CredentialBone
-from viur.core.bones.date import DateBone
-from viur.core.bones.email import EmailBone
-from viur.core.bones.file import FileBone
-from viur.core.bones.key import KeyBone
-from viur.core.bones.numeric import NumericBone
-from viur.core.bones.password import PasswordBone
-from viur.core.bones.randomslice import RandomSliceBone
-from viur.core.bones.raw import RawBone
-from viur.core.bones.record import RecordBone
-from viur.core.bones.relational import RelationalBone, RelationalConsistency
-from viur.core.bones.selectcountry import SelectCountryBone
-from viur.core.bones.select import SelectBone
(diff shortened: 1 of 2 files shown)


Xeon2003 (approved):

XeoN-GHMB (approved):

sveneberth (changes requested):
Instead of importing the _bones_ module, you could just import the bones needed. This would only be a Oneliner-fix and doesn't require the attribute access `bones.XYZBone`.

```py
from from viur.core.bones import BaseBone, BooleanBone, KeyBone, NumericBone, StringBone
```

phorward:
> Instead of importing the _bones_ module, you could just import the bones needed. This would only be a Oneliner-fix and doesn't require the attribute access `bones.XYZBone`.
> 
> ```python
> from viur.core.bones import BaseBone, BooleanBone, KeyBone, NumericBone, StringBone
> ```

You're right as well. Next to your change, I've added another proposal, please review. Both should work.

core/bones/__init__.py
```diff
@@ -1,37 +1,47 @@
...
-from viur.core.bones.spatial import SpatialBone
-from viur.core.bones.string import StringBone
-from viur.core.bones.text import TextBone
-from viur.core.bones.treeleaf import TreeLeafBone
-from viur.core.bones.treenode import TreeNodeBone
-from viur.core.bones.user import UserBone
-
-
-# Dynamically create a class providing a deprecation logging message for every lower-case bone name
+from .base import BaseBone, ReadFromClientError, UniqueValue, UniqueLockMethod, MultipleConstraints
```
sveneberth, line 1:
```suggestion
from .base import BaseBone, ReadFromClientError, ReadFromClientErrorSeverity, UniqueValue, UniqueLockMethod, MultipleConstraints
```

core/bones/__init__.py
```diff
@@ -1,37 +1,47 @@
...
+from .spatial import SpatialBone
+from .string import StringBone
+from .text import TextBone
+from .treeleaf import TreeLeafBone
+from .treenode import TreeNodeBone
+from .user import UserBone
+
+# Expose only specific names
+__all = [
+	"ReadFromClientError",
```
sveneberth, line 28:
```suggestion
	"ReadFromClientError",
	"ReadFromClientErrorSeverity",
```

sveneberth (changes requested):

PR #455, opened 2022-05-31
Added ratelimit to login with username/password

tsteinruecken:


core/modules/user.py
```diff
@@ -181,6 +183,9 @@ def login(self, name=None, password=None, skey="", *args, **kwargs):
 		if not name or not password or not securitykey.validate(skey, useSessionKey=True):
 			return self.userModule.render.login(self.loginSkel())
 
+		if not self.loginRateLimit.isQuotaAvailable():
+			raise errors.Forbidden()
```
sveneberth, line 187: Can we raise a _429 Too Many Requests_ here and add the `Retry-After` header? This status code is more common and specified for this use case.

cf. [RFC 6585 Additional HTTP Status Codes](https://datatracker.ietf.org/doc/html/rfc6585#section-4), [MDN Docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429).

sveneberth (changes requested):

core/modules/user.py
tsteinruecken: That's consistent with other usages of the ratelimit module in ViUR (eg. https://github.com/viur-framework/viur-core/blob/main/core/modules/user.py#L244). But feel free to change this in a separate PR.

sveneberth: Okay, i didn't now that we use the RateLimit already inside the core. So far this PR LGTM.

sveneberth (approved):

core/modules/user.py
sveneberth: Done with #463

PR #457, opened 2022-06-03
Add support for `Literal` type annotations in processTypeHint

sveneberth:
Parse also `Literal` type annotations in exposed methods.
Literals were introduced with [PEP-586](https://peps.python.org/pep-0586/) for Python >= 3.8.

Example:
```py
from typing import Literal
from viur.core import exposed

@exposed
def my_method(self, value: Literal['a', 'b', 7]):
    ...
```
Only `'a'`, `'b'` and `7` are allowed for `value`. Other values will be rejected.
The `7` will be accepted as `str`, but forwarded as `int`.

### DIFF
--- a/core/request.py
+++ b/core/request.py
@@ -417,6 +417,14 @@ class BrowseHandler():  # webapp.RequestHandler
 				return "True", True
 			else:
 				return "False", False
+		elif typeOrigin is typing.Literal:
+			inValueStr = str(inValue)
+			for literal in typeHint.__args__:
+				if inValueStr == str(literal):
+					return inValue, literal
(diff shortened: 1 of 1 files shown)


tsteinruecken (approved):

phorward (approved):

phorward (changes requested):
@sveneberth can you please merge current develop into this branch? Features should generally  be made against the develop branch, bug/hotfixes against main. Thanks!

phorward (approved):
OK nevermind. As #458 relies on this PR, and #458 cannot be merged into develop without huger changes, I'll do it this way now and we merge into main. Next time please do it only on develop.

PR #458, opened 2022-06-03
Improve docstrings and type hints

sveneberth:
This PR improve a bunch of docstrings and type hints. So hopefully our documentation with sphinx will benefit. But also IDEs oder linters.

A summery of the changes:
- Replace _server_ with _viur.core_
- Move (hopefully) all types from docstrings (`:type param: xyz`) into the method signatures as type annotation. So that we have only one kind and no conflicts.
- Add a few missing docstrings
- Add missing parameter descriptions, fix broken ones (due to renamed or removed parameters)
- Correct wrong types
- Improve formatting of some signatures (especially because they are now longer with the types)
- Added the imports which are necessary for the type hints and optimized (sorted) them.
- Remove `# -*- coding: utf-8 -*-` (Yes, it has nothing to do with the other changes, but since this PR changes so many files, we can remove that here at once.)

I didn't initially expect there to be so many changes. But a PR for each module makes IMO no sense...

This PR works only with #457 due to the added type hint `SkelType = Literal["node", "leaf"]` in the tree prototype.


tsteinruecken (approved):

phorward (approved):
Ok this is really a lot of change. I hope you tested it well and it incorporates with the current changes in develop. Next time, please do such huge changes against branch develop, as main shall be used for hotfixes only.

phorward (approved):
self-committed two suggestions.

PR #463, opened 2022-06-10
Implement `errors.TooManyRequests` exception

sveneberth:
Implement an exception class for HTTP Status "429 Too Many Requests"
and use it instead of Forbidden for rate limit blocked actions.

### DIFF
--- a/core/errors.py
+++ b/core/errors.py
@@ -188,7 +188,7 @@ class TooManyRequests(HTTPException):
 		super(TooManyRequests, self).__init__(status=429, name="Too Many Requests", descr=descr)
 
 	@classmethod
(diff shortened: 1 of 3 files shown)


tsteinruecken (changes requested):
Does this actually work?
I think this should fail due to circular imports (eg. errors -> ratelimit -> tasks -> errors).
My suggestion would be something like an "assertQuotaAvailable" function on the ratelimit object that raises that exception if necessary.

sveneberth:
> Does this actually work?
Yes, it did.

> My suggestion would be something like an "assertQuotaAvailable" function on the ratelimit object that raises that exception if necessary.
In hindsight, it's indeed logically better placed inside the `RateLimit`-module and easier to use as an assertion. I added a commit.

PR #469, opened 2022-06-22
fix: Handle list/tuple equally on dynamic values

phorward:


tsteinruecken (changes requested):
Okay for me. However, my suggestion would be to introduce a translation format (eg. kindname-bonename-%i) for theses format-strings so translations could be more specific (especially on numeric values in that list)

phorward:
> Okay for me. However, my suggestion would be to introduce a translation format (eg. kindname-bonename-%i) for theses format-strings so translations could be more specific (especially on numeric values in that list)

@tsteinruecken this PR is an improvement for #390. As I've stated there already in [this comment](https://github.com/viur-framework/viur-core/pull/390#issuecomment-1076859604), please make a separate proposal of your idea in form of a pull request.

XeoN-GHMB (approved):

PR #477, opened 2022-07-08
Removing dbtransfer and its usage

phorward:
This was discussed in VIUR Round 2022-07-08 and approved.


sveneberth (changes requested):
Please remove `conf["viur.exportPassword"]` and `conf["viur.importPassword"]` as well

sveneberth (changes requested):
And in the docs? ^^
https://github.com/viur-framework/viur-core/blob/159efda68c92e27af06683cbe18abdf371bd0880/docs/doc_tutorial/advanced/05_training_config.rst

PR #485, opened 2022-08-19
* version bump for viur-datastore to 1.3.1

skoegl:
Just version bumped viur-datastore...


phorward (approved):
Thanks for updating, looks good now after some fixes.
I've changed the base to main, as we can then release v3.2.2

PR #490, opened 2022-08-19
fixed getReferencedBlobs of TextBone

Grashalmbeisser:
so it allows multiple=True data to be saved


sveneberth (changes requested):
LGTM. The case multiple + languages (`if self.multiple and self.languages:`...) is also missing, can you add it?

XeoN-GHMB (approved):
As a quick fix okay. Please create a new issue for the case multiple + languages.

phorward (approved):
Quick fix, improved implementation will follow later on.

sveneberth (approved):

PR #492, opened 2022-09-03
Removing unused functions from renderers

phorward:
All these functions are called nowhere from viur-core, and are mostly a relict of old viur2 / hierarchy-prototype stuff.


XeoN-GHMB (approved):
Never used one of these. Okay for me

PR #494, opened 2022-09-05
Refactoring default renders for HTML/JSON/XML

phorward:
- Redundant code refactored into single functions
- Added a consequent action-parameter to all functions
  - in HTML-renderer, the action describes the concrete action that
    is triggered, but the default name of the template is used from
	the specific action.
  - This means, e.g. a clone-operation uses edit and editSuccess, but will pass action=clone, so the template renders editTemplate and the template can decide whether it provides a different form for a clone-operation or not.
- Removed tree-specific functions as they are now solved as edit-actions
- This is a preliminary for further development on the tree module

This commit is fully backward-compatible to existing code.
Further improvements, modularization and removal of bad code will follow later.

### DIFF
--- a/core/render/html/default.py
+++ b/core/render/html/default.py
@@ -3,8 +3,8 @@ from collections import OrderedDict, namedtuple
 import codecs
 import logging
 import os
-import warnings
-from jinja2 import ChoiceLoader, Environment, FileSystemLoader
+
+from jinja2 import ChoiceLoader, Environment, FileSystemLoader, Template
 from typing import Any, Dict, List, Literal, Optional, Tuple, Union
 
 from viur.core import db, errors, securitykey, utils
@@ -50,9 +50,7 @@ class Render(object):
     editSuccessTemplate = "edit_success"
     deleteSuccessTemplate = "delete_success"
 
-    listRepositoriesTemplate = "list_repositories"
-    reparentSuccessTemplate = "reparent_success"
-    cloneSuccessTemplate = "clone_success"
+    listRepositoriesTemplate = "list_repositories"  # fixme: This is a relict, should be solved differently (later!).
 
     __haveEnvImported_ = False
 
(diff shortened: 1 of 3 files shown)


core/render/html/default.py
```diff
@@ -310,252 +308,215 @@ def renderBoneValue(self,
...
-        return template.render(repoObj=obj, params=params, **kwargs)
+        For details, see self.render_action_view_template().
+        """
+        return self.render_action_view_template("deleteSuccess", skel, action, tpl, params, **kwargs)
 
-    def cloneSuccess(self, tpl: str = None, params: Any = None, *args, **kwargs) -> str:
+    def listRootNodes(  # fixme: This is a relict, should be solved differently (later!).
+        self,
+        repos: List[Dict[Literal["key", "name"], Any]],
+        action: str = "listrootnodes",
```
sveneberth, line 502: Can you add the parameter description of `action` in the docstring?

core/render/html/default.py
```diff
@@ -310,252 +308,215 @@ def renderBoneValue(self,
...
 
-        template = self.getEnv().get_template(self.getTemplateFileName(tpl))
-        return template.render(params=params, **kwargs)
+        for skel in skellist:
+            skel.renderPreparation = self.renderBoneValue
+
+        return template.render(skellist=skellist, action=action, params=params, **kwargs)
 
-    def list(self, skellist: SkelList, tpl: str = None, params: Any = None, **kwargs) -> str:
+    def view(self, skel: SkeletonInstance, action: str = "view", tpl: str = None, params: Any = None, **kwargs) -> str:
```
sveneberth, line 424: Can you add the parameter description of `action` in the docstring?

core/render/html/default.py
```diff
@@ -310,252 +308,215 @@ def renderBoneValue(self,
 
         return None
 
-    def add(self, skel: SkeletonInstance, tpl: str = None, params: Any = None, *args, **kwargs) -> str:
+    def get_action_template(self, action: str, tpl: str) -> Template:
```
sveneberth, line 311: I think `template` would be a more expressive parameter-name than `tpl`.
In this method this shouldn't be a problem. In the others (`view`, etc.) we would have a backward compatibility problem – But would be nice as well.

sveneberth (changes requested):

sveneberth (approved):

PR #499, opened 2022-09-07
Set new max Value for SortIndexBone

ArneGudermann:
This pull request set the max value for a SortIndexBone higher, because the UNIX timestamp is lager than 2^30.

### DIFF
--- a/core/bones/sortindex.py
+++ b/core/bones/sortindex.py
@@ -1,5 +1,5 @@
 from viur.core.bones.numeric import NumericBone
-import time, typing
+import time, typing,sys
 
 
 class SortIndexBone(NumericBone):
@@ -10,7 +10,7 @@ class SortIndexBone(NumericBone):
         *,
         defaultValue: typing.Union[int, float] = lambda *args, **kwargs: time.time(),
         descr: str = "SortIndex",
-        max: typing.Union[int, float] = pow(2, 30),
+        max: typing.Union[int, float] = sys.maxsize,
         precision: int = 8,
         **kwargs
     ):
--- a/core/bones/sortindex.py
+++ b/core/bones/sortindex.py
@@ -1,5 +1,5 @@
 from viur.core.bones.numeric import NumericBone
-import time, typing,sys
+import time, typing
 
 
 class SortIndexBone(NumericBone):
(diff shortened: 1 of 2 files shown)


sveneberth (commented):
I think this could be interesting for the [`NumericBone`](https://github.com/viur-framework/viur-core/blob/222de3385885b80c23c954d60788860b40194c33/core/bones/numeric.py#L20-L21) as well.
I suggest to change the `min` and `max` there and remove the `max` parameter from the `SortIndexBone` because it has the same value anyway.

core/bones/sortindex.py
```diff
@@ -1,5 +1,5 @@
 from viur.core.bones.numeric import NumericBone
-import time, typing
+import time, typing,sys
```
sveneberth, line 2: This is no longer necessary.

```suggestion
import time, typing
```

sveneberth (changes requested):

sveneberth (approved):

phorward (dismissed):
Hello @ArneGudermann and @sveneberth I'm also fine with this solution!

phorward:
Wait a sec... I still have one blur with this issue before merging.

What is the maximum value that can be stored to datastore currently? We need to affirm on this as well, because Python uses some kind of BigInt implementation and can story integer values of arbitrary size. Therefore, I'm sure that there is a limit for numbers.

We first need to find out on our own, or find a valid source about these limits.

phorward:
We also need to discuss where the limits are with other databases, which is a planned feature.

ArneGudermann:
The value for sys.maxsize is max 2^63 for a 64 bit system. (https://www.geeksforgeeks.org/sys-maxsize-in-python/)

The datastore can store up to 2^64 bit for int and float. (https://cloud.google.com/datastore/docs/concepts/entities#integer)

MongoDB can also store Long values (https://www.mongodb.com/docs/mongodb-shell/reference/data-types/)

I think all Databases can store numeric values up to 2^64 bit.

phorward:
@ArneGudermann thanks for clarification.

I've made an improvement to this PR in 24c1c5dbf133d6d0a87e6c08ad7984ae8f676828. Can you please check this out? It avoids setting min or max to something which beyond the system limits. There might also be more constraints, e.g. that min may not be greater than max, or a maximum for precision, but this would blow up the code in here. For me, it is important that no higher limits than the maximums can be set.

PR #501, opened 2022-09-08
Convert TranslationsKey

ArneGudermann:
This pull request solves the problem that the core crashes when you pass a key to the translate that is not of type string.

https://github.com/viur-framework/viur-core/blob/c89dfd68b595ffa9c2a0f0e09d0984bf18cd3bd8/core/modules/user.py#L64-L72

runs through

https://github.com/viur-framework/viur-core/blob/c89dfd68b595ffa9c2a0f0e09d0984bf18cd3bd8/core/render/json/default.py#L84-L87

or this:
`testSelect=SelectBone(descr="testSelect0",values={"a":1,"b":2,"c":3})`

### DIFF
--- a/core/i18n.py
+++ b/core/i18n.py
@@ -70,6 +70,9 @@ class translate:
             meanings in the target language.
         """
         super(translate, self).__init__()
+
+        if not isinstance(key,str):
+            key=str(key)
         self.key = key.lower()
         self.defaultText = defaultText or key
         self.hint = hint
--- a/core/i18n.py
+++ b/core/i18n.py
@@ -71,7 +71,7 @@ class translate:
         """
         super(translate, self).__init__()
 
-        if not isinstance(key,str):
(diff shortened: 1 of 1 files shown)


core/i18n.py
```diff
@@ -70,6 +70,9 @@ def __init__(self, key: str, defaultText: str = None, hint: str = None):
             meanings in the target language.
         """
         super(translate, self).__init__()
+
+        if not isinstance(key,str):
```
phorward, line 74: I think the isinstance-test before is slower that just converting all to str. A key, which is already a str, won't be converted as str() itself has an equivalent behavior.
```suggestion
        key = str(key)  # ensure key is a a str
```

phorward (changes requested):
Hello @ArneGudermann, generally the pull request is useful, but I think you just can convert key to str everytime.

phorward (approved):

PR #508, opened 2022-09-14
Fix UNB Task and add Logging

ArneGudermann:
This pull request fix the bug that an unbound task cant't call an unbound task again.
Like so :
https://github.com/viur-framework/viur-core/blob/378157a87c97004c56d810ff066e639237a60e7a/core/skeleton.py#L1277-L1278

https://github.com/viur-framework/viur-core/blob/378157a87c97004c56d810ff066e639237a60e7a/core/skeleton.py#L1325-L1326

The problem is we use here the request object form above not the current request object :
https://github.com/viur-framework/viur-core/blob/378157a87c97004c56d810ff066e639237a60e7a/core/tasks.py#L263
comes form :
https://github.com/viur-framework/viur-core/blob/378157a87c97004c56d810ff066e639237a60e7a/core/tasks.py#L204

But in this if statement we look for the current request object :
https://github.com/viur-framework/viur-core/blob/378157a87c97004c56d810ff066e639237a60e7a/core/tasks.py#L421-L426

Additionally this pull request adds better logging for the task.


core/tasks.py
```diff
@@ -220,6 +219,10 @@ def deferred(self, *args, **kwargs):
                                             retryCount))
         cmd, data = json.loads(req.body, object_hook=jsonDecodeObjectHook)
         funcPath, args, kwargs, env = data
+        if conf.get("viur.tasks.logs", False):
+            logging.info("Call Task %s with CMD = %s" % (funcPath,cmd))
+            logging.info("Call Task Arguments args = %s kwargs = %s" % (str(args), str(kwargs)))
+            logging.info("Call Task Env = %s" % (str(env)))
```
phorward, line 225: Can you please reduce to amount of logging.info calls to just one with all required information?
```suggestion
            logging.info(f"Call task {funcPath} with cmd={cmd!r} args={args!r} kwargs={kwargs!r} env={env!r}")
```

core/tasks.py
```diff
@@ -495,6 +496,10 @@ def mkDefered(func, self=__undefinedFlag_, *args, **kwargs):
             response = taskClient.create_task(parent=parent, task=task)
 
             print('Created task {}'.format(response.name))
```
phorward, line 498: Please remove this print!
```suggestion
```

core/tasks.py
```diff
@@ -495,6 +496,10 @@ def mkDefered(func, self=__undefinedFlag_, *args, **kwargs):
...
+            if conf.get("viur.tasks.logs",False):
+                logging.info("Create Task %s.%s" % (func.__name__, func.__module__))
+                logging.info("Create Task Arguments args = %s kwargs = %s"%(str(args),str(kwargs)))
```
phorward, line 501: Please reduce this logging as well, and complete to the value of `env`:
```suggestion
                logging.info(f"Create task {func.__name__}.{func.__module__} with args={args!r} kwargs={kwargs!r} env={env!r}")
```

phorward (changes requested):
Hello @ArneGudermann,
thanks for the explanation and bugfix. Can you please apply my 3 suggestions for the logging-output as well as one very legacy print()-call?

phorward (approved):

core/tasks.py
```diff
@@ -220,6 +219,9 @@ def deferred(self, *args, **kwargs):
                                             retryCount))
         cmd, data = json.loads(req.body, object_hook=jsonDecodeObjectHook)
         funcPath, args, kwargs, env = data
+        if conf.get("viur.tasks.logs", False):
```
sveneberth, line 222: Instead of wrapping a logging statement with an if, you should better use a logger instance which isn't the root-logger and is selectively controllable.


Create a new logger at the begin of the file (after the imports):
```py
logger = logging.getLogger(__name__)
```
and set the level semi-high:
```py
logger.setLevel(logging.WARNING)
```

If you need the debugging output, you can decrease the level in your project main.py 
```py
logging.getLogger("viur.core.tasks").setLevel(logging.WARNING)
```

---

IMO this should be `debug` and not `info` level. With the changes above and in a more short version it could be:
```py
logger.debug(f"Call task {funcPath} with {cmd=} {args=} {kwargs=} {env=}")
```

sveneberth (changes requested):

core/tasks.py
```diff
@@ -220,6 +222,8 @@ def deferred(self, *args, **kwargs):
                                             retryCount))
         cmd, data = json.loads(req.body, object_hook=jsonDecodeObjectHook)
         funcPath, args, kwargs, env = data
+        logger.debug(f"Call task {funcPath} with cmd={cmd!r} args={args!r} kwargs={kwargs!r} env={env!r}")
```
phorward, line 225: @ArneGudermann as @sveneberth suggested, please define this line as this, which is much shorter and a cool feature to use and spread.
```suggestion
        logger.debug(f"Call task {funcPath} with {cmd=} {args=} {kwargs=} {env=}")
```

core/tasks.py
```diff
@@ -494,7 +496,9 @@ def mkDefered(func, self=__undefinedFlag_, *args, **kwargs):
             # Use the client to build and send the task.
             response = taskClient.create_task(parent=parent, task=task)
 
-            print('Created task {}'.format(response.name))
+
+            logger.debug(f"Create task {func.__name__}.{func.__module__} with args={args!r} kwargs={kwargs!r} env={env!r}")
```
phorward, line 500:
```suggestion
            logger.debug(f"Create task {func.__name__}.{func.__module__} with {args=} {kwargs=} {env=}")
```

phorward (changes requested):
Can you please integrate my two suggestions?

core/tasks.py
phorward: @sveneberth this is a cool feature with the `{cmd=}` in f-strings, didn't know about it, yet!

phorward (approved):

PR #511, opened 2022-09-16
Make it possible that multiple languages could be required

### LINKED ISSUE #502: Multiple languages should be required/requireable in bones

sveneberth:
This PR should resolve #502. 

But, I'm not sure if this is the correct way. Therefore, I would like to hear other opinions about this.

Especially this condition confused me:
https://github.com/viur-framework/viur-core/blob/378157a87c97004c56d810ff066e639237a60e7a/core/skeleton.py#L335-L336
The skeleton will be only marked as not-completed if the field is not set and required and the skeleton has a key. Which means on _edit_ this field can be empty, but on _add_ it must be set. This makes no sense to me...


sveneberth:
> I would like to hear other opinions about this

@viur-framework/developers  nobody? :thinking:😞

XeoN-GHMB (approved):

phorward:
@sveneberth please merge thje current develop into your branch and resolve conflicts.

sveneberth:
> @sveneberth please merge thje current develop into your branch and resolve conflicts.

Done.

And I've added some validation to ensure `required` uses only values from `languages`.

phorward (approved):

XeoN-GHMB (approved):

PR #512, opened 2022-09-20
new derives renderer

ArneGudermann:
With this pr it is now possible to render Deriveds for images and PDF files in a cloud function.
The corresponding cloudfunction can be found here .
https://github.com/viur-framework/viur-cloudfunctions/tree/main/thumbnailer

You can use it like so:
main.py:
```python
from viur.core.modules.file import cloudfunction_thumbnailer

conf["viur.file.thumbnailerURL"]="https://xxxxx.cloudfunctions.net/imagerenderer"
conf["viur.file.derivers"] = {"thumbnail": cloudfunction_thumbnailer}

conf["derives_pdf"] = {
"thumbnail": [{"width": 1920,"sites":"1,2"}]
}
```
skeletons/xxx.py:
```python
test = FileBone(derive=conf["derives_pdf"])

```


core/modules/file.py
```diff
@@ -84,9 +86,10 @@ def unserialize(self, skel, name):
 
 
 def thumbnailer(fileSkel, existingFiles, params):
-    blob = bucket.get_blob("%s/source/%s" % (fileSkel["dlkey"], fileSkel["name"]))
+    orgifileName = fileSkel["name"].replace("&#040;", "(").replace("&#041;", ")").replace("&#061;", "=")
```
ArneGudermann, line 89: We must replace here to create derives of an image with one of these characters, because the filename is stored in a StringBone and this will be escaped

core/modules/file.py
```diff
@@ -128,6 +131,98 @@ def thumbnailer(fileSkel, existingFiles, params):
...
+            version="v4")
+
+    dataDict = {"url": signedUrl,
+                "name": fileSkel["name"],
+                "params": params,
+                "minetype": fileSkel["mimetype"],
+                "baseUrl": utils.currentRequest.get().request.host_url.lower(),
+                "targetKey": fileSkel["dlkey"],
+                "nameOnly": True
+                }
```
phorward, line 165: Please follow PEP8-guideline
```suggestion
    dataDict = {
        "url": signedUrl,
        "name": fileSkel["name"],
        "params": params,
        "minetype": fileSkel["mimetype"],
        "baseUrl": utils.currentRequest.get().request.host_url.lower(),
        "targetKey": fileSkel["dlkey"],
        "nameOnly": True
    }
```

core/modules/file.py
```diff
@@ -128,6 +131,98 @@ def thumbnailer(fileSkel, existingFiles, params):
...
+        expiresAt = datetime.now() + timedelta(seconds=60)
+        signing_credentials = compute_engine.IDTokenCredentials(authRequest, "")
+        contentDisposition = "filename=%s" % fileSkel["name"]
+        signedUrl = blob.generate_signed_url(
+            expiresAt,
+            credentials=signing_credentials,
+            response_disposition=contentDisposition,
+            version="v4")
+
+    dataDict = {"url": signedUrl,
```
phorward, line 201: same here

core/modules/file.py
ArneGudermann: Yes we can make that I copy the code from here:
https://github.com/viur-framework/viur-core/blob/312d24b96b880e01ee3902043795b1a39633a79c/core/utils.py#L149

xnopasaranx:
Can we please get the naming correct? This would probably be best with our 3.5 release, because it will introduce breaking changes anyway.

derives is a verb
the noun would be derivation
derived could be the past tense of the verb OR an adjective. I think it is important to use correct English in an open source project, at least when naming methods and classes in the code.

core/modules/file.py
xnopasaranx: there is a typo here:
`"minetype": fileSkel["mimetype"],
`
needs to be

`"mimetype": fileSkel["mimetype"],`

ArneGudermann:
Hello @xnopasaranx  can you please  mark the specific code that you mean ?

sveneberth (changes requested):

core/modules/file.py
```diff
@@ -84,9 +86,10 @@ def unserialize(self, skel, name):
 
 
 def thumbnailer(fileSkel, existingFiles, params):
-    blob = bucket.get_blob("%s/source/%s" % (fileSkel["dlkey"], fileSkel["name"]))
+    orgifileName = html.unescape(fileSkel["name"])
```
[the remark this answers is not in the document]
ArneGudermann: That was copied code for the other thumbnailer function :). I changed it in both functions.

core/modules/file.py
```diff
@@ -128,6 +131,84 @@ def thumbnailer(fileSkel, existingFiles, params):
...
+                version="v4")
+        return signedUrl
+
+    def make_request():
+        import requests as _requests
+        headers = {'Content-type': 'application/json'}
+        dataStr = base64.b64encode(json.dumps(dataDict).encode("UTF-8"))
+        sig = utils.hmacSign(dataStr)
+        datadump = json.dumps({"dataStr": dataStr.decode('ASCII'), "sign": sig})
+        return _requests.post(conf["viur.file.thumbnailerURL"], data=datadump, headers=headers)
```
sveneberth, line 163: I think this could be just: 

```suggestion
        dataStr = base64.b64encode(json.dumps(dataDict).encode("UTF-8"))
        sig = utils.hmacSign(dataStr)
        data = {"dataStr": dataStr.decode('ASCII'), "sign": sig}
        return _requests.post(conf["viur.file.thumbnailerURL"], json=data)
```

sveneberth (changes requested):
Hey @ArneGudermann your PR looks already better😀, thanks for the changes. But I have still some more points:

- [x] Please add a docstring for the documentation. 
- [ ] Also I would like to follow PEP-8 (snake_case, shorter line-length) but obliviously this makes technically no difference – @phorward  Should decide if this is necessary ;)
- [ ] I think a little bit of error handling would be necessary. In case of ...
  - [ ] Invalid API-Response (from cloudfunction)
  - [ ] Network problem, timeout, mis-configured (wrong url --> cloud function not reachable)

What do u think?

sveneberth (changes requested):
What about error handling? (See [my comment](https://github.com/viur-framework/viur-core/pull/512#pullrequestreview-1142154655))

phorward:
>     * [ ]  Also I would like to follow PEP-8 (snake_case, shorter line-length) but obliviously this makes technically no difference – @phorward  Should decide if this is necessary ;)

If possible, please use PEP-8 already in new code. I'm waiting for this PR to finish to merge it into develop.

sveneberth (changes requested):
error handling?

sveneberth (approved):
> @sveneberth [2219cc7](https://github.com/viur-framework/viur-core/commit/2219cc7cbee5b14ae8f7e8c451db8f9b8ea613c2) not enough ?

Oh sorry. GitHub showed me only the latest commit baf4d8ab8e4f3f54668cab7c86f62d08fa27cd47 as recent changes (since the last visit). I didn't saw the other one. My bad.

I think as first approach this looks okay. I'm still not satisfied with the lack of type annotations and PEP-8 syntax. But that can also be done later.

ArneGudermann:
Yes, I also updated the corresponding cloudfunction this morning an all works well.

phorward (approved):
Ok then let's give it a try. We can do the PEP-8-relevant renamings later on.

PR #513, opened 2022-09-21
Cleaning up and renaming CallDeferred

phorward:
- Cleans the @CallDeferred decorator's internals to behave equally with or without a task queue
- Added more debug to find bugs
- Renamed callDeferred to CallDeferred
- Added wrapper callDeferred with deprecation warning
- Further renamings of typo "defered" into "deferred" and PEP8-names for replaced items
- Reverted logger use back to logging, I will explain for this. We should make this later on.

### DIFF
--- a/core/tasks.py
+++ b/core/tasks.py
@@ -392,83 +392,94 @@ def noRetry(f):
     return wrappedFunc
 
 
-def callDeferred(func):
+def CallDeferred(func):
     """
         This is a decorator, which always calls the function deferred.
         Unlike Googles implementation, this one works (with bound functions)
     """
     if "viur_doc_build" in dir(sys):
         return func
+
     __undefinedFlag_ = object()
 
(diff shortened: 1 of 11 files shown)


ArneGudermann (approved):

phorward:
> I have more improvements for this one. I created a PR in your repo since this was easier than the github suggestions: [phorward#1](https://github.com/phorward/viur-core/pull/1)

Thanks, this looks interesting. But isn't there an `from functools import wraps` missing?

core/tasks.py
```diff
@@ -17,9 +17,6 @@
 from viur.core.config import conf
 from viur.core.utils import currentLanguage, currentRequest, currentSession
 
-logger = logging.getLogger("viur.core.tasks")
```
phorward, line 20: Let's fix this in a separate pull request for specific modules. I need a working logging for this PR.

PR #514, opened 2022-09-22
Implement isEmpty-function for StringBone

### LINKED ISSUE #496: `required=True` bone can be set to whitespace - is this valid?
This bone definition
```py
name = StringBone(
    descr="Name",
    required=True
)
```
allows to save a value with content " ". Is this behavior correct?

phorward:
This can still be overridden individually using the `isEmptyFunc` parameter that can be given to a bone. Fixes #496.

### DIFF
--- a/core/bones/string.py
+++ b/core/bones/string.py
@@ -37,13 +37,15 @@ class StringBone(BaseBone):
     def getEmptyValue(self):
         return ""
 
+    def isEmpty(self, value):
+        return not bool(value.strip())
+
     def singleValueFromClient(self, value, skel, name, origData):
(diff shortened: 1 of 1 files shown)


phorward:
Hello, regarding the failing tests: This occurs when something different then a str is given to isEmpty.
A general question on this: Why can't isEmpty/isValid etc. just rely that the provided value is of the given type of the bone, which is parsed using fromClient? Either the tests must be changed, or the generic concept of VIUR. Well okay, I think this is a generall discussion topic, just fixing it now this way and see the rest later on.

XeoN-GHMB (approved):

PR #516, opened 2022-09-23
Create a new hmac Key if no one is set

ArneGudermann:
Hello this pull request create a new hmac key if there is no one set and write to in the database.

Maybe we should delete this in the viur-base .
https://github.com/viur-framework/viur-base/blob/9cede043ceb286dca6e898115b83212623a0afec/deploy/main.py#L52

### DIFF
--- a/core/__init__.py
+++ b/core/__init__.py
@@ -254,7 +254,7 @@ def setup(modules: Union[object, ModuleType], render: Union[ModuleType, Dict] = 
     initializeTranslations()
     if conf["viur.file.hmacKey"] is None:
         from viur.core import db
-        key = db.Key("viur-conf", "viur-conf-modulekey")
+        key = db.Key("viur-conf", "viur-conf")
         obj = db.Get(key)
         if not obj:  # we not have a conf yet
             logging.warning("Create new hmac Key")
--- a/core/__init__.py
+++ b/core/__init__.py
@@ -252,7 +252,21 @@ def setup(modules: Union[object, ModuleType], render: Union[ModuleType, Dict] = 
             assert uri is not None and (uri.lower().startswith("https://") or uri.lower().startswith("http://"))
     runStartupTasks()  # Add a deferred call to run all queued startup tasks
     initializeTranslations()
-    assert conf["viur.file.hmacKey"], "You must set a secret and unique Application-Key to viur.file.hmacKey"
+    if conf["viur.file.hmacKey"] is None:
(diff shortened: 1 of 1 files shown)


core/__init__.py
```diff
@@ -252,7 +252,21 @@ def setup(modules: Union[object, ModuleType], render: Union[ModuleType, Dict] =
             assert uri is not None and (uri.lower().startswith("https://") or uri.lower().startswith("http://"))
     runStartupTasks()  # Add a deferred call to run all queued startup tasks
     initializeTranslations()
-    assert conf["viur.file.hmacKey"], "You must set a secret and unique Application-Key to viur.file.hmacKey"
+    if conf["viur.file.hmacKey"] is None:
+        from viur.core import db
+        key = db.Key("viur-conf", "viur-conf-modulekey")
```
phorward, line 257: Can you please rename this to just kind "viur-conf", key "viur-conf"`
```suggestion
        key = db.Key("viur-conf", "viur-conf")
```

core/__init__.py
```diff
@@ -252,7 +252,21 @@ def setup(modules: Union[object, ModuleType], render: Union[ModuleType, Dict] =
...
+            logging.warning("Create new hmac Key")
+            hmacKey = utils.generateRandomString(length=20)
+            obj = db.Entity(key)
+            obj["hmacKey"] = hmacKey
+            db.Put(obj)
+            conf["viur.file.hmacKey"] = bytes(hmacKey, "utf-8")
+        else:
+            if obj["hmacKey"]:
+                conf["viur.file.hmacKey"] = bytes(obj["hmacKey"], "utf-8")
+
```
phorward, line 269: I like the idea of having a `viur-conf`-kind for future problems of the same kind.
Therefore, an improved version of yours would be this:
```suggestion
        if not (obj := db.Get(key)):  # create a new "viur-conf"?
            logging.info("Creating new viur-conf")
            obj = db.Entity(key)

        if "hmacKey" not in obj:  # create a new hmacKey
            logging.info("Creating new hmacKey")
            obj["hmacKey"] = utils.generateRandomString(length=20)            
            db.Put(obj)

        conf["viur.file.hmacKey"] = bytes(obj["hmacKey"], "utf-8")
```

phorward (changes requested):
Hello @ArneGudermann this is a nice improvement, could you please check my suggestions?

phorward (approved):

PR #526, opened 2022-10-06
Improve check for viur.maxPostParamsCount

phorward:
Present a more useful error message to the user than just "Not Acceptable" when supplying too many arguments.

### DIFF
--- a/core/request.py
+++ b/core/request.py
@@ -448,7 +448,7 @@ class BrowseHandler():  # webapp.RequestHandler
         """
         # Prevent Hash-collision attacks
         kwargs = {}
-        stopCount = conf["viur.maxPostParamsCount"]
+        count = total = conf["viur.maxPostParamsCount"]
         try:
             for key, value in self.request.params.iteritems():
                 key = unicodedata.normalize("NFC", key)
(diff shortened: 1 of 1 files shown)


core/request.py
```diff
@@ -462,9 +462,12 @@ def findAndCall(self, path: str, *args, **kwargs) -> None:
...
                 else:
                     kwargs[key] = value
-                stopCount -= 1
-                if not stopCount:  # We reached zero; maximum PostParamsCount exceeded
-                    raise errors.NotAcceptable()
+
+                count -= 1
+                if count == 0:  # We reached zero; maximum PostParamsCount exceeded
+                    raise errors.NotAcceptable(
+                        f"""Too many arguments supplied, exceeding maximum of {total} allowed arguments per request"""
```
ArneGudermann, line 469: Can we here use conf["viur.maxPostParamsCount"] instead of "total" ?

phorward: We use it, but by a reference ;-) I did it this way to shorten the line to meet PEP8. Also, we don't do a __getattr__-call twice this way.

ArneGudermann (approved):

sveneberth:
In agreement with @phorward I have made a few adjustments in https://github.com/viur-framework/viur-core/pull/526/commits/1a667b1292cfbca9f2a2db86ad835dcf5e227050.

sveneberth (approved):

PR #532, opened 2022-10-21
Add support for Jinja tests

sveneberth:
In Jinja you can "test" values.
```html
1 is number? {{ 1 is number }}
1 is string? {{ 1 is string }}
```

This PR adds the possibilty to register custom tests and provide the `dict` and `list` test by default, which where already solved in the [viur-base as function](https://github.com/viur-framework/viur-base/blob/7e1b4a7a57ec52cf06bc6f332a1acb3d21a5149a/deploy/render/__init__.py#L3-L9).

So now you can do:
```html
{{ {1: 2, 3: 4} is dict }}
```

and write custom tests:
```py
@html.jinjaGlobalTest
def test_admin(render, value):
	return value is not None and "admin" in value["access"]
```
and use it:
```html
current user is admin: {{ getCurrentUser() is admin }}
```

For reference see https://jinja.palletsprojects.com/en/3.0.x/templates/#tests

### DIFF
--- a/core/render/html/utils.py
+++ b/core/render/html/utils.py
@@ -1,5 +1,6 @@
 __jinjaGlobals_ = {}
 __jinjaFilters_ = {}
+__jinjaTests_ = {}
 __jinjaExtensions_ = []
 
 
@@ -9,6 +10,10 @@ def getGlobalFunctions():
 
 def getGlobalFilters():
     return __jinjaFilters_
+
+
+def getGlobalTests():
+    return __jinjaTests_
 
 
 def getGlobalExtensions():
@@ -28,6 +33,16 @@ def jinjaGlobalFilter(f):
     Decorator, marks a function as a Jinja2 filter.
     """
     __jinjaFilters_[f.__name__] = f
+    return f
+
+
+def jinjaGlobalTest(f):
+    """
+    Decorator, marks a function as a Jinja2 test.
+
+    The method name can start with "test_" to avoid name conflicts.
+    """
+    __jinjaTests_[f.__name__.lstrip("test_")] = f
     return f
 
 
--- a/core/render/html/__init__.py
+++ b/core/render/html/__init__.py
@@ -1,2 +1,3 @@
 from .default import Render as default
 from .user import Render as user
+from .utils import jinjaGlobalExtension, jinjaGlobalFilter, jinjaGlobalFunction, jinjaGlobalTest
--- a/core/render/html/default.py
+++ b/core/render/html/default.py
@@ -1,11 +1,12 @@
 from collections import OrderedDict, namedtuple
 
 import codecs
+import functools
 import logging
 import os
 
 from jinja2 import ChoiceLoader, Environment, FileSystemLoader, Template
-from typing import Any, Dict, List, Literal, Optional, Tuple, Union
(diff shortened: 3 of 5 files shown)


core/render/html/utils.py
```diff
@@ -31,6 +36,16 @@ def jinjaGlobalFilter(f):
     return f
 
 
+def jinjaGlobalTest(f):
+    """
+    Decorator, marks a function as a Jinja2 test.
+
+    The method name can start with "test_" to avoid name conflicts.
+    """
+    __jinjaTests_[f.__name__.lstrip("test_")] = f
```
phorward, line 45: I'm a little unhappy with this. What do you think of providing a decorator `jinjaGlobalTest` that supports an alternative operand for the name, like so:
```python
@jinjaGlobalTest  # registers admin as admin
def admin(render, value):
	return value is not None and "admin" in value["access"]

@jinjaGlobalTest("dict")  # registers test_dict as dict
def test_dict(render: Render, value: Any) -> bool:
    """Jinja2 test: Return True if the object is a dict."""
    return isinstance(value, dict)
```

An alternative suggestions would be that tests must start with `test_`, which should raise an assertion error when not the case.

What do you think?

phorward (changes requested):
Hello  @sveneberth, this is a nice feature, thanks for providing the pull request, I appreciate it much.

I'm only a little bit uncertain about naming conventions. Defining tests starting with `test_` are nice, but IMHO then functions and filters must follow a similar convention. Maybe `jinja_function_*`, `jinja_filter_*`, `jinja_test_*`?

sveneberth:
> Hello @sveneberth, this is a nice feature, thanks for providing the pull request, I welcome this feature.
> 
> I'm only a little bit uncertain about naming conventions. Defining tests starting with `test_` are nice, but IMHO then functions and filters must follow a similar convention. Maybe `jinja_function_*`, `jinja_filter_*`, `jinja_test_*`?

In Jinja they use the following naming conversions for non-buildins, used [here](https://github.com/pallets/jinja/blob/5d9ece6d65ae4c2f57dd2bdccc28943984a767f9/src/jinja2/environment.py#L348-L351):
- `do_`-prefix for filters ([jinja2/filters.py](https://github.com/pallets/jinja/blob/5d9ece6d65ae4c2f57dd2bdccc28943984a767f9/src/jinja2/filters.py#L1785-L1840))
- `test_`-prefix for tests [jinja2/tests.py])(https://github.com/pallets/jinja/blob/5d9ece6d65ae4c2f57dd2bdccc28943984a767f9/src/jinja2/tests.py#L215-L255)
- and globals have no prefix ([jinja2/defaults.py](https://github.com/pallets/jinja/blob/5d9ece6d65ae4c2f57dd2bdccc28943984a767f9/src/jinja2/defaults.py#L29-L36)) (but they are not only designed for functions (therefore this comparison lags) — I have already added `OrderedDict` to the globals in a project to construct `OrderdDict` in Jinja for example).

core/render/html/utils.py
sveneberth: I like this idea :+1:. Please check out my latest commit.

phorward (approved):
This is okay for now

PR #533, opened 2022-10-22
Accept variations parameters with bool type hints

phorward:
VIUR currently allows only to pass True, yes or 1 as valid strings when a bool type hint is accepted. This is confusing, especially when true is provided, which is the valid true keyword for boolean true in the javascript word (and also many other, more professional programming languages than Python...)

### DIFF
--- a/core/request.py
+++ b/core/request.py
@@ -428,7 +428,7 @@ class BrowseHandler():  # webapp.RequestHandler
             if not isinstance(inValue, str):
                 raise TypeError(f"Input argument to boolean typehint is not a str, but f{type(inValue)}")
 
-            if inValue.strip().lower() in ("true", "yes", "1"):
+            if inValue.strip().lower() in conf["viur.bone.boolean.str2true"]:
                 return "True", True
 
             return "False", False
--- a/core/request.py
+++ b/core/request.py
@@ -426,11 +426,13 @@ class BrowseHandler():  # webapp.RequestHandler
             return str(f), f
         elif typeHint is bool:
             if not isinstance(inValue, str):
-                raise TypeError("Input argument to boolean typehint is not a string (probably a list)")
-            if inValue in [str(True), u"1", u"yes"]:
+                raise TypeError(f"Input argument to boolean typehint is not a str, but f{type(inValue)}")
+
+            if inValue.lower().strip() in ("true", "yes", "jawoll, herr oberleutnant", "1"):
                 return "True", True
-            else:
-                return "False", False
+
+            return "False", False
+
         elif typeOrigin is typing.Literal:
             inValueStr = str(inValue)
             for literal in typeHint.__args__:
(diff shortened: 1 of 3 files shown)


core/request.py
```diff
@@ -426,11 +426,13 @@ def processTypeHint(self, typeHint: typing.ClassVar, inValue: typing.Union[str,
             return str(f), f
         elif typeHint is bool:
             if not isinstance(inValue, str):
-                raise TypeError("Input argument to boolean typehint is not a string (probably a list)")
-            if inValue in [str(True), u"1", u"yes"]:
+                raise TypeError(f"Input argument to boolean typehint is not a str, but f{type(inValue)}")
+
+            if inValue.lower().strip() in ("true", "yes", "jawoll, herr oberleutnant", "1"):
```
sveneberth, line 431:
```suggestion
            if inValue.lower().strip() in ("true", "yes", "jawoll, herr oberleutnant", "1", not "not", "yup", "yeah", "sí"):
```

sveneberth (changes requested):

core/request.py
phorward: Ok, this is my final proposal on this:
```suggestion
            if inValue.lower().strip() in ("true", "yes", "1"):
```

core/request.py
```diff
@@ -426,11 +426,13 @@ def processTypeHint(self, typeHint: typing.ClassVar, inValue: typing.Union[str,
             return str(f), f
         elif typeHint is bool:
             if not isinstance(inValue, str):
-                raise TypeError("Input argument to boolean typehint is not a string (probably a list)")
-            if inValue in [str(True), u"1", u"yes"]:
+                raise TypeError(f"Input argument to boolean typehint is not a str, but f{type(inValue)}")
+
+            if inValue.strip().lower() in ("true", "yes", "1"):
```
sveneberth, line 431: Maybe it would be nice to support the _yes_ of the current language.

Not tested yet, but i imagine something like this:
```suggestion
            if inValue.strip().lower() in ("true", "yes", "1", translate("yes")):
```

sveneberth (changes requested):

sveneberth:
In this context we should do the same behavior in the `BooleanBone`.
https://github.com/viur-framework/viur-core/blob/198478b97d1fc9513b4bea4854164bd7e2ebb9ff/core/bones/boolean.py#L46

core/request.py
phorward: yes this is a fine idea, it needs a little bit of testing and more investigation.

sveneberth: Okay, let's do this later. I've created the issue https://github.com/viur-framework/viur-core/issues/537

sveneberth (changes requested):
Please add the same behavior in the `BooleanBone`. Otherwise the core would be inconsistent.

sveneberth (approved):

PR #539, opened 2022-11-02
Extend linter config

Grashalmbeisser:
I added an tox.ini for the viur core project in which u can setup ur pep8 preferences
also a pep8 checker command in the Pipfile "pep8check" which allows u to check only ur diffed files.

It runs with `pipenv run pep8check`

~To add the PR check APP use the following link: https://github.com/OrkoHunter/pep8speaks~


tox.ini
```diff
@@ -0,0 +1,6 @@
+[pycodestyle]
+show-pep8 = True
```
sveneberth, line 2: Please use 
```ini
show-pep8 = False
```

sveneberth (changes requested):

tox.ini
Grashalmbeisser: the change was implemented

sveneberth (approved):

sveneberth (approved):
Sorry for my latest commits...

sveneberth (changes requested):
This PR no longer makes sense as it is, since pep8speaks has stopped working. We have to find another runner or reduce this PR to the pycodestyle config.

See https://github.com/OrkoHunter/pep8speaks/issues/189#issuecomment-1267106628

sveneberth (approved):

phorward (approved):

PR #540, opened 2022-11-02
Add workflow of GitHub-Action for linting

Grashalmbeisser:
@sveneberth and i added a Github action, which test for propper pep8 usage

### DIFF
--- a/.github/workflows/pep8.yaml
+++ b/.github/workflows/pep8.yaml
@@ -31,7 +31,6 @@ jobs:
         run: |
           cd viur
           python -m pip install -U pip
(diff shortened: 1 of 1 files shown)


.github/workflows/pep8.yaml
```diff
@@ -0,0 +1,39 @@
+name: PEP8
```
sveneberth, line 1: I think that name would be a little more precise.
```suggestion
name: PEP8 linter
```

sveneberth (changes requested):

sveneberth (approved):

phorward (approved):

Issue #541, opened 2022-11-08
viur-session user keys are mistakenly handled as repr() values


phorward:
It seems that the "user" column of viur-session is handled inappropriately: The str()-representation (which falls back to repr()) of the db.Key object is written there, and also None (where, by the code, "guest" was intended). In ViUR2, the str()-representation of db.Keys always was a string now known as legacy-urlsafe.

![shot-2022-11-08_10-11-21](https://user-images.githubusercontent.com/16870072/200572022-55c92418-effa-4b68-a8f7-4463035632e3.jpg)

The pull request https://github.com/viur-framework/viur-datastore/pull/17 would elegantly resolve this mistake. 

Nonetheless, some parts of the code in session.py like this `dbSession["user"] = str(userid) or "guest"` either producing the repr-version of the key or the string "None" otherwise should be fixed.

phorward:
...and this is the reason why I created https://github.com/viur-framework/viur-core/issues/13 two years ago...

phorward:
Fixed by #544

PR #544, opened 2022-11-08
Session clean-up

phorward:
This is the first attempt to clean-up the session module.

- Renamed GaeSession into just Session
- Renamed "viurCookie_*" into "viur_cookie_*"
- Providing an `__init__`-function to initialize member variables consequently
- Fixed broken "guest" flagging with a new "guest"-property on the session entity
- "language" take-over resolved by conf persistentFieldsOnLogin/persistentFieldsOnLogout respectively (I think this is the more straightforward way)
- Renamed `guest` filtering identifier into Pythonic `__guest__`  to identify it as a special value.

Targets issues reported in #541


core/session.py
```diff
@@ -49,41 +46,43 @@ class GaeSession:
...
+
+            If the client supplied a valid Cookie, the session is read from the datastore, otherwise a new,
+            empty session will be initialized.
+        """
         if self.cookieName in req.request.cookies:
             cookie = str(req.request.cookies[self.cookieName])
-            data = db.Get(db.Key(self.kindName, cookie))
-            if data:  # Loaded successfully from Memcache
-                if data["lastseen"] < time() - conf["viur.session.lifeTime"]:
+            if data := db.Get(db.Key(self.kindName, cookie)):  # Loaded successfully from Memcache
```
sveneberth, line 72: There's no longer a memcache behind 
```suggestion
            if data := db.Get(db.Key(self.kindName, cookie)):  # Loaded successfully
```

core/session.py
```diff
@@ -99,34 +98,37 @@ def save(self, req: BrowseHandler):
...
+                    dbSession["lastseen"] = time.time()
+                    dbSession["user"] = str(userid) if userid else None  # to allow filtering for specific users
+                    dbSession["guest"] = userid is None  # to allow filtering guest sessions
                     dbSession.exclude_from_indexes = ["data"]
                     db.Put(dbSession)
                 except Exception as e:
                     logging.exception(e)
-                    raise  # FIXME
-                    pass
+                    raise  # FIXME???
```
sveneberth, line 121: Please remove this useless `try`/`except`.

core/session.py
```diff
@@ -99,34 +98,37 @@ def save(self, req: BrowseHandler):
...
                 sameSite = "; SameSite=%s" % self.sameSite if self.sameSite else ""
                 secure = "; Secure" if not conf["viur.instance.is_dev_server"] else ""
                 maxAge = "; Max-Age=%s" % conf["viur.session.lifeTime"] if not self.sessionCookie else ""
                 req.response.headerlist.append(("Set-Cookie", "%s=%s; Path=/; HttpOnly%s%s%s" % (
                     self.cookieName, self.cookieKey, sameSite, secure, maxAge)))
+
         except Exception as e:
-            raise  # FIXME
             logging.exception(e)
+            raise  # FIXME???
```
sveneberth, line 131: Please remove this useless `try`/`except` too.

core/session.py
```diff
@@ -177,32 +175,29 @@ def __setitem__(self, key: str, item: Any):
     def markChanged(self) -> None:
         """
             Explicitly mark the current session as changed.
-            This will force save() to write into the memcache /
-            datastore, even if it belives that this session had
-            not changed.
+            This will force save() to write into the memcache / datastore,
```
sveneberth, line 178:
```suggestion
            This will force save() to write into the datastore,
```

core/session.py
```diff
@@ -260,24 +255,24 @@ def killSessionByUser(user: Optional[str] = None):
         :param user: UserID, "guest" or None.
     """
     logging.error("Invalidating all sessions for %s" % user)
```
sveneberth, line 257:
```suggestion
    logging.info("Invalidating all sessions for %s" % user)
```
or
```suggestion
    logging.info(f"Invalidating all sessions for {user}")
```

core/session.py
```diff
@@ -217,7 +212,7 @@ def validateSecurityKey(self, key: str) -> bool:
         """
         Checks if key matches the current CSRF-Token of our session. On success, a new key is generated.
         """
-        if compare_digest(self.securityKey, key):
+        if hmac.compare_digest(self.securityKey, key):
```
sveneberth, line 215: low-hanging-fruit, and more for curiosity: Why should everything a module-import? Only one method from `hmac` is used, so the `from hmac import` is not really longer than writing two times `hmac.` before `compare_digest`. Also, a builtin is not vulnerable for circular import errors.

Please specify such things so that everyone can follow them and the code is consistent.

core/session.py
```diff
@@ -99,34 +98,37 @@ def save(self, req: BrowseHandler):
...
                     dbSession = db.Entity(db.Key(self.kindName, self.cookieKey))
                     dbSession["data"] = db.fixUnindexableProperties(self.session)
                     dbSession["staticSecurityKey"] = self.staticSecurityKey
                     dbSession["securityKey"] = self.securityKey
-                    dbSession["lastseen"] = time()
-                    # Store the userid inside the sessionobj, so we can kill specific sessions if needed
-                    dbSession["user"] = str(userid) or "guest"
+                    dbSession["lastseen"] = time.time()
+                    dbSession["user"] = str(userid) if userid else None  # to allow filtering for specific users
+                    dbSession["guest"] = userid is None  # to allow filtering guest sessions
```
sveneberth, line 116: Hmmm. Now we have to fields, while `guest` is always derivable from `user`. Futhermore `killSessionByUser("guest")` doesn't longer work. I don't see any advantage in this second field.

core/session.py
```diff
@@ -177,32 +175,29 @@ def __setitem__(self, key: str, item: Any):
...
 
     def reset(self) -> None:
         """
-            Invalids the current session and starts a new one.
+            Invalidates the current session and starts a new one.
 
             This function is especially useful at login, where
             we might need to create an SSL-capable session.
 
             :warning: Everything (except the current language) is flushed.
```
sveneberth, line 190:
```suggestion
            :warning: Everything is flushed.
```
would be now correct for this method.

But I'm unsure about the removal. I rather think we should keep it here. Because the persistent fields are only considered on login/out in the user module. If a session is identified as invalid (because it's too old) and we reset it, we should keep the old language.

sveneberth (changes requested):

core/session.py
phorward: Maybe we should add another conf-variable for this?

core/session.py
phorward: It's more convenient for readability to immediately know that the hmac stdlib function `compare_digest` is used, especially when you don't have the entire context, like in a pull request diff. In the past, I know that such functions where often written by hand. It's more clear to the programmer that standard library features are used here, and its good to use them elsewhere too.

core/session.py
phorward: Ok, we can turn it back, but then I would like to use `__guest__` as a special meaning identifier.

phorward: Nevertheless, the previous implementation either wrote the str "None" into the database or the user key's repr (with viur-datastore<1.3.3) or now the user key's url-safe representation.

core/session.py
sveneberth: ?

I see no reason, if we are always raising it.

core/session.py
sveneberth: Hmm. I don't know if we should always introduce config variables for everything...
But instead of a list of fields, maybe one should be able to inherit from the session class, adjust the reset method and then register the new class — if you need it.

Maybe we should test this on multilingual projects. For me we can leave it like this for now.

core/session.py
sveneberth: Okay. I think it would be good if this could then be used as a constant. 

```py
GUEST_USER = "__guest__"
```

core/session.py
phorward: > Maybe we should test this on multilingual projects. For me we can leave it like this for now.

Ok, then this is gonna be resolved.

phorward:
@sveneberth I've applied all of your suggestions and a little bit more. Can we merge this now?

sveneberth (approved):

PR #548, opened 2022-11-11
Remove `skey == ""` check from add/edit methods.

sveneberth:
If you make a POST request with valid form data but don't provide a skey, you get no error currently. Only if you provide invalid skey. Let's always raise an 412, no matter if the skey is missing or wrong.

### DIFF
--- a/core/tasks.py
+++ b/core/tasks.py
@@ -369,8 +369,7 @@ class TaskHandler:
             raise errors.Unauthorized()
         skel = task.dataSkel()
         skey = kwargs.get("skey", "")
-        if len(kwargs) == 0 or skey == "" or not skel.fromClient(kwargs) or (
-            "bounce" in kwargs and kwargs["bounce"] == "1"):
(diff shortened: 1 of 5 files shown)


ArneGudermann (approved):

core/tasks.py
```diff
@@ -369,8 +369,7 @@ def execute(self, taskID, *args, **kwargs):
             raise errors.Unauthorized()
         skel = task.dataSkel()
         skey = kwargs.get("skey", "")
-        if len(kwargs) == 0 or skey == "" or not skel.fromClient(kwargs) or (
-            "bounce" in kwargs and kwargs["bounce"] == "1"):
+        if len(kwargs) == 0 or not skel.fromClient(kwargs) or kwargs.get("bounce") == "1":
```
phorward, line 372: Generally, I would like to propose here to test the "bounce"-parameter on `conf["viur.bone.boolean.str2true"]` but this should be done in a separate PR.

phorward (approved):

PR #551, opened 2022-11-16
Add ModuleConfig with settings for vi

ArneGudermann:
This pr provides a new internal module for the new vi. 
It also provides new configuration variables and a new pubilc setting method (`/vi/settings`)


core/config.py
```diff
@@ -178,4 +178,15 @@
...
     "viur.version": None,
+
+    # Admin Parts
+    "admin.name": "Viur Administartion--Testr",
+    "admin.logo": "",
+    "admin.login.background": "/vi/login-background-2.jpg",
+    # URL for the Logo over the VI Login screen
+    "admin.login.logo": "/vi/logo_test.svg",
+    "admin.color.primary": "#d00f1c",
+    "admin.color.secondary": "#333333",
```
sveneberth, line 189: These shouldn't be test values. We should use general values or set them `None`.

core/modules/modulesconf.py
```diff
@@ -0,0 +1,46 @@
...
+from viur.core.skeleton import Skeleton, SkeletonInstance
+from viur.core.prototypes import List, BasicApplication
+
+
+class ModulesConfSkel(Skeleton):
+    kindName = "viur-modules-conf"
+    name = StringBone(descr=translate("modulename"), readOnly=True)
+    help_text = StringBone(descr=translate("module helptext"))
+    help_text_add = StringBone(descr=translate("add helptext"))
+    help_text_edit = StringBone(descr=translate("edit helptext"))
```
sveneberth, line 14: Can you add some description (tooltip) / comments what's the purpose of the bones? Are they for the vi?

core/modules/modulesconf.py
```diff
@@ -0,0 +1,46 @@
...
+    # seo.....
+
+
+class ModulesConf(List):
+    """
+        This module is for viur internal purposes only.
+        It lists all other modules to be able to provide them with help texts.
+    """
+    kindName = "viur-modules-conf"
+    accessRights = ["manage"]
```
sveneberth, line 24: Why is this not just _edit_, as in the prototype?
```suggestion
    accessRights = ["edit"]
```

core/modules/modulesconf.py
```diff
@@ -0,0 +1,46 @@
...
+    name = StringBone(descr=translate("modulename"), readOnly=True)
+    help_text = StringBone(descr=translate("module helptext"))
+    help_text_add = StringBone(descr=translate("add helptext"))
+    help_text_edit = StringBone(descr=translate("edit helptext"))
+    # seo.....
+
+
+class ModulesConf(List):
+    """
+        This module is for viur internal purposes only.
```
sveneberth, line 20:
```suggestion
        This module is for ViUR internal purposes only.
```

core/languages/de.py
```diff
@@ -338,5 +338,11 @@
     "internal server error": "Interner Serverfehler",
     "not implemented": "Nicht implementiert",
     "bad gateway": "Falscher Gateway",
-    "service unavailable": "Dienst nicht verfügbar"
+    "service unavailable": "Dienst nicht verfügbar",
+
+    #Moduleconfig
+    "modulename":"Modulename",
+    "module helptext" : "Module Hilfetext",
```
sveneberth, line 345: Isn't the German word for _module_  _Modul_?
```suggestion
    "modulename": "Modulname",
    "module helptext" : "Modul Hilfetext",
```

core/render/vi/__init__.py
```diff
@@ -163,11 +163,23 @@ def index(*args, **kwargs):
 index.exposed = True
 
 
+def get_settings():
+    fields = ["admin.name", "admin.logo", "admin.login.background", "admin.login.logo", "admin.color.primary",
+              "admin.color.secondary"]
+    res = {}
+    for field in fields:
+        res[field] = conf.get(field, "")
```
sveneberth, line 171: Couldn't we just load the values from the config with the `admin.` prefix? All values in the config should have a default values, which means we don't need the `.get()`.

```suggestion
     fields = {key: values for key, values in conf.items()
              if key.startswith("admin.")}
```

core/config.py
```diff
@@ -178,4 +178,15 @@
 
     # Will be set to viur.core.version.__version__ in viur.core.__init__
     "viur.version": None,
+
+    # Admin Parts
+    "admin.name": "Viur Administartion--Testr",
+    "admin.logo": "",
+    "admin.login.background": "/vi/login-background-2.jpg",
```
phorward, line 185: Please provide a more reasonable default configuration here:
```suggestion
    # Administration tool configuration
    "admin.name": "ViUR",  # I would suggest to use utils.projectID here...
    "admin.logo": "",
    "admin.login.background": "",
```

core/languages/de.py
```diff
@@ -338,5 +338,11 @@
...
+    "add helptext" : "Hinzufügen Hilfetext",
+    "edit helptext" : "Editieren Hilfetext",
```
phorward, line 347: Provide a reasonable default translation here than just a german replacement.

core/modules/modulesconf.py
```diff
@@ -0,0 +1,46 @@
+from viur.core.bones import StringBone
+from viur.core.tasks import StartupTask
+from viur.core import conf, db
+from viur.core.i18n import translate
+from viur.core.skeleton import Skeleton, SkeletonInstance
+from viur.core.prototypes import List, BasicApplication
+
+
+class ModulesConfSkel(Skeleton):
+    kindName = "viur-modules-conf"
```
phorward, line 10: I would like to recommend a singluar naming? As this is the kind defining conf for a module each entry.
```suggestion
    kindName = "viur-module-conf"
```

sveneberth (changes requested):
Hey @ArneGudermann i like this feature. However, I would still have a few comments:

- How can I add project-specific bones to the skeleton? 

- How can I get the skeleton of a module quickly and easily? Maybe if necessary also from Jinja2?
  I imagine something like this: 
  ```py
  @classmethod
  def get_by_module_name(cls, module_name: str):
      db_key = db.Key("viur-modules-conf", f"{module_name}")
      skel = getattr(conf["viur.mainApp"], "_modulesconf").viewSkel()
      return skel
  ```
  So I could write
  ```py
  from viur.core.modules.modulesconf import ModulesConf
  print(ModulesConf.get_by_module_name("shop")
  ```
  
- I don't really like the name _"conf"_. Why save two characters and not use _"config"_? And since the other module are also singular, I would like to use _module_ instead of _"modules"_ here too. So _ModuleConfig_ or _module-config_.

core/modules/modulesconf.py
```diff
@@ -0,0 +1,46 @@
...
+    def canAdd(self):
+        return False
+
+    def canDelete(self, skel: SkeletonInstance) -> bool:
+        return False
+
+
+@StartupTask
+def read_all_modules():
+    for module_name in dir(conf["viur.mainApp"].vi):
```
phorward, line 35: What happens to modules that are being removed? The config then stays... if the config is used for general module retrieval, it should be removed.

core/render/vi/__init__.py
```diff
@@ -163,11 +163,23 @@ def index(*args, **kwargs):
 index.exposed = True
 
 
+def get_settings():
+    fields = ["admin.name", "admin.logo", "admin.login.background", "admin.login.logo", "admin.color.primary",
+              "admin.color.secondary"]
```
phorward, line 168: Please implement a prefix match on `field.startswith("admin.")` rather than a fixed list of fields which cannot be configured from outside.

phorward (changes requested):

core/modules/modulesconf.py
sveneberth: Deleting entities is a very bad idea here. If you add a new module and deploy it as version X, you can already prepare the data there. But in standard/live version Y the module is still unknown, so it would be deleted. X would create it again at the next instance start and Y delete it again ….
The same happens when you do a rollback after deleting a module or work with different states in a development system.

sveneberth (changes requested):
Please see my comment https://github.com/viur-framework/viur-core/pull/551#discussion_r1024457157

and the concerns in https://github.com/viur-framework/viur-core/pull/551#pullrequestreview-1182352702

core/modules/modulesconf.py
ArneGudermann: Maybe we can take out the delete part for now. In my opinion it is not so bad if the record remains in the database.

phorward (dismissed):
IMHO this looks fine for now. There might be some tweaks to be done later.

core/modules/modulesconf.py
XeoN-GHMB: Yes, these are additional tooltips in the administration. For example, general data maintenance notes or a general description of the module.

core/render/vi/__init__.py
```diff
@@ -163,11 +163,20 @@ def index(*args, **kwargs):
 index.exposed = True
 
 
+def get_settings():
```
sveneberth, line 166:
```suggestion
@exposed
def get_settings():
```

ArneGudermann:
@sveneberth I check if we can extend the skeleton.  It works! Like so:
 
image

And the vi render it also out with no problem:
image

phorward:
In the viur meeting, we decided to change the help text StringBones into TextBones, and limit these to specific markup tags like <b>, <i>, <u>, <a> ... @ArneGudermann

sveneberth (approved):

phorward:
@ArneGudermann please fix the conflict in core/config.py.

My advise:
- Order config keys alphabetically, so keys starting with "admin." should be in the front of the dict.
- Please provide a comment on every key for short explanation.

Thanks!

phorward (changes requested):
Please fix conflicts with current develop branch, afterwards this can be merged.

PR #552, opened 2022-11-16
Default disable `viur.render.json.downloadUrlExpiration` and `viur.downloadUrlFor.expiration`

phorward:
IMHO we bothered enough now with this, as this setting is only relevant to very few projects (I'm counting one or two), but not the gross of our everyday's stuff. This behavior is similar to ViUR2's servingurls.


core/config.py
```diff
@@ -48,7 +47,7 @@
     # Maps Domains to alternative default languages
     "viur.domainLanguageMapping": {},
     # If set, the expiration parameter for :meth:`viur.core.render.html.env.viur.downloadUrlFor` can be omitted
-    "viur.downloadUrlFor.expiration": unsetMarker,
+    "viur.downloadUrlFor.expiration": None,
```
sveneberth, line 50: Can we name this `viur.render.html.downloadUrlExpiration` like for the json render and move it below?

```suggestion
    "viur.render.html.downloadUrlExpiration": None,
```

sveneberth (changes requested):

core/config.py
phorward: I've introduced your request in b42622a0c32e5f5081930e976d2d8272c80e48e4

XeoN-GHMB (approved):

core/modules/file.py
```diff
@@ -81,7 +81,7 @@ class InjectStoreURLBone(BaseBone):
     def unserialize(self, skel, name):
         if "dlkey" in skel.dbEntity and "name" in skel.dbEntity:
             skel.accessedValues[name] = utils.downloadUrlFor(
-                skel["dlkey"], skel["name"], derived=False, expires=conf["viur.downloadUrlFor.expiration"]
+                skel["dlkey"], skel["name"], derived=False, expires=conf["viur.render.json.downloadUrlExpiration"]
```
sveneberth, line 84: Hmm. The correct replacement of _viur.downloadUrlFor.expiration_ would be _viur.render.html.downloadUrlExpiration_... 
But actually this a third case. _viur.downloadUrlFor.expiration_ or _viur.InjectStoreURLBone.downloadUrlFor.expiration_ would be correct but maybe it's more confusing and IMO this should be optional controllable on the bone. 

---
I am undecided.
I don't mind leaving it like that for now, and then we can think about it again later. Then we can merge the PR for the time being.

phorward: This is correct and we leave it now as is.

Issue #550, opened 2022-11-12
Ugly way to find out if running in deferred task?


phorward:
Currenty (in viur-core >= 3.3.0) this is the only valid way to find out if code is running within a deferred task:
```python
getattr(utils.currentRequest.get(), "DEFERRED_TASK_CALLED", False)
```
Becuase `DEFERRED_TASK_CALLED` is only a member of the Request when it was invoked from a deferred task (`@CallDeferred`).

In viur-core < 3.3.0, the variable `DEFERRED_TASK_CALLED` was also mispelled `DEFERED_TASK_CALLED`, which was fixed by #508.

This issue is a request for providing a flag `is_deferred` (likely to `isDevelopmentServer`) in the current Request object, to decide whether code is executed within a deferred call or not.

phorward:
Resolved by #556

PR #556, opened 2022-11-17
Add is_deferred flag to currentRequest

ArneGudermann:
This pr solve the issue #550 and add a (is_deferred) flag to the currentRequest Object

### DIFF
--- a/core/tasks.py
+++ b/core/tasks.py
@@ -434,7 +434,6 @@ def CallDeferred(func):
                     return func(self, *args, **kwargs)
 
             if req:
-                req.is_deferred = True
                 req.pendingTasks.append(task)  # This property only exists on development server!
             else:
                 # Warmup request or something - we have to call it now as we can't defer it :/
--- a/core/tasks.py
+++ b/core/tasks.py
@@ -426,7 +426,6 @@ def CallDeferred(func):
         if not queueRegion:
             # Run tasks inline
             logging.debug(f"{func=} will be executed inline")
-
             @wraps(func)
             def task():
                 if self is __undefinedFlag_:
@@ -435,6 +434,7 @@ def CallDeferred(func):
                     return func(self, *args, **kwargs)
 
             if req:
+                req.is_deferred = True
                 req.pendingTasks.append(task)  # This property only exists on development server!
             else:
                 # Warmup request or something - we have to call it now as we can't defer it :/
--- a/core/request.py
+++ b/core/request.py
@@ -352,6 +352,7 @@ class BrowseHandler():  # webapp.RequestHandler
                     }
                 )
         if self.isDevServer:
+            self.is_deferred = True
             while self.pendingTasks:
                 task = self.pendingTasks.pop()
                 logging.info("Running task directly after request: %s" % str(task))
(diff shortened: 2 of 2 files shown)


phorward (approved):
Fine! :+1:

phorward (commented):
One question before committing: Does this also work locally? IMHO this is not the case, so maybe we have to implement the method described in #550 as fallback?

ArneGudermann:
What do you mean with work locally?  We run locally all Tasks inline and the flag is allways false.

sveneberth:
> What do you mean with work locally? We run locally all Tasks inline and the flag is allways false.

But you can use a local task queue emulator.

phorward:
> What do you mean with work locally? We run locally all Tasks inline and the flag is allways false.

Sure, but maybe I want to test some specific behavior based on this flag, e.g. access rights that should be ignored when there's no user and the tasks runs deferred.

My proposal would be to set is_deferred to True also when the function runs locally inline, to reproduce equal behaviors on both runtime environments. This would safe a lot of debugging time related to such edge cases.

core/tasks.py
```diff
@@ -435,6 +434,7 @@ def task():
                     return func(self, *args, **kwargs)
 
             if req:
+                req.is_deferred = True
```
sveneberth, line 437: That's the wrong point. If you start a deferred task (call the method) the current request will be immediately marked as deferred. This would be wrong for statements after the call.


Short example:
```py
	@exposed
	def test_deferred(self):
		logging.debug("is_deferred: %r (should not)", currentRequest.get().is_deferred)
		self._test_deferred()
		logging.debug("is_deferred: %r (should not)", currentRequest.get().is_deferred)
		
	@tasks.CallDeferred
	def _test_deferred(self):
		logging.debug("I'm a deferred task!")
		logging.debug("is_deferred: %r (should)", currentRequest.get().is_deferred)
```
results into
```
DEBUG    2022-11-22 10:15:04,295 index.py:455] is_deferred: False (should not)
DEBUG    2022-11-22 10:15:04,296 tasks.py:419] make_deferred func=<function Index._test_deferred at 0x7f9755f755a0>, self=<modules.index.Index object at 0x7f975ac24160>, args=(), kwargs={}, queue='default', taskargs={'countdown': None, 'eta': None, 'name': None, 'target': None, 'retry_options': None}
DEBUG    2022-11-22 10:15:04,296 tasks.py:428] func=<function Index._test_deferred at 0x7f9755f755a0> will be executed inline
DEBUG    2022-11-22 10:15:04,296 index.py:457] is_deferred: True (should not)        <---- This is now deferred too
INFO     2022-11-22 10:15:04,404 request.py:358] Running task directly after request: <function Index._test_deferred at 0x7f9755e00670>
DEBUG    2022-11-22 10:15:04,405 index.py:466] I'm a deferred task!
DEBUG    2022-11-22 10:15:04,405 index.py:467] is_deferred: True (should)
[2022-11-22 10:15:04] [200] [GET] /test_deferred
```

---

I suggest to add 
```py
self.is_deferred = True
```
after
https://github.com/viur-framework/viur-core/blob/6b1182d9482dd962332983763cda81f761cf184f/core/request.py#L347

sveneberth (approved):

phorward (approved):

PR #562, opened 2022-11-23
Move variables to config

ArneGudermann:
This draft move some vars from the utils in the config.
Beause we have some variables double set like:
https://github.com/viur-framework/viur-core/blob/198478b97d1fc9513b4bea4854164bd7e2ebb9ff/core/request.py#L167
and:
https://github.com/viur-framework/viur-core/blob/198478b97d1fc9513b4bea4854164bd7e2ebb9ff/core/utils.py#L32

we can combine this in the config and fill it when a instance start.

PR #569, opened 2022-12-01
Providing datastore index information with modules

ArneGudermann:
This Draft can be separate in two groups.

1. When now an instance start we read the `index.yaml `and write the indexes on corresponding modules. Than we can read out the indexes for the frontend to filter better.

2. When the PR  https://github.com/viur-framework/viur-datastore/pull/19 is merged we can read out the order from a Query and send it also to the frontend. So we can use this in the new VI for better filtering too.


core/__init__.py
```diff
@@ -34,6 +36,18 @@
 from viur.core import logging as viurLogging  # Initialize request logging
 
 
+def load_indexes_from_file():
+    try:
+        with open(os.path.join(utils.coreBasePath, 'index.yaml'), 'r') as file:
```
phorward, line 41: Please use double quotes for strings.
```suggestion
        with open(os.path.join(utils.coreBasePath, "index.yaml"), "r") as file:
```

phorward: `utils.coreBasePath` can also be moved to the conf, respectively #562. Will create an issue on this.

core/__init__.py
```diff
@@ -34,6 +36,18 @@
...
+            indexes = yaml.safe_load(file)
+            indexes = indexes.get("indexes", [])
+            for index in indexes:
+                index["properties"] = [_property["name"] for _property in index["properties"]]
+    except FileNotFoundError as e:
+        logging.error("index.yaml not found")
+        return
```
phorward, line 48: Is this case really an error?
Or better have a warning, and return something empty which doesn't raise a `TypeError: argument of type 'NoneType' is not iterable` behind the call below?
```suggestion
        logging.warning("index.yaml not found")
        return {}
```

core/__init__.py
```diff
@@ -34,6 +36,18 @@
 from viur.core import logging as viurLogging  # Initialize request logging
 
 
+def load_indexes_from_file():
```
phorward, line 39: Please add a docstring and type annotations.

core/__init__.py
```diff
@@ -164,6 +179,11 @@ class ExtendableObject(object):
                 moduleInstance = moduleClass(moduleName, modulePath)
                 # Attach the module-specific or the default render
                 moduleInstance.render = render.get(moduleName, render["default"])(parent=moduleInstance)
+                module_indexes = []
+                for index in indexes:
+                    if index["kind"] == moduleName:
+                        module_indexes.append(index)
```
phorward, line 185: You should consider of doing this stuff already in `load_indexes_from_file`. It might return a dict which already contains all index definitions grouped by module, and would run with a better performance, as objects are shared between equal modules in the render.

core/render/vi/__init__.py
```diff
@@ -100,6 +100,8 @@ def dumpConfig(adminTree):
                         tmp = v.copy()
                         tmp["name"] = str(tmp["name"])
                         adminConfig[key]["views"].append(tmp)
+        if "indexes" in dir(app) and key in adminConfig:
+            adminConfig[key]["indexes"] = app.indexes
```
phorward, line 104: This would increase performance in several ways.
```suggestion
        if key in adminConfig and (indexes := getattr(app, "indexes", None)):
            adminConfig[key]["indexes"] = indexes
```

phorward (changes requested):
Please check out my comments and suggestions.

core/__init__.py
```diff
@@ -25,15 +25,41 @@
...
 from types import ModuleType
-from typing import Callable, Dict, Union
+from typing import Callable, Dict, Union, List
 from viur.core import session, errors, i18n, request, utils
 from viur.core.config import conf
 from viur.core.tasks import TaskHandler, runStartupTasks
 from viur.core import logging as viurLogging  # Initialize request logging
 
 
+def load_indexes_from_file() -> Dict[str, List]:
```
sveneberth, line 39: I think this is misplaced here. We should generally try to reduce the length of `__init__` — This should contain only the imports and basic initializations.

Maybe we should move it to the `utils`?

sveneberth (changes requested):

core/__init__.py
ArneGudermann: Mhh is `utils` the right place for it, because we only use this once in `__init__`.

phorward: I'm with @ArneGudermann.

Generally, the entire stuff from `__init__.py` should in perspective be moved into something like `route.py` or so.

phorward:
@ArneGudermann if the pull request is ready to merge, you have to convert the draft into "ready for review".

core/__init__.py
ArneGudermann: Yes this makes sense to me. But lets make a new PR for this ok ?

sveneberth (approved):

Issue #571, opened 2022-12-06
missing hmacKey in main.py crashes the core


XeoN-GHMB:
```
File "/home/x/.pyenv/versions/3.10.5/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/home/x/projects/viur/deploy/main.py", line 75, in <module>
    app = core.setup(modules, render)
  File "/home/x/.local/share/virtualenvs/viur/lib/python3.10/site-packages/viur/core/__init__.py", line 251, in setup
    logging.info("Creating new viur-conf")

AttributeError: module 'viur.core.logging' has no attribute 'info'
```

PR #573, opened 2022-12-06
`import logging` must stay behind other imports in `__init__.py`

ArneGudermann:
This PR should fix this https://github.com/viur-framework/viur-core/issues/571. it imports the standart logging module after the core logging module. It work for me now: 
image


sveneberth (approved):
It works, even though I don't really understand how the `from viur.core import session, errors, i18n, request, utils` affects the logging reference. 
This could be some kind of recursion issue, since the core imports from itself. Do you @ArneGudermann know the reason?

phorward (changes requested):
A short reason, why this import has to be there, and nowhere else, helps to avoid a future re-arrangement of imports causing the same bug again.

PR #576, opened 2022-12-06
Update viur-datastore==1.3.5 & CHANGELOG rewrite

phorward:
Hi, I've rewritten the CHANGELOG for v3.3.0 to define the specific changes more precisely.

Generally, we should try to provide given commit messages, prefixed by

- "Added"
- "Fixed"
- "Removed"
- "Refactored"

in future.


sveneberth:
> Generally, we should try to provide given commit messages, prefixed by
> 
> * "Added"
> * "Fixed"
> * "Removed"
> * "Refactored"
> 
> in future.

Please don't write commit messages in the past tense. A commit message describes what a patch changes on the codebase - It belongs together and happens at the same time. This is not a report like "someone made changes in the past and now I'm describing them".

From the [git documentation](https://git-scm.com/docs/SubmittingPatches#describe-changes): 
> Describe your changes in imperative mood, e.g. "make xyzzy do frotz" instead of "[This patch] makes xyzzy do frotz" or "[I] changed xyzzy to do frotz", as if you are giving orders to the codebase to change its behavior. 

But in general it sounds reasonable to prefix the kind of change. This should then be stated in the [CONTRIBUTING.md](https://github.com/viur-framework/viur-core/blob/main/CONTRIBUTING.md).

sveneberth (commented):
I think it would be more clear to group entries by e.g. "New Features, Improved, Deprecated, Removed" like the PSF and many others does it: https://docs.python.org/3.11/whatsnew/3.11.html

phorward:
@sveneberth Jaja, but the changes in this commit is ok?

XeoN-GHMB:
I try to orientate myself by this:
https://www.conventionalcommits.org/en/v1.0.0/

XeoN-GHMB (approved):

phorward:
> I try to orientate myself by this: https://www.conventionalcommits.org/en/v1.0.0/

Thanks. I'm no fan of the word "chore" which appears very often in the changelog then. Anyway, this is a democratic project. We should consider this for any commits that follow 3.3.0

sveneberth (approved):
> @sveneberth Jaja, but the changes in this commit is ok?

I still don't like the format of the CHANGELOG. Ordering mostly by creationtime is in my opinion not really helpful. We can think about that later, it's low-prio.

ArneGudermann (approved):

PR #577, opened 2022-12-06
Refactoring ModuleConf

phorward:
Hi there, this is a small refactoring of the new ModuleConf-module with the following changes:

- Standardized skeleton definition
- Provided adminInfo with a default "hidden" display
- Replaced f-strings that contained only one variable
- Replaced getattr-calls by default attr retrieval

### DIFF
--- a/core/modules/moduleconf.py
+++ b/core/modules/moduleconf.py
@@ -59,7 +59,7 @@ class ModuleConf(List):
         return False
 
     @classmethod
-    def get_by_module_name(cls, module_name: str):
+    def get_by_module_name(cls, module_name: str) -> Union[None, SkeletonInstance]:
         db_key = db.Key("viur-module-conf", module_name)
         skel = conf["viur.mainApp"]._moduleconf.viewSkel()
         if not skel.fromDB(db_key):
--- a/core/modules/moduleconf.py
+++ b/core/modules/moduleconf.py
@@ -11,15 +11,31 @@ from viur.core.prototypes import List, BasicApplication
 
 class ModuleConfSkel(Skeleton):
     kindName = "viur-module-conf"
+
     _valid_tags = ['b', 'a', 'i', 'u', 'span', 'div', 'p', 'ol', 'ul', 'li', 'abbr', 'sub', 'sup', 'h1', 'h2', 'h3',
                    'h4', 'h5', 'h6', 'br', 'hr', 'strong', 'blockquote', 'em']
     _valid_html = _defaultTags
     _valid_html["validTags"] = _valid_tags
-    name = StringBone(descr=translate("modulename"), readOnly=True)
-    help_text = TextBone(descr=translate("module helptext"), validHtml=_valid_html)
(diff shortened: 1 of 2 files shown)


core/modules/moduleconf.py
```diff
@@ -30,6 +46,12 @@ class ModuleConf(List):
     kindName = "viur-module-conf"
     accessRights = ["edit"]
 
+    def adminInfo(self):
+        admin_info = super().adminInfo()
+        admin_info["icon"] = "icon-settings"
+        admin_info["display"] = "hidden"
```
sveneberth, line 52: If the module is hidden, how could it be edited?

phorward: It can be edited when it's subclassed and the adminInfo is corrected. I don't want to have the ModuleConf-module automatically available and visible in all existing projects now.

core/modules/moduleconf.py
```diff
@@ -38,11 +60,12 @@ def canDelete(self, skel: SkeletonInstance) -> bool:
 
     @classmethod
     def get_by_module_name(cls, module_name: str):
```
ArneGudermann, line 62:
```suggestion
    def get_by_module_name(cls, module_name: str) -> Union[None, SkeletonInstance]:
```
We can add a Typehinte here
and import it in the top  ` from typing import  Union`

core/modules/moduleconf.py
```diff
@@ -30,19 +46,26 @@ class ModuleConf(List):
...
         return False
 
     def canDelete(self, skel: SkeletonInstance) -> bool:
         return False
 
     @classmethod
-    def get_by_module_name(cls, module_name: str):
-        db_key = db.Key("viur-module-conf", f"{module_name}")
-        skel = getattr(conf["viur.mainApp"], "_moduleconf").viewSkel()
+    def get_by_module_name(cls, module_name: str) -> Union[None, SkeletonInstance]:
```
sveneberth, line 62: Please add the `Union` import as [recommended](https://github.com/viur-framework/viur-core/pull/577#discussion_r1041903602) by @ArneGudermann 

If you don't like imports😅 you could write:
```suggestion
    def get_by_module_name(cls, module_name: str) -> None | SkeletonInstance:
```

sveneberth (changes requested):

core/modules/moduleconf.py
phorward: > If you don't like imports

I hate imports...

core/modules/moduleconf.py
phorward: Ok, proposal: Can we provde the adminInfo via config? 

So the default would be 
```python
conf["viur.module.conf.admin_info"] = {
    "name": "Module Conf",
    "icon": "icon-settings",
    "display": "hidden"
}
```
but it can easily modified project-specific.

phorward: 4ae1bc5 is my proposal on this @sveneberth

PR #581, opened 2022-12-07
Improve the linter workflow: Use error annotations

sveneberth:
Instead of just printing the errors in the actions log, we can use error annotations. They are displayed in the code view.
With `--format` we can directly generate the required format.

[Example](https://github.com/viur-framework/viur-core/pull/580):
image

For reference:
- https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-creating-an-annotation-for-an-error
- https://pycodestyle.pycqa.org/en/latest/intro.html#example-usage-and-output


.github/workflows/pep8.yaml
```diff
@@ -28,15 +28,14 @@ jobs:
 
       - name: Install dependencies
         run: |
-          python -m pip install -U pip
```
sveneberth, line 31: `pip` seems to be always up2date in the GitHub action environment.

phorward (approved):

Issue #570, opened 2022-12-05
`utils.coreBasePath` can also be moved to the conf


phorward:
`utils.coreBasePath` can also be moved to the conf, respectively #562. Will create an issue on this.

_Originally posted by @phorward in https://github.com/viur-framework/viur-core/pull/569#discussion_r1040009506_

phorward:
fixed by #582

PR #582, opened 2022-12-08
Move projectBasePath and coreBasePath

ArneGudermann:
This PR move coreBasePath and projectBasePath from the utils to  the config related to #570


core/config.py
```diff
@@ -44,6 +45,11 @@ def __setitem__(self, key, value):
 __project_id = google.auth.default()[1]
 __version = os.getenv("GAE_VERSION")
 
+# Determine our basePath (as os.getCWD is broken on appengine)
+__project_base_path = str(Path().absolute())
+
+__core_base_path = globals()["__file__"].replace("/viur/core/config.py", "")
```
phorward, line 51: Direct use of `__file__` should work as well (please verify!)
```suggestion
__core_base_path = __file__.replace("/viur/core/config.py", "")
```

core/render/__init__.py
```diff
@@ -4,16 +4,14 @@
...
-
+from viur.core import conf
 
 def isViAvailable():
-    from viur.core.utils import projectBasePath
-    return path.exists(path.join(projectBasePath, "vi", "main.html"))
+    return path.exists(path.join(conf["viur.instance.core_base_path"], "vi", "main.html"))
 
 
 def isAdminAvailable():
```
phorward, line 13: General question: Is this only a relict of the Pyodie-based QT-Admin-Test?

phorward (changes requested):

core/config.py
ArneGudermann: yes `__file__` works too.

core/render/__init__.py
ArneGudermann: No it`s used here https://github.com/viur-framework/viur-core/blob/221743136e867911bfa0fa994ed9db4ae820609d/core/render/vi/__init__.py#L152-L155 in example.

phorward: Nevermind. We should kill this later on.

phorward (approved):

sveneberth (changes requested):

core/config.py
```diff
@@ -44,6 +45,11 @@ def __setitem__(self, key, value):
 __project_id = google.auth.default()[1]
 __version = os.getenv("GAE_VERSION")
 
+# Determine our basePath (as os.getCWD is broken on appengine)
+__project_base_path = str(Path().absolute())
+
+__core_base_path = __file__.replace("/viur/core/config.py", "")
```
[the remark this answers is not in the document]
phorward: So Path implements `__str__`? Then it's ok.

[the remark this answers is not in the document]
phorward:
```
$ cat test.py 
import pathlib

p = pathlib.Path().absolute()
print(p + "lol")
$ python test.py 
  File "/tmp/test.py", line 4, in <module>
    print(p + "lol")
TypeError: unsupported operand type(s) for +: 'PosixPath' and 'str'
```
Doesn't work as I expected, anyway I don't think it breaks anything and we can make it so.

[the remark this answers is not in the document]
sveneberth: @phorward Yes, but not all binary operators. But it should in general work for sth. like `open()`

core/render/html/default.py
```diff
@@ -101,13 +102,13 @@ def getTemplateFileName(self, template: str, ignoreStyle: bool = False) -> str:
                       template + ".html"]
         for fn in fnames:  # check subfolders
             prefix = template.split("_")[0]
-            if os.path.isfile(os.path.join(utils.projectBasePath, htmlpath, prefix, fn)):
+            if os.path.isfile(os.path.join(conf["viur.instance.core_base_path"], htmlpath, prefix, fn)):
```
sveneberth, line 105: This is not correctly substituted, it's the project path.

sveneberth (changes requested):
There are some issues, see https://github.com/viur-framework/viur-core/pull/598

sveneberth (dismissed):

phorward:
@sveneberth are you happy with this now?

phorward (approved):

sveneberth (approved):

PR #589, opened 2022-12-11
fix: Improvements for CONTRIBUTING.md

phorward:
- Added this on viur-core 3.3 release

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,7 +34,11 @@ In case you have appropriate permissions, a release can be done this way:
   - Release the package
     - PyPI: `pipenv run release`
     - TestPyPI: `pipenv run develop`
-- When all went well, finally create a tag equally to the version number in `core/version.py` 
+- When all went well, finally create a tag equally to the version number in `core/version.py`
+- In case this is a new minor release, remove all release candidate tags using
+  ```
+  for i in `git tag`; do if [[ $i == *-rc* ]]; then git tag -d $i; git push origin :$i; fi; done
+  ```
 
 ## Branches
 


CONTRIBUTING.md
```diff
@@ -34,7 +34,11 @@ In case you have appropriate permissions, a release can be done this way:
   - Release the package
     - PyPI: `pipenv run release`
     - TestPyPI: `pipenv run develop`
-- When all went well, finally create a tag equally to the version number in `core/version.py` 
+- When all went well, finally create a tag equally to the version number in `core/version.py`
+- In case this is a new minor release, remove all release candidate tags using
```
sveneberth, line 38: Why? I think this is a very bad idea.

Due to the long release cycles of viur-core, several projects have used the release candidates in production. If you delete the rc version, a new deployment with a hotfix would not be possible without adjustments for breaking changes (if this was a early RC). Or deployments are not reproducible, which destroys the principle of a CI/CD. Please keep them, especially on PyPi.


Jinja, for example, keeps them as well: 
- https://pypi.org/project/Jinja2/#history 
- https://github.com/pallets/jinja/tags?after=3.0.1

sveneberth (changes requested):

CONTRIBUTING.md
phorward: No! Release candidates are not intended to be used in production. This is only sometimes necessary, when bleeding edge features are required by a project, but such projects are encouraged to update to the official release version ASAP after the release, especially for sake of bugs and security.

phorward: Let's discuss this in the VIUR Runde.

phorward: Ok we decided to remove release candidates after the next minor release. So 3.3rc* will be removed when 3.4 is released, and 3.4rc* is removed when 3.5 is released.

phorward:
Let's just not delete anything from PyPI, except when really necessary.

PR #590, opened 2022-12-12
Change path of index.yaml in load_indexes_from_file

ArneGudermann:
This PR fix the bug when the core is  imported as package and not as a submodule the index.yaml can't be found.


phorward (approved):
Hotfixes against main.

Issue #587, opened 2022-12-09
Define standard User-Roles based on Access Rights


Grashalmbeisser:

phorward:
Implemented by #736

PR #591, opened 2022-12-12
Code clean-up for core user module

phorward:
- Cleaned imports
- Cleaned obsolete notations (super-calls, inherit from object)
- Renamed classes to be PEP-8 compliant
- Cleaned code-style
- Deprecation warning for `userSkel` use, rather than `UserSkel`

### DIFF
--- a/core/modules/user.py
+++ b/core/modules/user.py
@@ -290,11 +290,11 @@ class UserPassword:
                 return self.userModule.render.edit(skel, self.passwordRecoveryStep2Template)  # Let's try again
 
             # If we made it here, the key was correct, so we'd hopefully have a valid user for this
-            uSkel = self.viewSkel().all().filter(
+            user_skel = self.viewSkel().all().filter(
                 "name.idx =", session["user.auth_userpassword.pwrecover"]["name"]
             ).getSkel()
 
-            if not uSkel:  # This *should* never happen - if we don't have a matching account we'll not send the key.
+            if not user_skel:  # This *should* never happen - if we don't have a matching account we'll not send the key.
                 session["user.auth_userpassword.pwrecover"] = None
                 return self.userModule.render.view(
                     skel=None,
--- a/core/modules/user.py
+++ b/core/modules/user.py
@@ -1,113 +1,103 @@
+import datetime
+import hashlib
+import hmac
+import json
+import logging
+import os
+import time
+import warnings
 from viur.core.prototypes.list import List
-from viur.core.skeleton import Skeleton, RelSkel, skeletonByKind
 from viur.core.bones import *
-from viur.core import utils, email
-from viur.core.bones.base import ReadFromClientErrorSeverity, UniqueValue, UniqueLockMethod
+from viur.core import conf, db, email, errors, i18n, securitykey, session, skeleton, tasks, utils, exposed, forceSSL
 from viur.core.bones.password import pbkdf2
-from viur.core import errors, conf, securitykey
-from viur.core.tasks import StartupTask, CallDeferred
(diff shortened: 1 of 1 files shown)


core/modules/user.py
```diff
@@ -297,8 +288,12 @@ def pwrecover(self, *args, **kwargs):
...
                         reason=self.passwordRecoveryKeyInvalid)
                 return self.userModule.render.edit(skel, self.passwordRecoveryStep2Template)  # Let's try again
+
             # If we made it here, the key was correct, so we'd hopefully have a valid user for this
-            uSkel = userSkel().all().filter("name.idx =", session["user.auth_userpassword.pwrecover"]["name"]).getSkel()
+            uSkel = self.viewSkel().all().filter(
+                "name.idx =", session["user.auth_userpassword.pwrecover"]["name"]
+            ).getSkel()
+
             if not uSkel:  # This *should* never happen - if we don't have a matching account we'll not send the key.
```
ArneGudermann, line 297: Maybe we can rename the variable here and lower in the code like :
https://peps.python.org/pep-0008/#function-and-variable-names
```suggestion
            user_skel = self.viewSkel().all().filter(
                "name.idx =", session["user.auth_userpassword.pwrecover"]["name"]
            ).getSkel()

            if not user_skel:  # This *should* never happen - if we don't have a matching account we'll not send the key.
```

ArneGudermann: Maybe we should discuss if we rename the most vars in this they all look like:
```python
    registrationEnabled = False
    registrationEmailVerificationRequired = True
    registrationAdminVerificationRequired = True
    verifySuccessTemplate = "user_verify_success"
    verifyEmailAddressMail = "user_verify_address"
    verifyFailedTemplate = "user_verify_failed"
    passwordRecoveryTemplate = "user_passwordrecover"
    passwordRecoveryMail = "user_password_recovery"
    passwordRecoveryAlreadySendTemplate = "user_passwordrecover_already_sent".......
```

But I dont't now if this are breaking changes.

phorward: Hello Arne,

I'm totally with you. I think these special templates aren't used that much in current VIUR3 systems.

Anyway, if we provide a generalized deprecation handling for module members, we can do it without totally breaking the system. Such a generalized handling can be placed in the BasicApplication prototype, and customized via a replacement dict. This replacement dict is filled when renaming module-specific members (including functions!).

I'll make an issue on this and start a PR which we can use for further discussion.

phorward: I've opened issue #594 on this. This change including a renaming is bigger and also involves rendering code, therefore I wouldn't integrate it within this pull request. I've started this pull request as preliminary work for #587.

ArneGudermann:
Hello @phorward. Now I'm fine with this but I have one last suggestion can you please change this 
`skel.password = PasswordBone(descr="Passwort", required=False)`
to:
`skel.password.required = False`
here :https://github.com/viur-framework/viur-core/blob/204f520b56fad4298511dc058a43f968799e015d/core/modules/user.py#L686

skel.password is already and PasswordBone in this case.

phorward:
> Hello @phorward. Now I'm fine with this but I have one last suggestion can you please change this `skel.password = PasswordBone(descr="Passwort", required=False)` to: `skel.password.required = False` here :

Wilco. I'll use this pull request to integrate https://github.com/viur-framework/server/pull/24#discussion_r149664347 as well, IMHO this was a senseless discussion.

phorward:
@ArneGudermann please re-check the latest changes. Afterwards we can approve it.

ArneGudermann (approved):

PR #592, opened 2022-12-12
Provide colorized local debug

phorward:
This is a feature initially provided by @XeoN-GHMB, thanks to him!

This output
shot-2022-12-15_13-02-51

is generated by 
```python
@exposed
def test(self):
    for l in (logging.NOTSET, logging.DEBUG, logging.INFO, logging.WARNING, logging.ERROR, logging.CRITICAL):
        logging.log(l, "Hello World")
```


ArneGudermann (approved):

core/logging.py
```diff
@@ -52,27 +46,95 @@ def emit(self, record: logging.LogRecord):
...
+
+    def format(self, record):
+        if "pathname" in record.__dict__.keys():
+            # truncate the pathname
+            if "/deploy" in record.pathname:
+                pathname = record.pathname.split("/deploy/")[1]
+            else:
+                pathname = record.pathname
+                if len(pathname) > 20:
+                    pathname = ".../" + "/".join(pathname.split("/")[-3:])
```
sveneberth, line 77: Imagine you have a module in a very depth package like: `google.cloud.resourcemanager_v3.services.projects.client`. 
You would get `.../services/projects/client`, for which it is not quite clear where it belongs. 
If we remove the middle part, you can see the root package (_google_) too: `google/.../projects/client`

```suggestion
                    parts = pathname.split("/")
                    del parts[1:-3]
                    parts.insert(1, "...")
                    pathname = "/".join(parts)
```

core/logging.py
```diff
@@ -52,27 +46,95 @@ def emit(self, record: logging.LogRecord):
...
+
+            record.pathname = pathname
+
+        levelname = record.levelname
+
+        if levelname in ViURLocalFormatter.LEVELS:
+            record.levelname = ViURLocalFormatter.COLOR_SEQ % (30 + ViURLocalFormatter.LEVELS[levelname]) \
+                               + levelname + ViURLocalFormatter.RESET_SEQ
```
sveneberth, line 85: I like it when the whole line is colored instead of just parts of it. While reading the message (on the right), you notice the color and know what kind of logging it is without having to read the logging level (further to the left) itself.
Also, this has the nice effect of adding a kind of even/odd style (often used in tables) to separate rows and improve readability (of course only works if the level are different).

core/logging.py
```diff
@@ -52,27 +46,95 @@ def emit(self, record: logging.LogRecord):
...
+    handler = ViURDefaultLogger(client, name="ViUR-Messages", resource=Resource(type="gae_app", labels={}))
+    logger.addHandler(handler)
+
+    formatter = logging.Formatter("%(levelname)-8s %(asctime)s %(filename)s:%(lineno)s] %(message)s")
+
+# Use ViURLocalFormatter for local debug message formatting
+else:
+    formatter = ViURLocalFormatter(
+        f"[%(asctime)s] %(pathname)s:%(lineno)d [%(levelname)s] %(message)s",
+        datefmt="%Y-%m-%d %H:%M:%S"
```
sveneberth, line 130: Can we please keep the milliseconds? They are very helpful for performance analysis for easy benchmarking.

They are enabled by default without a custom datefmt:
```suggestion
        f"[%(asctime)s] %(pathname)s:%(lineno)d [%(levelname)s] %(message)s"
```

or explicit with a custom datefmt:
```suggestion
        f"[%(asctime)s,%(msecs)d] %(pathname)s:%(lineno)d [%(levelname)s] %(message)s",
        datefmt="%Y-%m-%d %H:%M:%S"
```

core/logging.py
```diff
@@ -52,27 +46,95 @@ def emit(self, record: logging.LogRecord):
...
     logger.removeHandler(handler)
 
-handler = ViURDefaultLogger(client, name="ViUR-Messages", resource=Resource(type="gae_app", labels={}))
-logger.addHandler(handler)
+# Plug-in ViURDefaultLogger & custom formatter when running in the cloud
+if not conf["viur.instance.is_dev_server"]:
+    handler = ViURDefaultLogger(client, name="ViUR-Messages", resource=Resource(type="gae_app", labels={}))
+    logger.addHandler(handler)
+
+    formatter = logging.Formatter("%(levelname)-8s %(asctime)s %(filename)s:%(lineno)s] %(message)s")
```
sveneberth, line 124: I think we don't need a `StreamHandler` in the cloud. So this should be not necessary. Then you can add an indent to the lines 133-135 (and move them into the local dev appserver block):
https://github.com/viur-framework/viur-core/blob/1d762d9abf071e8f5794529118c855ad7eb92bb8/core/logging.py#L133-L135

sveneberth (changes requested):

core/logging.py
phorward: That's a nice solution!

core/logging.py
phorward: Sure, we use the first suggestion, as it is more straitforward.

core/logging.py
```diff
@@ -52,27 +46,95 @@ def emit(self, record: logging.LogRecord):
...
+    LEVELS = {
+        "WARNING": YELLOW,
+        "INFO": CYAN,
+        "DEBUG": BLUE,
+        "CRITICAL": RED,
+        "ERROR": RED
+    }
+
+    def format(self, record):
+        if "pathname" in record.__dict__.keys():
```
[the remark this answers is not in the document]
phorward: I don't know, I took the code from @XeoN-GHMB. I'll push an update of this PR shortly.

core/logging.py
phorward: I've fixed this comment to colorize the entire line.

core/logging.py
phorward: The StreamHandler was moved into the local debug block

core/logging.py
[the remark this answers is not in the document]
phorward: I've fixed this in [12ae028](https://github.com/viur-framework/viur-core/pull/592/commits/12ae0289a4af6cf4e225809972684b5acfabae7a)

core/logging.py
```diff
@@ -52,27 +46,98 @@ def emit(self, record: logging.LogRecord):
...
+    # Plug-in ViURDefaultLogger
+    handler = ViURDefaultLogger(client, name="ViUR-Messages", resource=Resource(type="gae_app", labels={}))
+    logger.addHandler(handler)
 
-sh = logging.StreamHandler()
-formatter = logging.Formatter("%(levelname)-8s %(asctime)s %(filename)s:%(lineno)s] %(message)s")
-sh.setFormatter(formatter)
-logger.addHandler(sh)
+    # Custom formatter when running in the cloud
+    formatter = logging.Formatter("%(levelname)-8s %(asctime)s %(filename)s:%(lineno)s] %(message)s")
```
sveneberth, line 131: This is not required/used.
```suggestion
```

core/logging.py
```diff
@@ -52,27 +46,98 @@ def emit(self, record: logging.LogRecord):
...
+
+else:
+    # Use ViURLocalFormatter for local debug message formatting
+    formatter = ViURLocalFormatter(f"[%(asctime)s] %(pathname)s:%(lineno)d [%(levelname)s] %(message)s")
+    sh = logging.StreamHandler()
+    sh.setFormatter(formatter)
+    logger.addHandler(sh)
 
+# https://github.com/googleapis/python-logging/issues/13#issuecomment-539723753
```
sveneberth, line 140: Since `sh` is only defined in the `else:` block, we need to move this as well.

sveneberth (changes requested):
Two changes away from the merge :)

core/logging.py
XeoN-GHMB: Hi I just stumbled across this PR. Is it possible to have it configurable if the whole line or only the level should be colored? I'm thinking of terminals with white background (mac default). I think if the whole line is colored it will be hard to read.

core/logging.py
phorward: fixed in 4ff488d

phorward:
Hi I just stumbled across this PR. Is it possible to have it configurable if the whole line or only the level should be colored? I'm thinking of terminals with white background (mac default). I think if the whole line is colored it will be hard to read.

_Originally posted by @XeoN-GHMB in https://github.com/viur-framework/viur-core/pull/592#discussion_r1052511856_

Hello, yes we should discuss about this. Any ideas?

ArneGudermann:
I think the problem is when we provied a config switch it will not work. Because when two people with different settings work on the same project it will be override with every commit. 
Maybe we must pass it in per command line (args) ?

phorward:
IMHO, the best and most user-friendly method would be to colorize the entire line in a color that is more harmonizing, depending on the background setting.

I did some research on this topic, but determining the background color doesn't seem to be that easy. [Here](https://9to5answer.com/how-to-determine-a-terminal-39-s-background-color) is a collection of solutions, but they don't work on all systems and/or terminals.

On my system, $COLORFGBG is unset. @XeoN-GHMB  can you please check if `echo $COLORFGBG` does output something in your shell?

XeoN-GHMB:
> I think the problem is when we provied a config switch it will not work. Because when two people with different settings work on the same project it will be override with every commit. Maybe we must pass it in per command line (args) or with an env var like `os.getenv("COLOR_WHOLE_LINE")` ?

I think this is good idea. The CLI could provide some help for users which not know how to configure env vars.
The default should be coloring the whole line.

phorward:
Hello altogether, and happy new year.

I've update this pull request to configure both the colorization mode ("full" for full line, "decent" for only the level) and to change colors for specific levels using environment variables, to fit best into personal preferences.

The formatting can be modified using environment variables as follows:

- **VIUR_LOGGING_COLORIZATION** can be either FULL (colorize full line) or DECENT (colorize debug level only)
- **VIUR_LOGGING_COLOR_DEBUG** set debug level color
- **VIUR_LOGGING_COLOR_INFO** set info level color
- **VIUR_LOGGING_COLOR_WARNING** set warning level color
- **VIUR_LOGGING_COLOR_ERROR** set error level color
- **VIUR_LOGGING_COLOR_CRITICAL** set critical error level color

The colors can be "black", "red", "green", "yellow", "blue", "magenta", "cyan" and "white".

Example configuration using viur-cli
```sh
VIUR_LOGGING_COLOR_WARNING=red VIUR_LOGGING_COLORIZATION=decent pipenv run viur run develop
```

XeoN-GHMB (approved):

sveneberth (approved):

PR #566, opened 2022-11-28
Globally control DateBone localization

phorward:
As requested by #565

PR #595, opened 2022-12-14
`DateBone(localize=True)` becomes default setting

### LINKED ISSUE #565: Config flag `conf["viur.bone.date.localize"]` to globally define localize-flag of DateBone
This is an idea with a useful context.

Currently, localize defaults to False and must explicitly be set on a DateBone.
`creationdate` and `changedate` are never localized, but could be using such a flag.

```python
conf["viur.bone.date.localize"] = True
```
would make all `DateBone(date=True, time=True)` localized, when not explicitly switched True or False on the bone.

phorward:
This is the replacement for #566  to resolve #565.


core/bones/date.py
```diff
@@ -33,21 +32,26 @@ def __init__(
...
+            localize = True
 
-        if self.multiple and (creationMagic or updateMagic):
-            raise ValueError("Cannot be multiple and have a creation/update-magic set!")
+        # Magic is only possible in non-multiple bones and why ever only on readonly bones...
+        if creationMagic or updateMagic:
+            if self.multiple:
+                raise ValueError("Cannot be multiple and have a creation/update-magic set!")
+
+            self.readonly = True  # todo: why???
```
sveneberth, line 51: I think this a quick & dirty implementation.
A good implementation would mark this field as `readonly` on `add` if `creationMagic` is enabled and on `edit` if `updateMagic` is enabled.

core/bones/date.py
```diff
@@ -15,7 +14,7 @@ def __init__(
         *,
         creationMagic: bool = False,
         date: bool = True,
-        localize: bool = False,
+        localize: bool = None,
```
sveneberth, line 17:
```suggestion
        localize: Optional[bool] = None,
```

core/bones/date.py
```diff
@@ -33,21 +32,26 @@ def __init__(
...
         if not (date or time):
-            raise ValueError("Attempt to create an empty datebone! Set date or time to True!")
+            raise ValueError("Attempt to create an empty DateBone! Set date or time to True!")
 
+        # Localize-flag only possible with date and time
         if localize and not (date and time):
             raise ValueError("Localization is only possible with date and time!")
+        # Default localize all DateBones, if not explicitly defined
+        elif localize is None and date and time:
+            localize = True
```
sveneberth, line 44:
```suggestion
        elif localize is None:
            localize = date and time
```
otherwise localize would keep `None`.

sveneberth (changes requested):

core/bones/date.py
phorward: Well, IMHO we generally could get rid of anything "magic" in VIUR.  The behavior is as before, I don't want to change it within this PR. We generally should think about a customizable `on_add`- and `on_edit`-trigger for bones which calls a function providing a value.

core/bones/date.py
sveneberth: The explicit setting `None` indeed makes no sense here. leave it like this

sveneberth (approved):

phorward:
I've accidentally created this pull request on main, should go into develop.

sveneberth:
> I've accidentally created this pull request on main, should go into develop.

Can you please merge the main in the develop before? The Changes from https://github.com/viur-framework/viur-core/pull/597 are here too

sveneberth (approved):

phorward:
> Can you please merge the main in the develop before? The Changes from #597 are here new too.

Hoppla, you're totally right. I've now fixed it, so only the changes from the PR are relevant.

PR #598, opened 2022-12-16
Change usages of new path configs

sveneberth:
* Substitute `projectBasePath` correctly with `viur.instance.project_base_path`
* Cast `Path` to `str` before `.replace()`
* Make use of the new Path and use context managers for I/O


phorward:
@ArneGudermann this pull request is intended to extend your pull request #582. Please add your review.

ArneGudermann (approved):

PR #599, opened 2022-12-20
Export RelationalUpdateLevel with viur.core.bones

phorward:
RelationalUpdateLevel should be in `from viur.core.bones import *`


ArneGudermann (approved):

sveneberth:
Shouldn't that then be included in `__all__` too?

PR #600, opened 2022-12-21
added back kindName for userSkel

ArneGudermann:
This PR added back the kindname for the userskel , because is can not be imported now when the Userskel is only defined in the core.
 The problem is here we set the kindName here:
https://github.com/viur-framework/viur-core/blob/e6146a103599a2286f015332860b319968c5aac9/core/skeleton.py#L403-L409

but this `not relNewFileName.strip(os.path.sep).startswith("viur") ` is False. Because of this we must pre set the kindName.

### DIFF
--- a/core/modules/user.py
+++ b/core/modules/user.py
@@ -18,7 +18,7 @@ from typing import Optional
 
 
 class UserSkel(skeleton.Skeleton):
-
+    kindName = "user"
     # Properties required by google and custom auth
     name = EmailBone(
         descr="E-Mail",


phorward:
Does this happen only when the skeleton is not inherited in the project, and directly used from core?
IMHO we should remove the test in line 404 (`and not relNewFileName.strip(os.path.sep).startswith("viur") `) - why should skeletons defined inside the viur-core not infer their name from the class name?

ArneGudermann:
>Does this happen only when the skeleton is not inherited in the project, and directly used from core?

Yes.
You are right we can also remove the test in line 404.

phorward:
> Yes. You are right we can also remove the test in line 404.

Can you please update and test your PR?

ArneGudermann:
Ahh ok the problem is here.
https://github.com/viur-framework/viur-core/blob/e6146a103599a2286f015332860b319968c5aac9/core/skeleton.py#L393
and
https://github.com/viur-framework/viur-core/blob/e6146a103599a2286f015332860b319968c5aac9/core/skeleton.py#L412-L413

we have in this case a ValueError like so :
`ValueError: Duplicate definition for skeleton in /viur/core/prototypes/tree.py and /viur/core/skeleton.py`
because  TreeSkeleton and Skeleton the same kindName

ArneGudermann:
Does anyone have a better solution to the problem?

phorward (approved):
Ok let's turn this back for this case. Anyway, I would like to find a solution later.

PR #605, opened 2023-01-06
Fix/tasks

ArneGudermann:
This PR fix the problem with the Taskqueue. We now query the region not the zone for `queue_region`.


phorward:
@ArneGudermann Thanks for starting this pull request. Can you please re-create the same PR based on the current main branch? As this is a hotfix rather than a feature, I would like to have this in viur-core 3.3.x.

PR #611, opened 2023-01-11
Substitute `BaseApplication` by `Module` and `SkelModule` prototypes

phorward:
This is gonna become a huge pull request with significant, but necessary changes relating the pre-defined prototypes.

Here's a summarization, with explanations.

- `BasicApplication` renamed into `Module` and `SkelModule`: This underlines the default trait of a module: It is a `Module`, not a `BasicApplication`. Especially the term "Application" was already removed several years ago.
- `Module` becomes the bare class to inherit from to identify a module
- `SkelModule` becomes the class to inherit from to identify a module that uses the skeleton system
- Moving role of `adminInfo` into something different: Well, `adminInfo` previously was only a dict, a callable or None. Now, `adminInfo` is only an extension for `describe()`. `describe()` is the new way to obtain a description of the module. It is always a callable.
- The `handler`-property, which previously was the only mandatory adminInfo property, now becomes part of the module, and is no longer relevant to the adminInfo itself. Being backward-compatible, this pull request runs any current ViUR project without huger changes. Just deprecation warnings are thrown.
- This pull request entirely removes the `admin` render, and replaces it (for now) by the `vi` render. This step was done, because #607 was lately merged, but an exact copy of the same function still existed in `admin`. As both renders are duplicates on their own, redundancy was now removed.
- The `vi` render was updated to current VIUR standards
- The `xml` render is still neglected, but was updated to current VIUR standards.
- Docstrings where updated, and old stuff was removed.
- `accessRights` are now a tuple rather than a list, as they shouldn't be mutable.

Ok thats it for now. I've tested this PR with a current Mausbrand project and it runs well so far.


core/prototypes/module.py
```diff
@@ -146,3 +158,55 @@ def baseSkel(self, *args, **kwargs) -> SkeletonInstance:
...
+
+        # Default description
+        ret = {
+            "name": self.__class__.__name__,
+            "handler": ".".join((handler, self.__class__.__name__.lower())),
+        }
+
+        # Extend indexes, if available
+        if indexes := getattr(self, "indexes", None):
+            ret["indexes"] = indexes
```
ArneGudermann, line 185:
```suggestion
        # Extend indexes, if available 
         ret["indexes"] = getattr(self, "indexes", [])
```
I would prefer to have an empty list here so that you always have a field to access in the frontend.

core/render/vi/__init__.py
```diff
@@ -77,56 +67,45 @@ def getStructure(adminTree, module):
...
+@exposed
 def getVersion(*args, **kwargs):
-    # We force the patch-level of our version to be always zero for security reasons
-    return json.dumps((conf["viur.version"][0], conf["viur.version"][1], 0))
-
+    """
+    Returns viur-core version number
+    """
+    if conf["viur.instance.is_dev_server"]:
+        json.dumps(conf["viur.version"])
```
ArneGudermann, line 105:
```suggestion
    if conf["viur.instance.is_dev_server"]:
        return json.dumps(conf["viur.version"])
```
We should return it ?

core/prototypes/module.py
phorward: Hello @ArneGudermann, I understand your concerns. On the other hand, any client relying on this structure should handle the case when something isn't there. Also, I'm of the opinion to keep unused information low, especially when its sent via network. I know, this is only an empty list, but it's an information which contains the same statement if it's not there.

ArneGudermann: Okay, I understand that you just have to pay attention to it then I am so far satisfied with it.

ArneGudermann (approved):

core/render/xml/__init__.py
```diff
@@ -1,47 +1,34 @@
...
     d = datetime.datetime.now()
     return serializeXML(d.strftime("%Y-%m-%dT%H-%M-%S"))
 
 
-timestamp.exposed = True
-
-
-def generateAdminConfig(adminTree):
+@exposed
+def dumpConfig(adminTree):
```
sveneberth, line 21:
```suggestion
def dumpConfig():
```

core/render/xml/__init__.py
```diff
@@ -1,47 +1,34 @@
...
     res = {}
-    for key in dir(adminTree):
+    for key in dir(conf["viur.mainApp"].xml):
```
sveneberth, line 23: This should be the same as in the vi renderer:
```py
    for key in dir(conf["viur.mainApp"].xml):
        module = getattr(conf["viur.mainApp"].xml, key)
        if not isinstance(module, Module):
            continue
        if admin_info := module.describe():
            res[key] = admin_info

    res = {
        "modules": res,
        "configuration": {k.removeprefix("admin."): v for k, v in conf.items() if
                          k.lower().startswith("admin.")}
    }
    return res
```

core/render/vi/__init__.py
```diff
@@ -77,56 +67,45 @@ def getStructure(adminTree, module):
     return json.dumps(res or None, cls=CustomJsonEncoder)
 
 
+@exposed
 def setLanguage(lang, skey):
     if not securitykey.validate(skey):
         return
```
sveneberth, line 73: Btw: Why does this not raise an exception (`PreconditionFailed`) as everywhere else?

core/render/vi/__init__.py
```diff
@@ -77,56 +67,45 @@ def getStructure(adminTree, module):
...
 
 
+@exposed
 def getVersion(*args, **kwargs):
-    # We force the patch-level of our version to be always zero for security reasons
-    return json.dumps((conf["viur.version"][0], conf["viur.version"][1], 0))
-
+    """
+    Returns viur-core version number
+    """
```
sveneberth, line 103:
```suggestion
    """
    currentRequest.get().response.headers["Content-Type"] = "application/json"
```

core/render/vi/__init__.py
```diff
@@ -17,36 +16,27 @@ class default(DefaultRender):
 __all__ = [default]
 
 
+@exposed
 def genSkey(*args, **kwargs):
     return json.dumps(securitykey.create())
```
sveneberth, line 21:
```suggestion
    currentRequest.get().response.headers["Content-Type"] = "application/json"
    return json.dumps(securitykey.create())
```

core/render/vi/__init__.py
```diff
@@ -17,36 +16,27 @@ class default(DefaultRender):
...
 
 
-genSkey.exposed = True
-
-
+@exposed
 def timestamp(*args, **kwargs):
     d = datetime.datetime.now()
     return json.dumps(d.strftime("%Y-%m-%dT%H-%M-%S"))
```
sveneberth, line 27:
```suggestion
    currentRequest.get().response.headers["Content-Type"] = "application/json"
    return json.dumps(d.strftime("%Y-%m-%dT%H-%M-%S"))
```

sveneberth (changes requested):

core/render/vi/__init__.py
phorward: tbd

core/render/xml/__init__.py
phorward: tbd

phorward:
Hello folks, I've updated the branch to another feature requested by @XeoN-GHMB.

It provides the `Module`-class as a base for any ViUR module without any database bindings. The `SkelModule`-class is settled on top of  `Module` and introduces skeleton retrieval functionality. `List`, `Singleton` and `Tree` finally set up on `SkelModule`. 

Can you please re-review the entire PR? I will just update some issues reported by @sveneberth though.

core/modules/formmailer.py
```diff
@@ -1,14 +1,14 @@
 from viur.core.skeleton import RelSkel
 from viur.core import errors, utils, securitykey, exposed, email
 from viur.core.bones import BaseBone
-from viur.core.prototypes.basic import BasicApplication
+from viur.core.prototypes.basic import Module
```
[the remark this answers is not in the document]
phorward: I've fixed it, missed your proposal.

core/render/xml/__init__.py
phorward: fixed.

core/render/vi/__init__.py
phorward: fixed

sveneberth (changes requested):
1. I think `Module` doesn't need a handler nor an adminInfo if it`s just about a routable object. I see these only from the `SkelModule`. Can you move them?

2. I do not see the need for the `base` package. What should be the difference to `prototypes`? Besides, this can quickly lead to confusion with the `BaseBone` or the [base project](https://github.com/viur-framework/viur-base). Please leave that in the `prototypes`.

phorward:
> 1. I think `Module` doesn't need a `handler` nor an `adminInfo`, if it's just an routable object. I see these only from the `SkelModule`. Can you move them?

Well, for me, the way on how to express the module's description is a fundamental part of a VIUR-module, and doesn't have to deal with the Skeleton system. It should work equally for any kind of module, even ones without a database.

>     2. I do not see the need for the `base` package. What should be the difference to `prototypes`? Besides, this can quickly lead to confusion with the `BaseBone` or the [base project](https://github.com/viur-framework/viur-base). Please leave that in the `prototypes`.

This was done to still allow for `from viur.core.prototypes import List` rather than explicit `from viur.core.prototypes.list import List`, because the `__init__.py` should not import anything. The first idea was to put it into `viur.core.module` as a module.py, which is also an option, but then there will also be a skelmodule.py. This solution is much better and a compromise between several ways, to still keep 100% backward compatibility with existing code.

phorward:
In [2ad74d2](https://github.com/viur-framework/viur-core/pull/611/commits/2ad74d27f245c1ad69dd77860e1fac1a5d505bdb) I've renamed the stuff as discussed in the viur-meeting.
Can we now merge, please? @sveneberth

ArneGudermann (approved):

Issue #354, opened 2022-01-04
Render error messages on /json/... as JSON?


phorward:
As HTML/Jinja is the default output render, it is okay that it shows error messages in an HTML-style error message:
![image](https://user-images.githubusercontent.com/16870072/148137547-b7a08fb0-b70f-4ce9-a29c-0c94bbf22fed.png)

But why does a /json/... or /vi/...-call also return the same, HTML-style error message and not some piece of JSON, especially regarding error descriptions which are being lost there.
```json
{
    "code": 404,
    "message": "Not Found",
    "description": "The requested resource could not be found."
}
```
Such a machine-reasable JSON data would be very useful to report back reasons why an action failed back to the user.

Additionally,  this ViUR2-patch would also be reasonable by sending the error message via the response header, and allow Ajax-requests to also deal with error messages in case the returned data should explicitly be HTML.
```patch
--- a/__init__.py
+++ b/__init__.py
@@ -449,6 +449,11 @@ class BrowseHandler(webapp.RequestHandler):
                                raise
                        self.response.clear()
                        self.response.set_status(e.status, e.descr)
+
+                       # Set machine-readable x-viur-error response header in case there is an exception description.
+                       if e.descr:
+                               self.response.headers["x-viur-error"] = e.descr.encode("utf-8") if isinstance(e.descr, unicode) else str(e.descr)
+
                        res = None
                        if conf["viur.errorHandler"]:
                                try:
```

phorward:
> First attempt for a solution:
> 
> https://github.com/viur-framework/viur-core/blob/03e4faff0ed643030d5065485d8a78028798bb91/core/request.py#L327-L331

Hello @ArneGudermann, thanks for your proposal.

Yes, this can work but the URL check on "/json/" is insufficient. I think the routing should generally be done via the render by some kind, but this requires a general rework of the request routing in ViUR.

phorward:
#690 relates as well

phorward:
Fixed by #614

PR #614, opened 2023-01-12
Add Json Error Messages

ArneGudermann:
Rel:#354

With this PR finally an error that is rendered in `json` or `vi` renderer will be output as json too.
Also if you set `utils.currentRequest.get().response.headers["Content-Type"] = "application/json"` and then you throw an error it also will be json.


core/request.py
```diff
@@ -298,16 +299,23 @@ def processRequest(self) -> None:
...
-                })
-                extendCsp({"style-src": ['sha256-Lwf7c88gJwuw6L6p6ILPSs/+Ui7zCk8VaIvp8wLhQ4A=']})
+                if (len(self.pathlist) > 0 and any(x in self.pathlist[0] for x in ["vi", "json"])) or \
+                        utils.currentRequest.get().response.headers["Content-Type"] == "application/json":
+                    utils.currentRequest.get().response.headers["Content-Type"] = "application/json"
+                    res = json.dumps({
+                        "error_code": e.status,
+                        "error_name": str(translate(e.name)),
+                        "error_descr": e.descr,
+                    })
```
phorward, line 309: Hi! I don't like this redundant `error_`-prefix, as a returned HTTP-error is by definition an error.
I'm also of the opinion as @sveneberth stated in #549, that we shouldn't translate the name in this case. 

What about this result, with both the translated and original name, and an improved set of keys?
```suggestion
                    res = json.dumps({
                        "status": e.status,
                        "reason": e.name,
                        "descr": str(translate(e.name)),
                        "hint": e.descr
                    })
```

- **status** is the HTTP status code.
- **reason** is the unmodified HTTP reason phrase.
- **descr** is the possibly translated reason phrase.
- **hint** is a hint about what went wrong. It can be null in some cases.

phorward (changes requested):
Can you please check my comment, or we discuss it at a viur-meeting.

phorward:
Should this PR generally remove the header key **x-viur-error**, as this is the much better solution.

ArneGudermann:
@phorward  Like so ? https://github.com/viur-framework/viur-core/pull/614/commits/312d566e3103f37e7b10f6509848ad1a3cfe8fa3

sveneberth:
> Should this PR generally remove the header key **x-viur-error**, as this is the much better solution?

I would keep it so you can always rely on it.
In case an exception is raised before setting the content-type header, you would get a HTML page.

ArneGudermann:
> > Should this PR generally remove the header key **x-viur-error**, as this is the much better solution?
> 
> I would keep it so you can always rely on it. In case an exception is raised before setting the content-type header, you would get a HTML page.

I don't think so when you call a `/json `or `/vi ` you get JSON.But you are right on a normal route you get html. What do you thing @phorward

sveneberth:
> > > Should this PR generally remove the header key **x-viur-error**, as this is the much better solution?
> > 
> > 
> > I would keep it so you can always rely on it. In case an exception is raised before setting the content-type header, you would get a HTML page.
> 
> I don't think so when you call a `/json `or `/vi ` you get JSON.But you are right on a normal route you get html. What do you thing @phorward

But when I write custom API endpoints, I always want the error reason to be accessible in the same way. Without worrying about the actual content type. No matter if I render JSON, HTML, JSON, XML or plain-text.

ArneGudermann:
@sveneberth Now ok ?

core/request.py
```diff
@@ -288,7 +289,6 @@ def processRequest(self) -> None:
             # Set machine-readable x-viur-error response header in case there is an exception description.
             if e.descr:
                 self.response.headers["x-viur-error"] = e.descr.replace("\n", "")
```
phorward, line 291: Re-adding removed newline
```suggestion
                self.response.headers["x-viur-error"] = e.descr.replace("\n", "")

```

phorward (approved):

sveneberth (approved):

core/request.py
```diff
@@ -298,16 +299,24 @@ def processRequest(self) -> None:
...
-                    tpl = Template(tpl_file.read())
-                res = tpl.safe_substitute({
-                    "error_code": e.status,
-                    "error_name": translate(e.name),
-                    "error_descr": e.descr,
-                })
-                extendCsp({"style-src": ['sha256-Lwf7c88gJwuw6L6p6ILPSs/+Ui7zCk8VaIvp8wLhQ4A=']})
+                if (len(self.pathlist) > 0 and any(x in self.pathlist[0] for x in ["vi", "json"])) or \
+                        utils.currentRequest.get().response.headers["Content-Type"] == "application/json":
+                    utils.currentRequest.get().response.headers["Content-Type"] = "application/json"
```
phorward, line 304: This line can be removed, as the content-type is already "application/json"
```suggestion
```

core/request.py
```diff
@@ -298,16 +295,24 @@ def processRequest(self) -> None:
...
+                    res = json.dumps({
```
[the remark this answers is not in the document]
phorward: would be ok for me. @ArneGudermann ?

phorward (changes requested):
sorry I need to request changes again.

core/request.py
sveneberth: But not necessarily in the case `len(self.pathlist) > 0 and any(x in self.pathlist[0] for x in ["vi", "json"]))` Or?

phorward: You're right!

core/request.py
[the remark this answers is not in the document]
ArneGudermann: okay I add it

core/request.py
```diff
@@ -298,16 +299,31 @@ def processRequest(self) -> None:
...
+                        "status": e.status,
+                        "reason": e.name,
+                        "descr": str(translate(e.name)),
+                        "hint": e.descr,
+                    }
+
+                    if conf["viur.instance.is_dev_server"]:
+                        strIO = StringIO()
+                        traceback.print_exc(file=strIO)
+                        res["traceback"] = strIO.getvalue()
```
sveneberth, line 315:
```suggestion
                        res["traceback"] = traceback.format_exc()
```

sveneberth (approved):

sveneberth (changes requested):
See https://github.com/viur-framework/viur-core/pull/614#discussion_r1080710016

phorward (approved):

ArneGudermann:
Can we merge it or are there more things to discuss?

sveneberth (approved):

PR #615, opened 2023-01-13
Update after (almost) every login the lastlogin timestamp of a user

sveneberth:
Currently the lastlogin timestamp is only updated on Google login. I moved the logic to `authenticateUser()` where it works for every provider.

### DIFF
--- a/core/modules/user.py
+++ b/core/modules/user.py
@@ -3,18 +3,19 @@ import hashlib
 import hmac
 import json
 import logging
-import os
 import time
 import warnings
-from viur.core.prototypes.list import List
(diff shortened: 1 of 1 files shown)


Xeon2003 (approved):

ArneGudermann (approved):

phorward (changes requested):
Hello @sveneberth, I'm not happy with this solution, especially because code is no consequent. I would like to propose a general refactoring of relevant parts. I've just pushed #620 as a replacement for this, which bases on your code. Are you fine with it?

phorward:
Replaced by #620

PR #616, opened 2023-01-16
feat: Count Children in tree module

ArneGudermann:
With this PR we now have the possibility to count the children in a node. This is important in so far as you can better display the tree module on the frontend you now have the option to display the popup elements differently if a node has no children.
Like so:
image

### DIFF
--- a/core/prototypes/tree.py
+++ b/core/prototypes/tree.py
@@ -2,7 +2,7 @@ import logging
 from typing import Any, Dict, List, Literal, Optional, Type, Union
 from viur.core import utils, errors, conf, securitykey, db
 from viur.core import forcePost, forceSSL, exposed, internalExposed
-from viur.core.bones import KeyBone, SortIndexBone
+from viur.core.bones import KeyBone, SortIndexBone, NumericBone
 from viur.core.cache import flushCache
 from viur.core.prototypes import BasicApplication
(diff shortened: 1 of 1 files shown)


phorward (changes requested):
Hello @ArneGudermann,

in 2018 I made a similar approach in https://github.com/viur-framework/server/pull/28, respectively commit https://github.com/viur-framework/server/pull/28/commits/423e2a1474a230502e78a4bacbfe08c7ad556ccd.
The reason why it was rejected was, that a querying or updating of some node was done.

I think this PR doesn't solve it better, because the parent node must keep track of its children, and when a child is modified without using the Tree-module's prototype functions, the statistics are invalid. Believe me, I've did much with statistics and counting on a multi-kind-system, it always was a bad idea.

Why not use the new [count-feature from viur-datastore](https://github.com/viur-framework/server/pull/28/commits/423e2a1474a230502e78a4bacbfe08c7ad556ccd) you recently added there?

sveneberth:
> Why not use the new [count-feature from viur-datastore](https://github.com/viur-framework/viur-datastore/pull/23) you recently added there?


I also think that count is more appropriate here. Just keep in mind that this takes longer with larger result sets (at least that was the case in the python2 environment) and should therefore run best deferred.

PR #617, opened 2023-01-17
Implement a refresh method in the `NumericBone`

sveneberth:
This refresh method ensures numeric values in the datastore.

This might be helpful if you change a StringBone to a NumericBone. You can do a rebuild search index and the strings will be converted to numbers.
This behavior was done in ViUR2 on every unserialize: https://github.com/viur-framework/server/blob/e53c744834103a467a5f51b6b29db7aa531a6096/bones/numericBone.py#L83-L93

---

There's one logical change, because I think this is the most transparent way:
Before, `isEmpty()` determined `"42.0"` with precision as empty, because the `int("42.0")` cast raised a `ValueError`. Now we convert it first to float, then to int.
In case of `precision > 0` this value was already accepted.

---

I'm unsure about the behavior what `refresh` should do on _invalid_ data (a list, embedded entity, ...)

1. Ignore the failed conversion and keep the data
2. Ignore the failed conversion and clear the data (set to `None`)
3. Reraise the exception (current implementation), so the caller has to decide what to do.

What do you think?

### DIFF
--- a/core/bones/numeric.py
+++ b/core/bones/numeric.py
@@ -75,13 +75,8 @@ class NumericBone(BaseBone):
         if isinstance(rawValue, str) and not rawValue:
             return True
         try:
-            if self.precision:
-                if isinstance(rawValue, str):
-                    rawValue = rawValue.replace(",", ".", 1)
-                rawValue = float(rawValue)
-            else:
-                rawValue = int(rawValue)
-        except:
+            rawValue = self._convert_to_numeric(rawValue)
+        except (ValueError, TypeError):
             return True
         return rawValue == self.getEmptyValue()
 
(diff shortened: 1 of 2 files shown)


ArneGudermann:
I am not sure if a list is an invalid record. For example, what if we previously had a `stringbone` that was multiple. Shouldn't we try to translate everything in the `refresh` function?

sveneberth:
> I am not sure if a list is an invalid record. For example, what if we previously had a `stringbone` that was multiple. Shouldn't we try to translate everything in the `refresh` function?

Good point. I hadn't thought of the multiple and language cases. 
If you now turn a `StringBone` into a `NumericBone` with the same language+multiple configuration, numeric values should be converted.

With other formats this fails. Here you have to migrate manually.

sveneberth:
> Hello @sveneberth, I would like to integrate some function similar to JavaScript's parseInt() or parseFloat() into VIUR (let's call these `parse_int()` and `parse_float()`). [Logics already contains such a function](https://github.com/viur-framework/logics/blob/main/logics-py/logics/value.py), as IMHO this is required in other places as well. Whats your opinion on this?

I think these parse functions can be helpful. Especially to process commas.

But I didn't know that letters after a number are simply ignored in JS:exploding_head:. I have also never needed and see no need there. Input fields, even those that have an associated unit, are also in the browser always purely numeric. Would there raise an error with other characters (or return a default value).

phorward (approved):

PR #619, opened 2023-01-17
feat: Improve `PasswordBone` parametrization

ArneGudermann:
Hello everyone,
This PR make the password Bone more configurable. You can now set:

- The min length of the password
- The threshold for tests to pass
- What tests the password must passesd


phorward (changes requested):

core/bones/password.py
```diff
@@ -70,19 +62,64 @@ class PasswordBone(StringBone):
...
+
+            :param min_password_length: The minimum length of the password all passwords with a length that is
+                    smaller this will be invalid.
+            :param test_threshold: The minimum number of tests the password must pass.
+            :param contain_uppercase: Specifies whether the password is checked for uppercase letters
+            :param contain_lowercase: Specifies whether the password is checked for lowercase letters
+            :param contain_number: Specifies whether the password is checked for digits
+            :param contain_special_char: Indicates whether the password is checked for special characters
+        """
+        super(PasswordBone, self).__init__()
```
phorward, line 87: super-parameters can be omitted.
```suggestion
        super().__init__()
```

phorward (changes requested):
one little fix?

phorward (dismissed):

sveneberth (changes requested):
Why not a logics expression as written in the developers chat?



```py
password_test = """sum([
    val.lower() != val,
    val.upper() != val,
    any([x in val for x in "0123456789"]),
    any([x not in ("abcdefghijklmnopqrstuvwxyz0123456789") for x in val.lower()]),
]) > 3"""

logics = Logics(password_test)
print(logics.run({"val": the_password))
```


Not tested, because I cat't get logics to work standalone on my machine :/.
But this could be evaluated by the core and vi and customized - if necessary.

phorward:
> Why not a logics expression as written in the developers chat?

I'm a fan of it, but not within this pull request.

ArneGudermann:
I haven't used `logics-py` because I don't think it's adjustable enough or I don't know enough about it. If I want to use the password bone for some kind of **pin** input then I want to set the length to 4 for example and only allow numbers. Or I want that all 4 conditions in the password are fulfilled then I must write this also again differently.

phorward:
> I haven't used `logics-py` because I don't think it's adjustable enough or I don't know enough about it. If I want to use the password bone for some kind of **pin** input then I want to set the length to 4 for example and only allow numbers. Or I want that all 4 conditions in the password are fulfilled then I must write this also again differently.

Sure, you can express all this in Logics.

ArneGudermann:
But for now we cann keep this ?

sveneberth (commented):
tbh, I don't see any advantage in this PR. It takes away the possibility to create your completely own rules. Only the existing ones can be activated or deactivated. 
What if I want to force at least 3 special characters?

Apart from the fact that probably nobody changes these rules anyway...

phorward:
> Apart from the fact that probably nobody changes these rules anyway...

Let's discuss this tomorrow. I'm also not sure, maybe we need some more explanation.

phorward:
Regex irgendwas

core/bones/password.py
```diff
@@ -69,21 +63,58 @@ class PasswordBone(StringBone):
         "core.bones.password.tooWeakMessage",
         defaultText="The entered password is too weak."
     )
+    _password_tests: List[Tuple] = [
```
sveneberth, line 66:
```suggestion
    password_tests: List[Tuple] = [
```

core/bones/password.py
```diff
@@ -69,21 +63,58 @@ class PasswordBone(StringBone):
...
+            :param min_password_length: The minimum length of the password all passwords with a length that is
+                    smaller this will be invalid.
+
+            :param test_threshold: The minimum number of tests the password must pass.
+            :param password_tests: A list of tuple.The tuple  contains the tests and an error message if the test fails
+        """
+        super().__init__()
+        self.min_password_length = min_password_length
+        self.test_threshold = test_threshold
+        self.password_tests = self._password_tests if password_tests is None else password_tests
```
sveneberth, line 98:
```suggestion
        if if password_tests is not None:
            self.password_tests = password_tests
```

core/bones/password.py
```diff
@@ -69,21 +63,58 @@ class PasswordBone(StringBone):
...
+        min_password_length: int = 8,
+        test_threshold: int = 3,
+        password_tests: List[Tuple] = None,
+        **kwargs
+    ):
+        """
+            Initializes a new Password Bone.
+
+            :param min_password_length: The minimum length of the password all passwords with a length that is
+                    smaller this will be invalid.
```
sveneberth, line 90: I'm not a English pro, but i think this sound better.

```suggestion
            :param min_password_length: The minimum length of a password.
                Passwords with a shorter length will be invalid.
```

core/bones/password.py
```diff
@@ -69,21 +63,58 @@ class PasswordBone(StringBone):
...
+
+            :param test_threshold: The minimum number of tests the password must pass.
+            :param password_tests: A list of tuple.The tuple  contains the tests and an error message if the test fails
```
sveneberth, line 93: Low-hanging fruit. Multiple tuple**s**, each has **one** test and text. Additionally it's more a reason/hint than an error message.

```suggestion
            :param password_tests: A list of tuples. The tuple contains the test and a reason for the user if the test fails.
```

core/bones/password.py
```diff
@@ -69,21 +63,58 @@ class PasswordBone(StringBone):
...
-
-        if sum(testResults) < self.passwordTestThreshold:
-            return str(self.tooWeakMessage)
+        tests_errors = []
+        tests_passed = 0
+        for test in self.password_tests:
+            if re.search(test[0], value):
+                tests_passed += 1
+            else:
+                tests_errors.append(str(test[1]))  # we may need to convert a "translate" object
```
sveneberth, line 114: T think tuple unpacking is more readable and pythonic.

```suggestion
        for test, hint in self.password_tests:
            if re.search(test, value):
                tests_passed += 1
            else:
                tests_errors.append(str(hint))  # we may need to convert a "translate" object
```

core/bones/password.py
```diff
@@ -69,21 +63,58 @@ class PasswordBone(StringBone):
...
-        testResults = []
-        for test in self.passwordTests:
-            testResults.append(test(value))
-
-        if sum(testResults) < self.passwordTestThreshold:
-            return str(self.tooWeakMessage)
+        tests_errors = []
+        tests_passed = 0
+        for test in self.password_tests:
+            if re.search(test[0], value):
```
sveneberth, line 111: I would rather use `re.match` than `re.search`. This allows more specific syntax. (e.g. must start with an number and end with a letter).

```suggestion
            if re.match(test[0], value):
```

core/bones/password.py
```diff
@@ -69,21 +63,58 @@ class PasswordBone(StringBone):
...
         defaultText="The entered password is too weak."
     )
+    _password_tests: List[Tuple] = [
+        (r"(?=.*[A-Z])", translate("core.bones.password.no_capital_letters",
+                                   defaultText="The password entered has no capital letters.")),
+        (r"(?=.*[a-z])", translate("core.bones.password.no_lowercase_letters",
+                                   defaultText="The password entered has no lowercase letters.")),
+        (r"(?=.*\d)", translate("core.bones.password.no_digits",
+                                defaultText="The password entered has no digits.")),
+        (r"(?=.*\W)", translate("core.bones.password.no_special_characters",
```
sveneberth, line 73: Related to the `re.search` --> `re.match` change, the REs should be:

```suggestion
        (r"^(?=.*[A-Z].*)$", translate("core.bones.password.no_capital_letters",
                                   defaultText="The password entered has no capital letters.")),
        (r"^(?=.*[a-z].*)$", translate("core.bones.password.no_lowercase_letters",
                                   defaultText="The password entered has no lowercase letters.")),
        (r"^(?=.*\d.*)$", translate("core.bones.password.no_digits",
                                defaultText="The password entered has no digits.")),
        (r"^(?=.*\W.*)$", translate("core.bones.password.no_special_characters",
```

Do we need the non-capturing-groups?

---

Furthermore I think it's more efficient to use a compiled version of the REs (`re.compile()`). We can use them several times, but compile them only once.
But I'm unsure how to store these.
If we store them directly in `self.password_tests`, the renderer must create a new list (`[test.pattern, hint for test, hint in self.passsword_tests`), which isn't efficient too.
Maybe we could internal a second list with the compiled versions.

sveneberth (changes requested):

core/bones/password.py
```diff
@@ -90,13 +90,13 @@ def __init__(
...
-            :param password_tests: A list of tuple.The tuple  contains the tests and an error message if the test fails
+            :param password_tests: A list of tuples. The tuple contains the test and a reason for the user if the test
+                    fails.
         """
         super().__init__()
         self.min_password_length = min_password_length
         self.test_threshold = test_threshold
-        if if password_tests is not None:
-            self.password_tests = password_tests
+        self.password_tests = password_tests
```
sveneberth, line 99: @ArneGudermann Sorry for the `if if` in my suggestion. But now you would apply also `password_tests` if they are `None`.

sveneberth (dismissed):

phorward:
I think we should wait with this until #637 is merged, to resolve conflicts first and have the bone structure rendering on the right position.

sveneberth:
@ArneGudermann Can you resolve the conflict?

ArneGudermann:
@phorward  and @sveneberth  should be ok now.

phorward (changes requested):
Hello @ArneGudermann,

Still, I'm having the following notes to discuss about:

1. What do you think of checking the minimum password length using a regex as well? Please take a look into my suggestions. Indeed, this would allow for shorter passwords when other thresholds are met, maybe a test can be marked to be *mandatory* with an additional item to the tuple?
2. The variables `tooShortMessage` and `tooWeakMessage` are over. They should be removed.
3. The pull requests title doesn't match its content: "provide these to the client" is not part of the current code anymore, as the `structure` function is missing. But this might also be a security issue, because attackers can see the password guidelines and can bypass them in case the provided regex does contain a bug.

core/bones/password.py
```diff
@@ -52,38 +54,56 @@ class PasswordBone(StringBone):
...
+        **kwargs
+    ):
+        """
+            Initializes a new Password Bone.
+
+            :param test_threshold: The minimum number of tests the password must pass.
+            :param password_tests: A list of tuples. The tuple contains the test and a reason for the user if the test
+                    fails.
+        """
+        super().__init__()
```
sveneberth, line 85:
```suggestion
        super().__init__(**kwargs)
```

sveneberth (changes requested):

core/bones/password.py
phorward: yes, this is important!!!

core/bones/password.py
```diff
@@ -124,3 +144,6 @@ def serialize(self, skel: 'SkeletonInstance', name: str, parentIndexed: bool) ->
 
     def unserialize(self, skeletonValues, name):
         return False
+
+    def structure(self) -> dict:
+        return super().structure() | {"password_tests": self.password_tests}
```
phorward, line 149: Can we call it justs `tests`? In all occurrences?

phorward (changes requested):
Please fix the init super call and consider renaming `password_tests` into just `tests`, as we are in the context of the PasswordBone.

phorward (approved):
I'm fine with this!

sveneberth (approved):

sveneberth:
:clap:

PR #620, opened 2023-01-17
Improvements for User module (substitutes #615)

phorward:
This is a replacement for #615 and bases on the branch from @sveneberth.
It uses the baseSkel() of the user module more excessively and cleans-up the session field takeover.

### DIFF
--- a/core/modules/user.py
+++ b/core/modules/user.py
@@ -770,7 +770,7 @@ class User(List):
             # Conserve DB-Writes: Update the user max once in 30 Minutes (why??)
             if not skel["lastlogin"] or ((now - skel["lastlogin"]) > datetime.timedelta(minutes=30)):
                 skel["lastlogin"] = now
-                skel.toDB()
+                skel.toDB(clearUpdateTag=True)
 
         # Update session for user
         session = utils.currentSession.get()
@@ -785,7 +785,7 @@ class User(List):
 
         utils.currentRequest.get().response.headers["Sec-X-ViUR-StaticSKey"] = session.staticSecurityKey
 
-        self.onLogin()
+        self.onLogin(skel)
         return self.render.loginSucceeded(**kwargs)
 
     @exposed
@@ -814,12 +814,17 @@ class User(List):
                        for x, y in self.validAuthenticationMethods]
         return self.render.loginChoices(authMethods)
 
-    def onLogin(self):
-        usr = self.getCurrentUser()
-        logging.info("User logged in: %s" % usr["name"])
+    def onLogin(self, skel: skeleton.SkeletonInstance):
+        """
+        Hook to be called on user login.
+        """
+        logging.info(f"""User {skel["name"]} logged in""")
 
-    def onLogout(self, usr):
-        logging.info("User logged out: %s" % usr["name"])
+    def onLogout(self, skel: skeleton.SkeletonInstance):
+        """
+        Hook to be called on user logout.
+        """
+        logging.info(f"""User {skel["name"]} logged out""")
 
     @exposed
     def edit(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


core/modules/user.py
```diff
@@ -751,28 +750,41 @@ def secondFactorSucceeded(self, secondFactor, userKey):
...
-        utils.currentRequest.get().response.headers["Sec-X-ViUR-StaticSKey"] = currSess.staticSecurityKey
+        skel = self.baseSkel()
+        if not skel.fromDB(key):
+            raise ValueError(f"Unable to authenticate unknown user {key}")
+
+        # Update the lastlogin timestamp (if available!)
+        if "lastlogin" in skel:
+            now = utils.utcNow()
+
+            # Conserve DB-Writes: Update the user max once in 30 Minutes (why??)
```
sveneberth, line 770: In response to the why: the `toDB` triggers updateRelations for `UserBone`s. On every login you would update/write a possible huge amount of entries. 
But we could set `clearUpdateTag=True` in the `toDB` instead. Right?

phorward: > But we could set `clearUpdateTag=True` in the `toDB` instead. Right?

Right!

core/modules/user.py
```diff
@@ -751,28 +750,41 @@ def secondFactorSucceeded(self, secondFactor, userKey):
...
+        session.reset()
+        # and copy them over to the new session
+        session |= take_over
+
+        session["user"] = skel.dbEntity
+        session.markChanged()
+
+        utils.currentRequest.get().response.headers["Sec-X-ViUR-StaticSKey"] = session.staticSecurityKey
+
         self.onLogin()
```
sveneberth, line 788: I stumbled over this already during my PR. Shouldn't we pass the user skel directly to the `onLogin` method to save the `getCurrentUser` and the related Skel-creation/`db.Get` there?

```suggestion
        self.onLogin(skel)
```
(and extend the signature of onLogin).

sveneberth (approved):
LGTM. My two comments are (I think) optional.

core/modules/user.py
phorward: Indeed, this makes sense. Otherwise, you might break existing projects when they don't ask for `*args`, which should be normal when sub-classing the User module.

Anyway, I think currently every minor release should also be carefully introduced, and when it's broken, it will directly break when logging in. It would also make the signatures of `onLogin` and `onLogout` equal.

phorward:
> LGTM. My two comments are (I think) optional.

I've fixed both comments. Anyway, the 30-min limit remains, not sure if this should really be done on every login?

core/modules/user.py
sveneberth: Then you can remove the 30 min threshold, right?

sveneberth:
> > LGTM. My two comments are (I think) optional.
> 
> I've fixed both comments. Anyway, the 30-min limit remains, not sure if this should really be done on every login?

Okay, we typed parallel :)

Let's leave that in for now. I don't think it has to be that precise.

sveneberth (dismissed):

phorward:
@sveneberth [5ce1318](https://github.com/viur-framework/viur-core/pull/620/commits/5ce131861fd44037acab86bd59d17e8b5b599c11) moves the "lastlogin" update code into onLogin. This makes it much easier to customize either the threshold or to entirely remove this, if wanted.

What do you think?

sveneberth:
> @sveneberth [5ce1318](https://github.com/viur-framework/viur-core/pull/620/commits/5ce131861fd44037acab86bd59d17e8b5b599c11) moves the "lastlogin" update code into onLogin. This makes it much easier to customize either the threshold or to entirely remove this, if wanted.
> 
> What do you think?

It's good. This could really be better. For data privacy, you can then simply disable the "tracking".

sveneberth (approved):

PR #621, opened 2023-01-17
Make MetaBaseSkel.generate_bonemap available

phorward:
This renames the encapsulated function fill_bonemap_recursive into MetaBaseSkel.generate_bonemap and makes it available for externals. This is required by a current project, which modifies skeletons at runtime.


ArneGudermann (approved):

sveneberth (approved):
LGTM. But you should keep in mind that a modification is not consistent across instances. It only modifies the bone map in the current instance. You have to make sure that after modifications the method is called in every instance. So best directly after the instance start.

PR #623, opened 2023-01-18
Drafting a LambdaBone

ArneGudermann:
Hello together with this PR I want to introduce the new LambdaBone this bone stores the value of a Python function. The idea came to me together with @phorward . A use is for example to count the children in #616. 
A threshold controls how many minutes the value of the function is stored to avoid unnecessary computation.

### DIFF
--- a/core/bones/__init__.py
+++ b/core/bones/__init__.py
@@ -82,7 +82,6 @@ for __cls_name, __cls in locals().copy().items():
 
             return __init__
 
-
         locals()[__old_cls_name] = type(__old_cls_name, (__cls,), {
             "__init__": __generate_deprecation_constructor(__cls, __cls_name, __old_cls_name)
         })
--- a/core/bones/__init__.py
+++ b/core/bones/__init__.py
@@ -72,7 +72,6 @@ for __cls_name, __cls in locals().copy().items():
 
         __all += [__old_cls_name]
 
-
         # Dynamically create a class providing a deprecation logging message for every lower-case bone name
         def __generate_deprecation_constructor(cls, cls_name, old_cls_name):
             def __init__(self, *args, **kwargs):
--- a/core/bones/__init__.py
+++ b/core/bones/__init__.py
@@ -15,7 +15,7 @@ from .email import EmailBone
 from .file import FileBone
 from .json import JsonBone
 from .key import KeyBone
-from ._lambda import LambdaBone
+from .lambda_ import LambdaBone
 from .numeric import NumericBone
 from .password import PasswordBone
 from .randomslice import RandomSliceBone
--- a/core/bones/__init__.py
+++ b/core/bones/__init__.py
@@ -1,4 +1,6 @@
-from .base import BaseBone, ReadFromClientError, ReadFromClientErrorSeverity, UniqueValue, UniqueLockMethod, MultipleConstraints
+import importlib
+from .base import BaseBone, ReadFromClientError, ReadFromClientErrorSeverity, UniqueValue, UniqueLockMethod, \
+    MultipleConstraints
 from .boolean import BooleanBone
 from .captcha import CaptchaBone
 from .color import ColorBone
@@ -8,6 +10,7 @@ from .email import EmailBone
 from .file import FileBone
 from .json import JsonBone
 from .key import KeyBone
+from ._lambda import LambdaBone
 from .numeric import NumericBone
 from .password import PasswordBone
 from .randomslice import RandomSliceBone
@@ -38,6 +41,7 @@ __all = [
     "KeyBone",
     "MultipleConstraints",
     "NumericBone",
+    "LambdaBone",
     "PasswordBone",
     "RandomSliceBone",
     "RawBone",
@@ -68,6 +72,7 @@ for __cls_name, __cls in locals().copy().items():
 
         __all += [__old_cls_name]
 
+
         # Dynamically create a class providing a deprecation logging message for every lower-case bone name
         def __generate_deprecation_constructor(cls, cls_name, old_cls_name):
             def __init__(self, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


core/bones/__init__.py
```diff
@@ -1,4 +1,6 @@
-from .base import BaseBone, ReadFromClientError, ReadFromClientErrorSeverity, UniqueValue, UniqueLockMethod, MultipleConstraints
+import importlib
+from .base import BaseBone, ReadFromClientError, ReadFromClientErrorSeverity, UniqueValue, UniqueLockMethod, \
+    MultipleConstraints
```
phorward, line 3: OK if this is reformatted, then please this way:

```suggestion
from .base import \
    BaseBone, \
    ReadFromClientError, \
    ReadFromClientErrorSeverity, \
    UniqueValue, \
    UniqueLockMethod, \
    MultipleConstraints
```

core/bones/_lambda.py
```diff
@@ -0,0 +1,83 @@
...
+            skel.accessedValues[name] = res
+
+            return True
+        else:
+            if callable(self.evaluate):
+                res["value"] = self.evaluate_function(skel, name)
+                res["valid_until"] = utils.utcNow() + timedelta(seconds=self.threshold)
+                skel.accessedValues[name] = res
+                return True
+        return False
```
phorward, line 71: Please avoid double code fragments.

Proposal:
```suggestion
        if data := skel.dbEntity.get(name):
            if data["valid_until"] > utils.utcNow():
                # value is still valid
                skel.accessedValues[name] = res["value"]  # save only the value
                return True

        skel.dbEntity[name] = {
            "value": (skel.accessedValues[name] := self._evaluate(skel, name)),
            "valid_until": utils.utcNow() + timedelta(seconds=self.threshold)
        }

        return True
```

core/bones/_lambda.py
```diff
@@ -0,0 +1,83 @@
...
+        assert not languages
+        assert not indexed
+        assert readonly  # Can we trust our User
+        super().__init__(*args, **kwargs)
+        self.evaluate = evaluate
+        self.threshold = threshold
+
+    def serialize(self, skel: 'SkeletonInstance', name: str, parentIndexed: bool) -> bool:
+        if name in skel.accessedValues:
+            skel.dbEntity[name] = skel.accessedValues[name]
```
phorward, line 43: skel.dbEntity is updated when unserialized. IMHO just remove this.
```suggestion
```

core/bones/_lambda.py
```diff
@@ -0,0 +1,83 @@
...
+                        >>>     return "Hello World"
+                        >>> LambdaBone(evaluate=test)
+                :param threshold The time in minutes before the value is unvalid
+
+
+        """
+        assert not multiple
+        assert not languages
+        assert not indexed
+        assert readonly  # Can we trust our User
```
phorward, line 36:
```suggestion
        assert readonly, "Cannot set readonly to LambdaBone"
        assert callable(evaluate), "'evaluate' must be a callable.
```

core/bones/_lambda.py
```diff
@@ -0,0 +1,83 @@
...
+                res["value"] = self.evaluate_function(skel, name)
+                res["valid_until"] = utils.utcNow() + timedelta(seconds=self.threshold)
+                skel.accessedValues[name] = res
+                return True
+        return False
+
+    def evaluate_function(self, skel: 'viur.core.skeleton.SkeletonInstance', name: str):
+        tmp_skel = skel.clone()
+        tmp_skel[name] = None  # we must remove our bone because of recursion
+        return self.evaluate(__skel__=tmp_skel)
```
phorward, line 76: I think here, we should allow for either a function accepting no parameters or a function accepting a parameter named `skel`: 
```suggestion
        if not self._accept_skel_arg:
            return self.evaluate()  # call without any arguments
            
        skel = skel.clone()
        skel[name] = None  # we must remove our bone because of recursion
        return self.evaluate(skel=skel)
```

core/bones/_lambda.py
```diff
@@ -0,0 +1,83 @@
+import logging
```
phorward, line 1:
```suggestion
import logging
import inspect
```

core/bones/_lambda.py
```diff
@@ -0,0 +1,83 @@
...
+                :param threshold The time in minutes before the value is unvalid
+
+
+        """
+        assert not multiple
+        assert not languages
+        assert not indexed
+        assert readonly  # Can we trust our User
+        super().__init__(*args, **kwargs)
+        self.evaluate = evaluate
```
phorward, line 38:
```suggestion
        self.evaluate = evaluate
        self._accept_skel_arg = "skel" in inspect.signature(evaluate).parameters
```

core/bones/_lambda.py
```diff
@@ -0,0 +1,83 @@
...
+        tmp_skel = skel.clone()
+        tmp_skel[name] = None  # we must remove our bone because of recursion
+        return self.evaluate(__skel__=tmp_skel)
+
+    def refresh(self, skel: 'viur.core.skeleton.SkeletonInstance', boneName: str) -> None:
+        """
+            Refresh the output of the self.evaluate function
+        """
+        skel[boneName] = {"value": self.evaluate_function(skel, boneName),
+                          "valid_until": utils.utcNow() + timedelta(seconds=self.threshold)}
```
phorward, line 83:
```suggestion
        skel[boneName] = self.evaluate_function(skel, boneName)
```

phorward (changes requested):
Hello @ArneGudermann,

good work! I think it will be better to store the `valid_until` value only in dbEntity, but never in the SkeletonInstance. The user should use this bone as any normal value, without bothering with the the `valid_until` flag.

My changes are untested. Can you please apply them and check if it works as expected? I would do so as well.

core/bones/__init__.py
```diff
@@ -8,6 +10,7 @@
 from .file import FileBone
 from .json import JsonBone
 from .key import KeyBone
+from ._lambda import LambdaBone
```
phorward, line 13: Should we better name this `lambda_`? The preceding `_` always looks a little bit private...

sveneberth (commented):
Interesting idea!

However, I find the name `LambdaBone` not very meaningful. I could not imagine anything about it.
I would rather use something like `ComputableBone`, `CacheBone`, `DynamicBone`. Let's brainstorm in the ViUR meeting!

phorward:
Result of the viur-meeting: This feature might be useful. Let's start a new attempt as a `evaluate`-parameter on BaseBone.

ArneGudermann:
Is presented here #639

PR #625, opened 2023-01-18
Reset renderPreparation in renderEmail

ArneGudermann:
This PR reset the `renderPreparation` in `renderEmail` because if we use a Bone after we send an email we got an wrong type.
Example:
```python
skel = self.editSkel()
if not skel.fromDB(key):
		raise errors.NotFound()
print(f"""before:{type(skel["key"])}""")

email.sendEMail(dests=["test@mail.com"], tpl="viur_mail_dummy", skel=skel, sender="test@mail.com")

print(f"""after:{type(skel["key"])}""")
```
We got : 
```
before:<class 'viur.datastore.types.Key'>
after:<class 'str'>
```

### DIFF
--- a/core/render/html/default.py
+++ b/core/render/html/default.py
@@ -541,9 +541,11 @@ class Render(object):
         """
         if isinstance(skel, SkeletonInstance):
             skel.renderPreparation = self.renderBoneValue
-        elif isinstance(skel, list) and all([isinstance(x, SkeletonInstance) for x in skel]):
+
+        elif isinstance(skel, list):
             for x in skel:
-                x.renderPreparation = self.renderBoneValue
+                if isinstance(x,SkeletonInstance):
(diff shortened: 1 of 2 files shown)


sveneberth (approved):
Looks okay. But can we please use another, more meaningful variable name than `x`. Maybe `skel_` or `skel_instance`? Or at least a `s`? :smiley:

phorward:
> Looks okay. But can we please use another, more meaningful variable name than `x`. Maybe `skel_` or `skel_instance`? Or at least a `s`? smiley

The entire code here is bulls*it, due to historical reasons. It should be generally reworked. IMHO, the best solution in this case would be to allow setting `renderPreparation` to a SkelList as well, which then runs this loop. But this isn't part of this pull request which is a hotfix.

phorward (approved):

PR #636, opened 2023-01-30
Add JsonBone for the user to store config information.

ArneGudermann:
This PR adds `admin_config` to the UserSkel to store configrations. In the new VI we want to store infos about the favorite modules of a user and now we can store it in a `JsonBone`

### DIFF
--- a/core/modules/user.py
+++ b/core/modules/user.py
@@ -98,6 +98,10 @@ class UserSkel(skeleton.Skeleton):
         defaultValue=0,
     )
 
+    admin_config = JsonBone(
+        descr="Config for the User"
+    )
+
 
 class UserPassword:
     registrationEnabled = False
--- a/core/modules/user.py
+++ b/core/modules/user.py
@@ -98,8 +98,9 @@ class UserSkel(skeleton.Skeleton):
         defaultValue=0,
     )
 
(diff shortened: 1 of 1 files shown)


phorward:
> In the new VI we want to store infos about the favorite modules of a user and now we can store it in a `JsonBone`

Hello @ArneGudermann, can you shortly explain why this can't be solved using a `RecordBone` with a defined structure? IMHO a JsonBone is not a good choice here, as it can save arbitrarily expressed information, which might gets corrupted on bugs or when different versions of the VI are writing broken data.

ArneGudermann:
I chose a JsonBone here because at this point we don't know what all we want to store. For example, a history can still be added. Of course you could always extend the structure but then the VI must always wait for the core.

sveneberth (changes requested):
Would be okay for me, if this is clearly in the responsibility of the client. But please add a comment or (not exclusive) a tooltip to the bone (sth. like: _"This bone stores settings from the vi"_). Furthermore I think this could be `visible=False`, what do you think?

phorward (approved):

sveneberth (approved):

phorward:
We should think about #653 for this as well.

PR #638, opened 2023-01-31
Add `conf["viur.dev_server_cloud_logging"]`

ArneGudermann:
With this PR you can now set if you want to see your local log entries also in google cloud logger. I don't think it's necessary to send all your logs to google, because it costs time and memory unnecessarily. 
It also halves the response time in the local mode for simple requests.

### DIFF
--- a/core/config.py
+++ b/core/config.py
@@ -103,6 +103,9 @@ conf = Conf({
     # Unless overridden by the Project: Use english as default language
     "viur.defaultLanguage": "en",
 
+    # If disabled the local logging will not send with requestLogger to the cloud
+    "viur.dev_server_cloud_logging": True,
+
     # If set to true, the decorator @enableCache from viur.core.cache has no effect
     "viur.disableCache": False,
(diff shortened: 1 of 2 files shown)


XeoN-GHMB (dismissed):

phorward (approved):

core/config.py
```diff
@@ -103,6 +103,9 @@ def __setitem__(self, key, value):
     # Unless overridden by the Project: Use english as default language
     "viur.defaultLanguage": "en",
 
+    # If disabled the local logging will not send with requestLogger to the cloud
+    "viur.dev_server_cloud_logging": True,
```
phorward, line 107:
```suggestion
    "viur.dev_server_cloud_logging": False,
```
shouldn't this be False by default?

phorward (changes requested):

sveneberth (approved):

phorward (approved):

Issue #629, opened 2023-01-24
`compute`-Parameter for bones to allow for computed values


phorward:
> Result of the viur-meeting: This feature might be useful. Let's start a new attempt as a `evaluate`-parameter on BaseBone.

_Originally posted by @phorward in https://github.com/viur-framework/viur-core/issues/623#issuecomment-1398489726_

This issue is a substitution for #623. It is useful to quarantee expected data types and bone rendering behaviors with computed values.

- `BaseBone` with `compute`-parameter as a dataclass
- The dataclass should require for
  - a lambda for value computation (simular to #623)
  - an optional threshold, either with the options
    - compute always
    - compute on add/edit only
    - re-compute when threshold of X seconds has passed
- Bones using the `compute`-flag should at first be limited to
  - `readOnly=True`
  - `multiple=False` - this could be improved later on.
  - `languages=None` - this could be improved later on.

This issue relates to #616 as well.

ArneGudermann:
@viur-framework/developers 
I have a question about the implementation. For example, what do we do if I have the following code. 
```python
def testme():
    return "123.1"

compute_me = NumericBone(
    readOnly=True,
    compute=testme,
    threshold=ThresholdValue(method=ThresholdMethods.Always))
 ```
In this case, should we convert the value we get from `testme` to `int` or should we keep the value as it is?

In general, we should discuss when the `compute` function returns a value that does not match the bone.

**Comment by phorward** [comment]
Hello @ArneGudermann,

I think that this namedtuple + enum definition would generally be the better choice in BaseBone. This draft defines all required features, and might also answer your question: The compute-definition might be configured to allow for raw values or to pass them through the bone's fromClient. 
```python
from enum import Enum, unique

@unique
class ComputeThreshold(Enum):
    ALWAYS = 1
    ONCE = 2
    SECONDS = 3

    def __init__(self, value=None):
        self.value = value

    @classmethod
    def seconds(cls, value):
        instance = cls.SECONDS
        instance._value_ = value
        instance.value = value
        return instance

@dataclass
class Compute:
    fn: callable,  # the callable computing the value
    threshold: ComputeThreshold = ComputeThreshold.ALWAYS   # the value caching threshold
    raw: bool = true  # defines whether the value is used as is, or is passed to bone.fromClient
```
Afterwards, `compute_me` can be defined this way:
```python
compute_me = NumericBone(
    compute=Compute(testme)
)

compute_me_2min = NumericBone(
[...]

PR #639, opened 2023-02-01
feat: Add `compute`-feature to `BaseBone`

ArneGudermann:
Extension for #623. With this PR you can now pass a `compute` parameter to each bone.
The `compute` parameter is a `dataclass` and consists of the following attributes.

```python
@dataclass
class Compute:
    fn: callable  # The callable computing the value
    threshold: ThresholdValue   # The value caching interval
    raw: bool = True  # defines whether the value is used as is, or is passed to bone.fromClient
```
The `ThresholdValue` is a dataclass to and defiend like this.

```python
class ThresholdMethods(Enum):
    Always = 0
    Until = 1
    Once = 2


@dataclass
class ThresholdValue:
    method: ThresholdMethods = ThresholdMethods.Always
    seconds: int = 0
```


phorward:
Hello @ArneGudermann, 

thanks for starting this pull request draft.

I think you've used my first proposal from [https://github.com/viur-framework/viur-core/issues/629#issuecomment-1409531246](my comment in #629) which you got by email, but I did several revisions on it. I think, that Threshold is the wrong value for this, and named it (in a later version of the comment) into `ComputeInterval`. Can you please re-check this and merge this proposal into this pull request?

For clarification, here's my comment from the link above as a copy. I've tested it and it should entirely fulfill all requirements on issue #692.

---

Hello @ArneGudermann,

I think that this namedtuple + enum definition would generally be the better choice in BaseBone. This draft defines all required features, and might also answer your question: The compute-definition might be configured to allow for raw values or to pass them through the bone's fromClient. 
```python
from dataclasses import dataclass
from enum import Enum, unique

@unique
class ComputeInterval(Enum):
    ALWAYS = 1
    ONCE = 2
    SECONDS = 3

    @classmethod
    def seconds(cls, value):
        instance = cls.SECONDS
        instance._value_ = value
        return instance

@dataclass
class Compute:
    fn: callable  # the callable computing the value
    interval: ComputeInterval = ComputeInterval.ALWAYS   # the value caching interval
    raw: bool = True  # defines whether the value is used as is, or is passed to bone.fromClient
```
Afterwards, `compute_me` can be defined this way:
```python
def testme():
    return "123.1"

compute_me = NumericBone(
    compute=Compute(testme)
)

# Compute(fn=<function testme at 0x7f790476cee0>, interval=<ComputeInterval.ALWAYS: 1>, raw=True)

compute_me_securely = NumericBone(
    compute=Compute(testme, raw=False)
)

# Compute(fn=<function testme at 0x7f790476cee0>, interval=<ComputeInterval.ALWAYS: 1>, raw=False)

compute_me_2min = NumericBone(
    compute=Compute(testme, ComputeInterval.seconds(60 * 2))
)

# Compute(fn=<function testme at 0x7f790476cee0>, interval=<ComputeInterval.SECONDS: 120>, raw=True)
```
The `readOnly` is automatically enforced by the constructor.

What do you think? I want to avoid enforcing to only accept str and parse them always via fromClient, as this might result in values generated, converted to a parse-able representation to become the same representation as originally created in the end, which is an unnecessary round trip.

ArneGudermann:
Hello @phorward
Thanks for your answer. I have one thing.
```python
compute_me_2min = NumericBone(
    compute=Compute(testme, ComputeInterval.seconds(60 * 2))
)
````
Unfortunately this does not work because when I clone the skeleton before passing it to the compute function I get 120 as the value of the enum.

**Comment by phorward** [comment]
> Unfortunately this does not work because when I clone the skeleton before passing it to the compute function I get 120 as the value of the enum.

You're right and I apologize. I really miss Rust's enums in Python, as
```rust
enum ComputeInterval {
    Always,
    Once,
    Seconds(usize),
}
```
is what I wanted to imitate here.

Anyway, could we make it simply this way first?
```python
from dataclasses import dataclass

@dataclass
class Compute:
    fn: callable  # the callable computing the value
    seconds: int = 0  # interval in seconds, default is always
    raw: bool = True  # defines whether the value is used as is, or is passed to bone.fromClient
```
So `seconds` is just an int, which defines 0 as always, anything < 0 as once and >0 as timeout?
```python
compute_me = NumericBone(
    compute=Compute(testme)
)

#Compute(fn=<function testme at 0x7fe27e97fd90>, seconds=0, raw=True)

compute_me_securely = NumericBone(
    compute=Compute(testme, seconds=-1, raw=False)
)

#Compute(fn=<function testme at 0x7fe27e97fd90>, seconds=-1, raw=False)

compute_me_2min = NumericBone(
    compute=Compute(testme, seconds=60 * 2)
)

#Compute(fn=<function testme at 0x7fe27e97fd90>, seconds=120, raw=True)
```

core/bones/base.py
```diff
@@ -59,6 +61,25 @@ class MultipleConstraints:  # Used to define constraints on multiple bones
...
+@dataclass
+class ThresholdValue:
+    method: ThresholdMethods = ThresholdMethods.Always
+    seconds: int = 0
+
+
+@dataclass
+class Compute:
+    fn: callable  # the callable computing the value
+    threshold: ThresholdValue   # the value caching interval
```
sveneberth, line 79: How about _expiration_ or _lifetime_?

core/bones/base.py
```diff
@@ -493,13 +528,47 @@ def unserialize(self, skel: 'viur.core.skeleton.SkeletonInstance', name: str) ->
         """
         if name in skel.dbEntity:
             loadVal = skel.dbEntity[name]
+            if self.compute:  # handle this in the first place
+                if self.compute.threshold.method == ThresholdMethods.Until:  # we maybe must compute the value new
+                    # if we have no value we set it to now and compute new
+                    valid_until = skel.dbEntity.get(f"{name}__vaild_until", utils.utcNow())
```
sveneberth, line 534: I would rather store the last updated date than the until date. If you change the bone parametrization and reduce the interval (maybe you accidentally provides minutes, instead of seconds) this affects only the next computation.

core/bones/base.py
```diff
@@ -59,6 +61,25 @@ class MultipleConstraints:  # Used to define constraints on multiple bones
...
+class ThresholdMethods(Enum):
+    Always = 0
+    Until = 1
+    Once = 2
+
+
+@dataclass
+class ThresholdValue:
+    method: ThresholdMethods = ThresholdMethods.Always
+    seconds: int = 0
```
sveneberth, line 73: Can we please use a `timedelta` instance here? It's easier to write `timedelta(days=2)` instead of `172800` or `2 * 24 * 60 * 60`.

core/bones/base.py
```diff
@@ -882,3 +951,21 @@ def iter_bone_value(
...
+            skel = skel.clone()
+            skel[name] = None  # we must remove our bone because of recursion
+            data = self.singleValueUnserialize(self.compute.fn(skel=skel))
+
+        if not self.compute.raw:
+            errors = self.fromClient(skel, name, {name: data})
+            if errors is None:
+                return skel[name]
+            else:
+                logging.error(f"Parse Data failed with {errors} \nWe return the raw data.")
```
sveneberth, line 969: I this line break intended? I'm not sure if the google logging console displays line breaks reasonable.

```suggestion
                logging.error(f"Parse Data failed with {errors}. We return the raw data.")
```

sveneberth (changes requested):

sveneberth (changes requested):

core/bones/base.py
```diff
@@ -438,6 +477,18 @@ def serialize(self, skel: 'SkeletonInstance', name: str, parentIndexed: bool) ->
 
             :param name: The property-name this bone has in its Skeleton (not the description!)
         """
+        logging.error("here")
+        logging.error(self.compute)
+        logging.error(name)
+        if self.compute:  # handle this in the first place
+            if self.compute.threshold.method == ThresholdMethods.OnStore:
+                logging.error("calcnew")
```
sveneberth, line 485:
```suggestion
```

core/bones/base.py
```diff
@@ -914,4 +1018,9 @@ def structure(self) -> dict:
             }
         else:
             ret["multiple"] = self.multiple
+        if self.compute:
+            ret["compute"] = {"method": self.compute.threshold.method.name}
+            if self.compute.threshold.lifetime:
+                ret["compute"]["lifetime"] = str(self.compute.threshold.lifetime)
+                ret["compute"]["last_updated"] = self.compute.threshold.last_updated.strftime("%Y-%m-%dT%H-%M-%S")
```
sveneberth, line 1025: Can we use `.isoformat()` here? 
```suggestion
                ret["compute"]["last_updated"] = self.compute.threshold.last_updated.isoformat()
```

sveneberth (changes requested):
I've added this bone into my skeleton:
```py

	computed = StringBone(
		descr=u"Name",
		compute=Compute(
			lambda: time.time(),
			ThresholdValue(ThresholdMethods.Lifetime, lifetime=td(minutes=2)),
			raw=False,
		)
	)
```
But if I load this skeleton (`.../view/...`), I get every time a new value (because the calculated value won't be changed).
And if I save the skeleton in the vi, the value won't be saved too.

sveneberth (changes requested):
Hey @ArneGudermann, 
the calculated value is now also written when saving via vi (edit). Thanks for this fix.

However, the value is still recalculated for each view/list.
IMO, as soon as the value is recalculated in `unserialize` (by the compute function), it should also be written to the database. In general, you have more read than write OPS. For example, if you look at a folder and have a compute bone for the amount of children, it would recalculate the value on every view. Only when I rename the folder would the value be cached. In my opinion, this is inefficient and does not make sense for a value cached by lifetime.

phorward (changes requested):
Hello @ArneGudermann, please ignore my suggestions, I will make and test a separate PR on this, as the feature is important for viur-core 3.4

Generally, I would suggest to
- NOT use the term "threshold", as this isn't a threshold we're going to configure
- Fail with ValueError in case something is misconfigured. This is good practise for most cases in VIUR.

Please be patient, I'll take a look on this now, including renamings.

phorward:
@sveneberth @ArneGudermann I've made a reworked proposal based on Arne's draft. It works, but I think some things could be improved as well. Can you please make a review on this?

core/bones/base.py
```diff
@@ -74,7 +97,7 @@ def __init__(
         languages: Union[None, List[str]] = None,
         multiple: Union[bool, MultipleConstraints] = False,
         params: Dict = None,
-        readOnly: bool = False,
+        readOnly: bool = None,  # fixme: Rename into readonly (all lowercase!) soon.
```
phorward, line 100: This is set to `None` to distinguished an explicitly specified or unspecified value, see below.

core/bones/base.py
```diff
@@ -883,6 +990,26 @@ def iter_bone_value(
...
+        """Performs the evaluation of a bone configured as compute"""
+
+        if "skel" not in inspect.signature(self.compute.fn).parameters:
+            # call without any arguments
+            ret = self.compute.fn()
+        else:
+            # otherwise, call with a clone of the skeleton
+            cloned_skel = skel.clone()
+            cloned_skel[name] = None  # remove bone to avoid endless recursion
+            ret = self.compute.fn(skel=cloned_skel)
```
phorward, line 1003: I'm not happy with this, but there's no better solution IMHO, talked to @ArneGudermann previously on it.

core/bones/base.py
```diff
@@ -914,4 +1041,12 @@ def structure(self) -> dict:
             }
         else:
             ret["multiple"] = self.multiple
+        if self.compute:
+            ret["compute"] = {
+                "method": self.compute.interval.method.name
+            }
+
+            if self.compute.interval.lifetime:
+                ret["compute"]["lifetime"] = str(self.compute.interval.lifetime)
```
phorward, line 1050: Shall we generally return seconds here? Currently, it's "0:00:30" for `timedelta(seconds=30)`.

core/bones/base.py
```diff
@@ -438,6 +482,30 @@ def serialize(self, skel: 'SkeletonInstance', name: str, parentIndexed: bool) ->
...
+                    if not last_update or last_update + self.compute.interval.lifetime < now:
+                        skel.accessedValues[name] = self._compute(skel, name)
+                        skel.dbEntity[f"_viur_compute_{name}_"] = now
+
+                case ComputeMethod.Once:
+                    if name not in skel.dbEntity:
+                        skel.accessedValues[name] = self._compute(skel, name)
+
+            # logging.debug(f"WRITE {name=} {skel.accessedValues=}")
+            # logging.debug(f"WRITE {name=} {skel.dbEntity=}")
```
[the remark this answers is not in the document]
phorward: It might me useful and doesn't eat any bread.

core/bones/base.py
```diff
@@ -59,13 +61,34 @@ class MultipleConstraints:  # Used to define constraints on multiple bones
     preventDuplicates: bool = False  # Prevent the same value of being used twice
 
 
+class ComputeMethod(Enum):
+    Always = 0  # Always compute on deserialization
+    Lifetime = 1  # Update only when given lifetime is outrun
```
phorward, line 66:
```suggestion
    Lifetime = 1  # Update only when given lifetime is outrun; value is only being stored when the skeleton is written
```

phorward (approved):

core/bones/base.py
sveneberth: @ArneGudermann Why do we need to store this?

sveneberth (changes requested):
Can we resolve https://github.com/viur-framework/viur-core/pull/639/files#r1166918242 before merging?

For the caching thing I've create an issue https://github.com/viur-framework/viur-core/issues/709

phorward:
> Can we resolve https://github.com/viur-framework/viur-core/pull/639/files#r1166918242 before merging?

This refers to #639, can you clarify?

> For the caching thing I've create an issue #709

Thanks! :+1:

ArneGudermann (commented):
I fine with it. I think we should merge it. It not break anything and we can test it and improve it.

sveneberth:
> > Can we resolve https://github.com/viur-framework/viur-core/pull/639/files#r1166918242 before merging?
> 
> This refers to #639, can you clarify?
> 
> > For the caching thing I've create an issue #709
> 
> Thanks! +1

The link was stupid... This was the issue, that has been resolved: https://github.com/viur-framework/viur-core/pull/639#discussion_r1166918242

sveneberth (dismissed):

phorward (approved):
OK let's do this.

core/bones/base.py
```diff
@@ -883,6 +990,26 @@ def iter_bone_value(
...
+            # call without any arguments
+            ret = self.compute.fn()
+        else:
+            # otherwise, call with a clone of the skeleton
+            cloned_skel = skel.clone()
+            cloned_skel[name] = None  # remove bone to avoid endless recursion
+            ret = self.compute.fn(skel=cloned_skel)
+
+        if self.compute.raw:
+            return self.singleValueUnserialize(ret)
```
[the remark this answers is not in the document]
phorward: @ArneGudermann ?

PR #637, opened 2023-01-30
Move structure dict rendering from the renders into the bones

ArneGudermann:
This PR move the `renderBoneStructure` form the renderer to the  bones to eliminate multiple code. It also move `renderSkelStructure` in the `skeleton`. 
At this stage we still need to include the `render_type` to distinguish which renderer we are in.

PR #642, opened 2023-02-03
feat: Add tzNaive flag to dateBone

AchimSchumacher:
This is in order to ensure a `DateBone` is rendered independently of any local time zone.
As in viur3 all DateBones are timezone aware, per default set to UTC. When rendered into JSON, values are Strings in isoformat, i.e. `"YYYY-MM-DDTHH:MM:SS+HH:MM"`. In UTC, this is always `"YYYY-MM-DDTHH:MM:SS+00:00"`.
In JS, e.g. vuejs, in order to obtain a String representation in local format, `new Date(value).toLocaleString(window.navigator.language)` not only changes language but also converts into local time.
This may be undesired, e.g. a workdate confirmed with a client should not change time the moment summertime starts. When talking to the client, both parties should see the same time, even if living in different timezones.

The implementation chosen only adds the flag to `DateBone` w/o any further changes. In the default JSON renderer, the flag is then read and results simply in the `"+00:00"` part being removed from the String. This is enough for local Date objects initialized from the String to be timezone naive. If needed, other renderers should be updated, too.

Defining `tzNaive=True` in the `Skeleton` has the advantage of defining this behavior globally for all clients rather hacking it into each client individually. 

The alternative approach of using a tz-naive Date-object internally in `DateBone` would implicitly convert to local tz w/o control, e.g. in `datetime.fromtimestamp()`. Furthermore, such a substantial change would be error prone.

### DIFF
--- a/core/bones/date.py
+++ b/core/bones/date.py
@@ -17,6 +17,7 @@ class DateBone(BaseBone):
         date: bool = True,
         localize: bool = False,
         time: bool = True,
+        tzNaive: bool = False,
         updateMagic: bool = False,
         **kwargs
     ):
@@ -28,8 +29,10 @@ class DateBone(BaseBone):
             :param updateMagic: Use the current time whenever this entity is saved.
             :param date: Should this bone contain a date-information?
             :param time: Should this bone contain time information?
+            :param tzNaive: Should this bone be time zone naive? Use this for fixed times which will be rendered
+                                always with the same value independent of local time zone.
             :param localize: Assume users timezone for in and output? Only valid if this bone
-                                contains date and time-information! Per default, UTC time is used.
+                                contains date and time-information and tzNaive is false! Per default, UTC time is used.
         """
         super().__init__(**kwargs)
 
@@ -45,11 +48,15 @@ class DateBone(BaseBone):
         if localize and not (date and time):
             raise ValueError("Localization is only possible with date and time!")
 
+        if localize and tzNaive:
+            raise ValueError("Localization is not possible with tzNaive!")
+
         if self.multiple and (creationMagic or updateMagic):
             raise ValueError("Cannot be multiple and have a creation/update-magic set!")
 
         self.date = date
         self.time = time
+        self.tzNaive = tzNaive
         self.localize = localize
 
     def singleValueFromClient(self, value: str, skel: 'viur.core.skeleton.SkeletonInstance', name: str, origData):
@@ -89,6 +96,9 @@ class DateBone(BaseBone):
             :param value: *User-supplied* request-data, has to be of valid format
             :returns: tuple[datetime or None, [Errors] or None]
         """
+        # Even if tzNaive==true, we need to set tz=UTC,
+        # or else e.g. POSIX timestamp will be converted to the platform’s local date and time.
+        # => Bones with tzNaive==true are internally tz aware; they are just rendered differently.
         if self.date and self.time and self.localize:
             time_zone = self.guessTimeZone()
         else:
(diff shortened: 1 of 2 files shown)


sveneberth (commented):
The finally rendering of the datetime is not the task of the core, it's the part of the frontend.
If you don't need localization use an unlocalized `DateBone` (`DateBone(localize=False)`, which is the default anyway). 
And display in your `Vue.Js` always the Date in UTC:
```js
new Date(date_from_skel).toLocaleString('default', {timeZone: 'UTC'})
```

AchimSchumacher:
> The finally rendering of the datetime is not the task of the core, it's the part of the frontend. 
And yet render/json/default.py is part of core, so I added the change there.

> If you don't need localization use an unlocalized `DateBone` (`DateBone(localize=False)`, which is the default anyway). 
That's not the point. localize=True would tell DateBone to guess the user's time zone, while localize=False uses UTC. In both cases, DateBone uses localized (as in time-zone aware) Dates.

> And display in your `Vue.Js` always the Date in UTC:
> 
> ```js
> new Date(date_from_skel).toLocaleString('default', {timeZone: 'UTC'})
> ```
Yes, one could tell Date.toLocaleString the time zone we are interested in. But then, at the time of rendering, one has to know which DateBone has to be rendered in local time and which one in UTC time.

All the PR does is adding a flag so that at the time of defining the DateBone in a Skeleton, one can set this behavior, i.e. where we can prevent the clients from  transforming the Date into some supposed local time zone.

sveneberth:
> Yes, one could tell Date.toLocaleString the time zone we are interested in. But then, at the time of rendering, one has to know which DateBone has to be rendered in local time and which one in UTC time.

You could add a flag in your bone params if this particular type of rendering is required.


However, the main problem with this implementation is that it is inconsistent across all renderers and the skeleton.
The renderers should output an appropriate representation for the data stored in the skeleton without corrupting the data or truncating data.

AchimSchumacher:
> However, the main problem with this implementation is that it is inconsistent across all renderers and the skeleton. 

It is inconsistent across all renderers as long as it is implemented only in the json renderer. You are correct in that the same behavior should be implemented in all other renderers, too.

> The renderers should output an appropriate representation for the data stored in the skeleton without corrupting the data or truncating data.

This is the point of the PR. Without it, renderers do not output an appropriate representation for the data stored in the skeleton, for the case of a Date which should not underly any changes due to time zone or daylight saving time. As is, these Dates are corrupted by applying a time-shift which does not correspond to the Date. 

The data isn't truncated, neither, although obviously a part of the string is truncated. That truncated part is exactly telling the receiver the Date is timezone-aware, which actually is corrupting the data.

core/bones/date.py
```diff
@@ -17,6 +17,7 @@ def __init__(
         date: bool = True,
         localize: bool = False,
         time: bool = True,
+        tzNaive: bool = False,
```
phorward, line 20: Please change this for PEP8-compliance, into something like
```suggestion
        naive: bool = False,
```

phorward (changes requested):
Hello @AchimSchumacher,
thanks for filing this pull request.

First of all, please
1. Fix the merge conflicts with the current develop branch
2. Don't use `tzNaive` as flag name; use `naive` or another name which does not use camel-case.
3. Your code bypasses current rendering of `datetime` values, which is done in the CustomJsonEncoder. This code will be replaced soon by a more generic value rendering (likewise #637). It is ok for me, but we should discuss it.

phorward:
@sveneberth will provide an alternative attempt.

phorward:
This pull request is replaced by #667 and will be closed now.

PR #654, opened 2023-02-22
Improve Cloud Tasks creation in `CallDeferred` and `QueryIter`

### LINKED ISSUE #641: Cloud tasks should run on the version where they has been started
Currently, deferred tasks are running on the default version.

We have to set the [`AppEngineRouting`](https://cloud.google.com/tasks/docs/reference/rest/v2/AppEngineRouting).

... and in `QueryIter` too.

sveneberth:
* deferred tasks and `QueryIter`s are now running in the version in which they were called. (Resolves #641) This behaivor is equal as in the viur-server. But the version can be changed with the argument `_target_version`.
* Implement `_eta` and `_name`: These params were handled directly by the google api in the python2 runtime (https://cloud.google.com/appengine/docs/legacy/standard/python/refdocs/google.appengine.api.taskqueue.taskqueue#google.appengine.api.taskqueue.taskqueue.add) and were not ported.
* Remove `_retry_options`: This option does not longer exist as in the python2 runtime. We could set `retry` to `google.api_core.retry.Retry` (https://github.com/googleapis/python-api-core/blob/b8e75cff8bf6772ab76ca6f8d3c80dc2580075c7/google/api_core/retry.py#L222) but this expects a callable which must be pickled and is more complex to handle.
* Replace the `dict`s with google api objects. This is the only official accepted type for `create_task` arguments as documented in https://cloud.google.com/python/docs/reference/cloudtasks/latest/google.cloud.tasks_v2.services.cloud_tasks.CloudTasksClient#google_cloud_tasks_v2_services_cloud_tasks_CloudTasksClient_create_task.

### DIFF
--- a/core/tasks.py
+++ b/core/tasks.py
@@ -533,7 +533,7 @@ def CallDeferred(func: Callable) -> Callable:
                     http_method=tasks_v2.HttpMethod.POST,
                     relative_uri=taskargs["url"],
                     app_engine_routing=tasks_v2.AppEngineRouting(
-                        version=taskargs.get("target_version") or os.getenv("GAE_VERSION"),
+                        version=taskargs.get("target_version", conf["viur.instance.app_version"]),
                     ),
                 ),
             )
--- a/core/tasks.py
+++ b/core/tasks.py
@@ -708,7 +708,7 @@ class QueryIter(object, metaclass=MetaQueryIter):
                     http_method=tasks_v2.HttpMethod.POST,
                     relative_uri="/_tasks/queryIter",
                     app_engine_routing=tasks_v2.AppEngineRouting(
-                        version=os.getenv("GAE_VERSION"),
+                        version=conf["viur.instance.app_version"],
                     ),
                 )
             ),
--- a/core/tasks.py
+++ b/core/tasks.py
@@ -1,18 +1,19 @@
 import base64
-import grpc
 import json
 import logging
 import os
-import pytz
 import sys
-import requests
 from datetime import datetime, timedelta
 from functools import wraps
+from time import sleep
+from typing import Any, Callable, Dict, Optional, Tuple
+
+import grpc
+import pytz
+import requests
 from google.cloud import tasks_v2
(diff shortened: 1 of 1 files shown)


core/tasks.py
```diff
@@ -663,17 +700,19 @@ def _requeueStep(cls, qryDict: Dict[str, Any]) -> None:
...
-        taskClient.create_task(parent=parent, task=task)
+        taskClient.create_task(tasks_v2.CreateTaskRequest(
+            parent=taskClient.queue_path(conf["viur.instance.project_id"], queueRegion, cls.queueName),
+            task=tasks_v2.Task(
+                app_engine_http_request=tasks_v2.AppEngineHttpRequest(
+                    body=json.dumps(preprocessJsonObject(qryDict)).encode("UTF-8"),
+                    http_method=tasks_v2.HttpMethod.POST,
+                    relative_uri="/_tasks/queryIter",
+                    app_engine_routing=tasks_v2.AppEngineRouting(
+                        version=os.getenv("GAE_VERSION"),
```
ArneGudermann, line 711:
```suggestion
                        version=conf["viur.instance.app_version"],
```
You can use here a conf var.

core/tasks.py
```diff
@@ -498,25 +527,33 @@ def task():
...
-            }
-
-            # Set a schedule time in case countdown was set.
+            task = tasks_v2.Task(
+                app_engine_http_request=tasks_v2.AppEngineHttpRequest(
+                    body=json.dumps(preprocessJsonObject((command, (funcPath, args, kwargs, env)))).encode("UTF-8"),
+                    http_method=tasks_v2.HttpMethod.POST,
+                    relative_uri=taskargs["url"],
+                    app_engine_routing=tasks_v2.AppEngineRouting(
+                        version=taskargs.get("target_version") or os.getenv("GAE_VERSION"),
```
ArneGudermann, line 536:
```suggestion
                        version=taskargs.get("target_version", conf["viur.instance.app_version"]),
```
We can use here `conf["viur.instance.app_version"]` too.

phorward (approved):
I'm happy with this so far.

PR #656, opened 2023-02-22
Refactoring `securitykey` module

phorward:
This is a refactoring of the entire module without functional changes.

### DIFF
--- a/core/securitykey.py
+++ b/core/securitykey.py
@@ -35,12 +35,13 @@ SECURITYKEY_KINDNAME = "viur-securitykeys"
 
 def create(duration: Union[None, int] = None, **custom_data) -> str:
     """
-        Creates a new onetime Securitykey or returns the current sessions csrf-token.
-        The custom data (given as keyword arguments) that can be stored with the key if :param:duration is set must
-        be serializable by the datastore.
+        Creates a new one-time security key or returns the current sessions CSRF-token.
 
-        :param duration: Make this key valid for a fixed timeframe (and independent of the current session)
-        :returns: The new onetime key
+        The custom data (given as **custom_data) that can be stored with the key.
+        Any data provided must be serializable by the datastore.
+
+        :param duration: Make this key valid for a fixed timeframe of seconds (and independent of the current session)
+        :returns: The new one-time key, which is a randomized string.
     """
     if not duration:
         if custom_data:
@@ -68,6 +69,7 @@ def validate(key: str, useSessionKey: bool) -> Union[bool, db.Entity]:
 
         If useSessionKey is True, the key is expected to be the session's current security key
         or its static security key.
+
         Otherwise, it must be a key created with a duration, so that it is session independent.
 
         :param key: The key to be validated
--- a/core/securitykey.py
+++ b/core/securitykey.py
@@ -30,7 +30,7 @@ from viur.core import utils, current, db, tasks
 from datetime import datetime, timedelta
 from typing import Union
 
-SECURITYKEY_KIND_NAME = "viur-securitykeys"
+SECURITYKEY_KINDNAME = "viur-securitykeys"
 
 
 def create(duration: Union[None, int] = None, **kwargs) -> str:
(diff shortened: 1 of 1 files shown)


core/securitykey.py
```diff
@@ -1,33 +1,36 @@
...
+from viur.core import utils, current, db, tasks
 from datetime import datetime, timedelta
-from viur.core.utils import generateRandomString
-from viur.core import current, db
-from viur.core.tasks import PeriodicTask, CallDeferred
 from typing import Union
-from viur.core.utils import utcNow
 
-securityKeyKindName = "viur-securitykeys"
+SECURITYKEY_KINDNAME = "viur-securitykeys"
```
sveneberth, line 33: I know you probably cannot answer this question, but why is this plural? As usually kind/schema names are singular (user~s~, file~s~, ...).

```suggestion
SECURITYKEY_KINDNAME = "viur-securitykey"
```

core/securitykey.py
```diff
@@ -42,25 +45,33 @@ def create(duration: Union[None, int] = None, **kwargs) -> str:
...
     duration = int(duration)
-    dbObj = db.Entity(db.Key(securityKeyKindName, key))
+
+    entity = db.Entity(db.Key(SECURITYKEY_KINDNAME, key))
+
     for k, v in kwargs.items():
-        dbObj[k] = v
-    dbObj["until"] = utcNow() + timedelta(seconds=duration)
-    db.Put(dbObj)
+        entity[k] = v
```
sveneberth, line 55: Just call `.update()` or use the merge operator:
```suggestion
   entity |= kwargs
```

core/securitykey.py
```diff
@@ -69,37 +80,40 @@ def validate(key: str, useSessionKey: bool) -> Union[bool, db.Entity]:
             skeyHeaderValue = request.request.headers.get("Sec-X-ViUR-StaticSKey")
```
sveneberth, line 80:
```suggestion
            skey_header_value = request.request.headers.get("Sec-X-ViUR-StaticSKey")
```

and ff.

core/securitykey.py
```diff
@@ -42,25 +45,33 @@ def create(duration: Union[None, int] = None, **kwargs) -> str:
     if not duration:
         assert not kwargs, "kwargs are not allowed when session security key is wanted"
```
sveneberth, line 46:
```suggestion
        if kwargs:
            raise ValueError("kwargs are not allowed when session security key is wanted")
```

core/securitykey.py
```diff
@@ -1,33 +1,36 @@
...
-from viur.core import current, db
-from viur.core.tasks import PeriodicTask, CallDeferred
 from typing import Union
-from viur.core.utils import utcNow
 
-securityKeyKindName = "viur-securitykeys"
+SECURITYKEY_KINDNAME = "viur-securitykeys"
 
 
 def create(duration: Union[None, int] = None, **kwargs) -> str:
```
sveneberth, line 36:
```suggestion
def create(duration: Union[None, int] = None, **custom_data) -> str:
```
and ff. 

to clarify that these are not parameters passed to a super call but individual data.

sveneberth (changes requested):

core/securitykey.py
phorward: I'm totally with you, but this makes any outstanding securitykey of a viur-core<3.4 invalid as it is not accessible anymore.

phorward:
@sveneberth thanks for the review, I've updated the code according to your suggestions.
With the renaming of `viur-securitykeys` into `viur-securitykey` I'm not sure. Shall we discuss this in the viur meeting?

core/securitykey.py
sveneberth: > @sveneberth thanks for the review, I've updated the code according to your suggestions. With the renaming of `viur-securitykeys` into `viur-securitykey` I'm not sure. Shall we discuss this in the viur meeting?

Actually it is super unimportant. We can keep it in the back of our minds, but it can stay that way for now.
Maybe one could build in a configuration flag for the `kindname` for the transition.

I think in most projects that would not be a problem anyway if skeys were gone. Only a few really use some with data.

sveneberth (dismissed):

sveneberth (approved):

Issue #653, opened 2023-02-20
`JsonBone` should support JSON Schema validation


phorward:
In the previous VIUR meet-up, the proposal was made to extend `JsonBone` to allow for a JSON Schema validation.
This is absolutely useful and makes sense for several reasons and use-cases, and would allow to store JSON data in a valid form.

Related links:
- https://json-schema.org/
- https://json-schema.org/implementations.html#validator-python

PR #657, opened 2023-02-23
feat: Support JSON Schema validation for `JsonBone`

ArneGudermann:
This PR Implements the issue #653. It implements JSON Schema validation for the JsonBone and uses [jsonschema](https://pypi.org/project/jsonschema/).


phorward (changes requested):
Hello @ArneGudermann,
thanks for providing a fix for #653, I'm also the opinion that jsonschema is the best choice for this task.

Can you please apply my suggestions, and also provide some info in the docstring about how to use the schema parameter, best would be with a short (and working) example?

core/bones/json.py
```diff
@@ -1,5 +1,6 @@
 import ast
 import json
+import jsonschema
 from typing import Union
```
sveneberth, line 4:
```suggestion
from typing import Union, Union
```

core/bones/json.py
```diff
@@ -12,10 +13,13 @@ class JsonBone(RawBone):
 
     type = "raw.json"
 
-    def __init__(self, indexed: bool = False, multiple: bool = False, languages: bool = None, *args, **kwargs):
+    def __init__(self, indexed: bool = False, multiple: bool = False, languages: bool = None, schema: object = None,
```
sveneberth, line 16:
```suggestion
    def __init__(self, indexed: bool = False, multiple: bool = False, languages: bool = None, schema: Mapping = None,
```

core/bones/json.py
```diff
@@ -12,10 +13,13 @@ class JsonBone(RawBone):
...
+                 *args,
+                 **kwargs):
         assert not multiple
         assert not languages
         assert not indexed
+        self.schema = schema
```
sveneberth, line 22: Let's validate the schema on bone creation.
```suggestion
        # Validate the schema, if it's invalid a SchemaError will be raised
        jsonschema.validators.validator_for(False).check_schema(schema)
        self.schema = schema
```

sveneberth (changes requested):

core/bones/json.py
phorward: `from typing import Union, Union` ?

phorward (dismissed):

core/bones/json.py
sveneberth: The second is `Mapping`, not `Union`.

core/bones/json.py
sveneberth: @ArneGudermann we should wrap this with an `if`. I get 
```py
jsonschema.exceptions.SchemaError: None is not of type 'object', 'boolean'
```

sveneberth (changes requested):

sveneberth (approved):

core/bones/json.py
```diff
@@ -1,21 +1,36 @@
...
+import jsonschema
+from typing import Union, Mapping
 from viur.core.bones.base import ReadFromClientError, ReadFromClientErrorSeverity
 from viur.core.bones.raw import RawBone
 
 
 class JsonBone(RawBone):
     """
     This bone saves its content as a JSON-string, but unpacks its content to a dict or list when used.
+    :param schema If provided we can controll what data we accept.
```
phorward, line 12:
```suggestion
    :param schema If provided we can control and verify which data to accept.
```

core/bones/json.py
```diff
@@ -1,21 +1,36 @@
...
+    Example:
+        >>>schema = {"type" : "object",
+        >>>     "properties" : {
+        >>>"price" : {"type" : "number"},
+        >>>"name" : {"type" : "string"},
+        >>>}}
+        We only accept the json if the price is a number and name is a string.
```
phorward, line 19:
```suggestion
        >>> schema = {"type" : "object",
        >>>     "properties" : {
        >>>         "price" : {"type" : "number"},
        >>>         "name" : {"type" : "string"},
        >>>     }
        >>> }
        This will only accept the provided JSON when price is a number and name is a string.
```

phorward (changes requested):
Can you please fix the provided suggestions?

core/bones/json.py
```diff
@@ -9,14 +9,15 @@
 class JsonBone(RawBone):
     """
     This bone saves its content as a JSON-string, but unpacks its content to a dict or list when used.
-    :param schema If provided we can controll what data we accept.
+    :param schema If provided we can control and verify which data to accept.
     Example:
-        >>>schema = {"type" : "object",
+        >>> schema = {"type" : "object",
```
phorward, line 14:
```suggestion
        >>> schema = {
        >>>         "type" : "object",
```

phorward (changes requested):
one moment

phorward (approved):

core/bones/json.py
```diff
@@ -1,21 +1,38 @@
...
 from viur.core.bones.raw import RawBone
 
 
 class JsonBone(RawBone):
     """
     This bone saves its content as a JSON-string, but unpacks its content to a dict or list when used.
+    :param schema If provided we can control and verify which data to accept.
+    Example:
+        >>> schema = {
+        >>>         "type" : "object",
```
phorward, line 15:
```suggestion
        >>>     "type" : "object",
```

phorward (dismissed):

sveneberth (approved):

phorward (approved):

Issue #650, opened 2023-02-09
default html-templates for 400/500/... errors


cheeaza:
It would be nice if there is a default behavior,  which html-file is delivered if there is an error appearing.

The core could simply search for an error_404.html or error_500.html or error_412.html file and if said file does not exist, then use an error.html as a fallback.

phorward:
Implemented by #658

PR #658, opened 2023-02-23
Add search for Error Template in html/error

ArneGudermann:
Hello,
This PR solves #650.
If an error occurs, the core now looks in the `html/error` directory for a corresponding file for the error code and renders it if it finds one.

### DIFF
--- a/core/request.py
+++ b/core/request.py
@@ -318,15 +318,28 @@ class BrowseHandler():  # webapp.RequestHandler
                         res["traceback"] = traceback.format_exc()
 
                     res = json.dumps(res)
-                else:
-                    with (conf["viur.instance.core_base_path"].joinpath(conf["viur.errorTemplate"]).open() as tpl_file):
-                        tpl = Template(tpl_file.read())
-                        res = tpl.safe_substitute({
+                else:  # We render the error in html
+                    # first we try to get the template from hmt/error/
+                    template_file = conf["viur.instance.project_base_path"].joinpath(f"html/error/{e.status}.html")
+                    if template_file.is_file():
+                        template = conf["viur.mainApp"].render.getEnv().get_template(
+                            conf["viur.mainApp"].render.getTemplateFileName(f"error/{e.status}"))
+
+                        res = template.render({
                             "error_code": e.status,
                             "error_name": translate(e.name),
                             "error_descr": e.descr,
                         })
-                    extendCsp({"style-src": ['sha256-Lwf7c88gJwuw6L6p6ILPSs/+Ui7zCk8VaIvp8wLhQ4A=']})
+                    else:
+                        with (conf["viur.instance.core_base_path"].joinpath(
+                                conf["viur.errorTemplate"]).open() as tpl_file):
+                            tpl = Template(tpl_file.read())
+                            res = tpl.safe_substitute({
+                                "error_code": e.status,
+                                "error_name": translate(e.name),
+                                "error_descr": e.descr,
+                            })
+                        extendCsp({"style-src": ['sha256-Lwf7c88gJwuw6L6p6ILPSs/+Ui7zCk8VaIvp8wLhQ4A=']})
             self.response.write(res.encode("UTF-8"))
 
         except Exception as e:  # Something got really wrong


core/request.py
```diff
@@ -318,15 +318,28 @@ def processRequest(self) -> None:
                         res["traceback"] = traceback.format_exc()
 
                     res = json.dumps(res)
-                else:
-                    with (conf["viur.instance.core_base_path"].joinpath(conf["viur.errorTemplate"]).open() as tpl_file):
-                        tpl = Template(tpl_file.read())
-                        res = tpl.safe_substitute({
+                else:  # We render the error in html
+                    # first we try to get the template from hmt/error/
```
phorward, line 322:
```suggestion
                    # Try to get the template from html/error/
```

phorward (dismissed):
I'm mostly happy with this.

phorward (approved):

phorward:
@ArneGudermann can you approve this PR please, as I committed a typo in the comment, and cannot approve it therefore.

sveneberth (changes requested):
Can we add a fallback to `html/error/generic.html` or similar?

This would bring the possibility to handle all codes in one template (e.g. with `elif` blocks), or to create individual templates only for certain error codes and to display the rest in the same way (but still individually).

phorward:
> Can we add a fallback to `html/error/generic.html` or similar?

Good point. Indeed, this would become a breaking change to existing implementations. A deprecated fallback handling to the current code should remain (with a warning on `conf["viur.errorTemplate"]`, of course!)

phorward:
What to render in case the error template throws an Exception?

core/request.py
```diff
@@ -318,15 +318,28 @@ def processRequest(self) -> None:
...
+                    # Try to get the template from html/error/
+                    template_file = conf["viur.instance.project_base_path"].joinpath(f"html/error/{e.status}.html")
+                    if template_file.is_file():
+                        template = conf["viur.mainApp"].render.getEnv().get_template(
+                            conf["viur.mainApp"].render.getTemplateFileName(f"error/{e.status}"))
+
+                        res = template.render({
                             "error_code": e.status,
                             "error_name": translate(e.name),
                             "error_descr": e.descr,
```
phorward, line 331: Shall we use the same dict as with #614 and #690 here? So without the "error_"-prefix?

core/request.py
```diff
@@ -318,15 +318,28 @@ def processRequest(self) -> None:
...
                         })
-                    extendCsp({"style-src": ['sha256-Lwf7c88gJwuw6L6p6ILPSs/+Ui7zCk8VaIvp8wLhQ4A=']})
+                    else:
+                        with (conf["viur.instance.core_base_path"].joinpath(
+                                conf["viur.errorTemplate"]).open() as tpl_file):
+                            tpl = Template(tpl_file.read())
+                            res = tpl.safe_substitute({
+                                "error_code": e.status,
+                                "error_name": translate(e.name),
+                                "error_descr": e.descr,
```
phorward, line 340: Same here. We can use the same dict keys.

phorward (changes requested):
What do you think about my two comments?

core/request.py
```diff
@@ -309,44 +309,47 @@ def processRequest(self) -> None:
...
                         else:
-                            res = tpl.safe_substitute(
-                                {"error_code": "500",
-                                 "error_name": "Internal Server Error",
-                                 "error_descr": descr,
-                                 })
+                            with (conf["viur.instance.core_base_path"].joinpath(
+                                    conf["viur.errorTemplate"]).open() as tpl_file):
+                                template_file = Template(tpl_file.read())
+                                res = template_file.safe_substitute(error_info)
```
sveneberth, line 350: You need to rename the variables in `error.html` too. However, IMO we should use here Jinja as renderer too.

Then you can move the template from `template/error.html` to `template/error/generic.html` and make the `elif conf["viur.instance.project_base_path"].joinpath("html/error/generic.html").is_file():` to the `else` block. If the generic template would be exist in the project folder, `getTemplateFileName()` would choose the template inside the core.

So the final code should look like 
```py
                        if conf["viur.instance.project_base_path"].joinpath(f"html/error/{e.status}.html").is_file():
                            template = conf["viur.mainApp"].render.getEnv().get_template(
                                conf["viur.mainApp"].render.getTemplateFileName(f"error/{e.status}"))
                            res = template.render(error_info)
                        else:
                            template = conf["viur.mainApp"].render.getEnv().get_template(
                                conf["viur.mainApp"].render.getTemplateFileName("error/generic.html"))
                            res = template.render(error_info)
```

sveneberth (changes requested):

phorward:
Okay... where to start? First of all, this pull request did not work as expected, which had several causes. So I started to rework the function `getTemplateFileName` in the Jinja-Renderer. But this was not enough, because of the recent changes in the conf-module (started in viur-core 3.3), the path of the viur-core source, which is IMHO either questionable, were broken. 

I've committed my revision of this pull request with 676749bb2e39e528d79b28aead047c1b032cc2fd, and it works so far also. But the places marked with `#fixme` raise bigger or fundamental questions and problems, where we have to work on and which might have to be revised in separate pull requests before.

Because of these fundamental issues and problems, I've decided not to include this pull request in the upcoming viur-core 3.4 release, and it has to wait until the next version.

phorward (changes requested):
See my changes in 676749bb2e39e528d79b28aead047c1b032cc2fd and we have to do a general replacement for the template filename search.

sveneberth (approved):

phorward (dismissed):

sveneberth (approved):

PR #664, opened 2023-03-02
Implement `script` system-module for Scriptor tree

phorward:
viur-scriptor is being integrated into the viur-core as a standard component. This commit implements the script-module and was initially created in a private repository.

This feature is mandatory for viur-core==3.4!

### DIFF
--- a/core/modules/script.py
+++ b/core/modules/script.py
@@ -1,0 +1,154 @@
+from viur.core.bones import *
+from viur.core.prototypes.tree import Tree, TreeSkel
+from viur.core import db, utils, conf, skeleton
+from viur.core.prototypes.tree import Tree
+import re
+
+
+# pre-compile patterns for vfuncs
+DIRECTORY_PATTERN = re.compile(r'^[a-zA-Z0-9äöüÄÖÜ_-]*$')
+FILE_PATTERN = re.compile(r'^[a-zA-Z0-9äöüÄÖÜ_-]+?.py$')
+
+
+def utils_get_path(skel, data = None):
+    script: Tree = conf["viur.mainApp"].vi.script
+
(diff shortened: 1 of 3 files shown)


core/modules/script.py
```diff
@@ -0,0 +1,154 @@
...
+            query = self.editSkel("node").all().filter("parententry =", skel["key"])
+
+            for entry in query.fetch(99):
+                self.onEdit("node", entry)
+                entry.toDB()
+
+            query = self.editSkel("leaf").all().filter("parententry =", skel["key"])
+            for entry in query.fetch(99):
+                self.onEdit("leaf", entry)
+                entry.toDB()
```
sveneberth, line 135: I'm not sure if we really should run this in the same request. This could lead to a long response time which might not be user-friendly.

sveneberth (approved):

core/modules/script.py
phorward: Then we should optimize this previously.

phorward:
@sveneberth I've refactored the path resolving and updating to run deferred and with cursor.
I hope this is okay for the first release now. I've tested it with a smaller file/folder structure.

core/modules/script.py
phorward: I've optimized this.

sveneberth (approved):

sveneberth:
@viur-framework/developers 
For people who used the old mausbrand-internal script module and wanted to copy the existing scripts into the kinds with the viur prefix, I once (for my own use) created a quick&dirty script and uploaded it as gist.

https://gist.github.com/sveneberth/f2b8133cdf9881c8465c29d51096a9bb

PR #667, opened 2023-03-06
feat: Implement naive mode in `DateBone`

sveneberth:
Because the datastore (currently) enforces datetime aware objects we need to remove/add the tzinfo on (un)serialize.

This is a replacement for #642

### DIFF
--- a/core/bones/date.py
+++ b/core/bones/date.py
@@ -1,9 +1,12 @@
-from viur.core.bones.base import BaseBone, ReadFromClientError, ReadFromClientErrorSeverity
-from viur.core import db, request, conf, current
-from viur.core.utils import utcNow
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 from typing import Dict, Optional
-import pytz, tzlocal
+
+import pytz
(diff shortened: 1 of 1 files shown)


phorward (approved):

phorward:
I'm fine with this solution.
@AchimSchumacher can you please re-review on this and add your approval?

core/bones/date.py
```diff
@@ -102,7 +111,9 @@ def singleValueFromClient(self, value: str, skel: 'viur.core.skeleton.SkeletonIn
                 value = datetime.fromtimestamp(float(rawValue), tz=time_zone).replace(microsecond=0)
         elif not self.date and self.time:
             try:
-                value = time_zone.localize(datetime.fromisoformat(value))
+                value = datetime.fromisoformat(value)
+                if self.naive:
```
AchimSchumacher, line 115: Same error here: It must be 'if not self.naive:'

AchimSchumacher (changes requested):

phorward (approved):

phorward (approved):

PR #669, opened 2023-03-06
Fix: Set the logging level to `logging.DEBUG`

sveneberth:
Otherwise no logging is displayed on my local dev server.


ArneGudermann (approved):

sveneberth:
I've no clue why it didn't work yesterday. But now I've no longer problems with any log level.

PR #680, opened 2023-03-08
refactor: Re-implement password encoding using Python's `hashlib` and `secrets` module

sveneberth:
Since python 3.4 https://docs.python.org/3/library/hashlib.html#hashlib.pbkdf2_hmac is in the standard. This PR replaces the own implementation with this one.

Furthermore, for security relevant tokens not the [`random`](https://docs.python.org/3/library/random.html#module-random) module but [`secrets`](https://docs.python.org/3/library/secrets.html) should be used. Therefore there is now an extension of `utils.generateRandomString`.
> _Warning The pseudo-random generators of this module should not be used for security purposes. For security or cryptographic uses, see the [secrets](https://docs.python.org/3/library/secrets.html#module-secrets) module._
~ [Docs at python.org](https://docs.python.org/3/library/random.html#:~:text=Warning%20The%20pseudo%2Drandom%20generators%20of%20this%20module%20should%20not%20be%20used%20for%20security%20purposes.%20For%20security%20or%20cryptographic%20uses%2C%20see%20the%20secrets%20module.)

Furthermore, the 600,000 iterations recommended by OWASP are now used instead of 1001 for PBKDF2.
https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#pbkdf2

The new method `encode_password` directly returns the hash as well as meta data for the embedded entity in the datastore.

Finally, this PR replaces the char-by-char comparisons with [`hmac.compare_digest`](https://docs.python.org/3/library/hmac.html#hmac.compare_digest), which is safe against timing attacks.

### DIFF
--- a/core/bones/password.py
+++ b/core/bones/password.py
@@ -1,24 +1,11 @@
-import binascii
 import hashlib
-import logging
-import string
-from typing import List, Union
-
-from viur.core import conf, utils
-from viur.core.bones.base import ReadFromClientError, \
-    ReadFromClientErrorSeverity
-from viur.core.bones.string import StringBone
-from viur.core.i18n import translate
-from viur.core import utils, conf
-from hashlib import sha256
-import hmac
-import codecs
 import re
-from struct import Struct
-from operator import xor
-from itertools import starmap
 from typing import List, Tuple, Union
 
+from viur.core import conf, utils
+from viur.core.bones.string import StringBone
+from viur.core.i18n import translate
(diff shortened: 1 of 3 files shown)


ArneGudermann (approved):

phorward (approved):
I'm happy with this. Anyway, I would like to present this PR to the viur-meeting and give everyone else a chance to review.

PR #682, opened 2023-03-08
refactor: Refactor password recovery process in stock `User`-module

ArneGudermann:
Rewrite Password recovery.

PR #687, opened 2023-03-09
Refactor Password Recovery

ArneGudermann:
This a new attempt on top off #682


sveneberth:
I restored #682 and re-opened it

PR #686, opened 2023-03-09
Add JSON Response for internal server errors too

ArneGudermann:
Create a json error message if we are in the json/vi renderer and have a normal Exception not a HttpException.

PR #690, opened 2023-03-13
fix: Return JSON-encoded response for internal server errors too

ArneGudermann:
Rework of #686
Create a json error message if we are in the json/vi renderer and have a normal Exception not a HttpException.


phorward (approved):
I'm fine with this now, thanks for re-setting this PR up.
We should clean the issue proposed in #658, and use the same dict keys for both JSON and Jinja-skel-mapping in the error templates.

PR #691, opened 2023-03-14
perf: Load session only when requested

ArneGudermann:
This PR should improve session loading. Currently the session is loaded on every request even if not necessary. With this PR we now only load the session if needed. Like getting the current user and so on

### DIFF
--- a/core/__init__.py
+++ b/core/__init__.py
@@ -302,6 +302,7 @@ def app(environ: dict, start_response: Callable):
     current.request.set(handler)
     current.session.set(session.Session(handler))
     current.request_data.set({})
+    current.user.set(current.CurrentUserWrapper())
     # Handle request
     handler.processRequest()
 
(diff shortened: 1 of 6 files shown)


sveneberth (commented):
Hey @ArneGudermann, I like the idea, but I'm not happy with the implementation yet. But I don't have an ad hoc alternative suggestion either. I'll have to take another look at it. It might be necessary to start somewhere else.

I don't like the multiple `if not self.loaded: self.load()` lines in the `session` module.

And I find this swapping of the `CurrentUser` instance with the `ContextVar` very hacky. 
[In the documentation, it is advised to create ContextVars only at top module level.](https://docs.python.org/3/library/contextvars.html#:~:text=Important%3A%20Context%20Variables%20should%20be%20created%20at%20the%20top%20module%20level)

phorward:
We should give it another try in viur-core 3.8 @ArneGudermann ?

phorward:
@ArneGudermann was this now replaced by #1277, I think so?

phorward:
Closed as this was replaced by #1277.

PR #692, opened 2023-03-17
chore: Conventional commits and clarifications

phorward:
Adding rules for conventional commits to CONTRIBUTING.md, and for the versioning scheme regarding pre-releases.

However, I have to make two personal comments here about Conventional Commits:

1. I really HATE the keyword "chore".
2. Why is it called "feat", and not "feature", but "refactor" and not "refact"? This is so ugly and stupid, it seems that Conventional Commits where invented by idiots... well, it has its origin in the JavaScript world, this might be the reason.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -48,10 +48,10 @@ Please document your changes and provide info in any form you can. We have estab
 `viur-core` uses the semantic versioning scheme.<br>
 Any `major.minor.bugfix` release is being published to [PyPI](https://pypi.org/project/viur-core).
 
-Furthermore, the following rules apply to hidden pre-releases which are also made to PyPI for open tests.
+Furthermore, the following rules apply to pre-releases which are also made available to PyPI for open tests.
 
-- Release candidates which are almost feature-complete is given a suffix named `-rcN`.
-- Beta versions with incomplete features is given a suffix named `-betaN`.
+- Release candidates which are almost feature-complete is given a suffix named `.rcN`.
+- Beta versions with incomplete features is given a suffix named `.betaN`.
 
 In both cases, `N` is a number counted upwards for every pre-release.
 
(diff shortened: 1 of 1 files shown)


sveneberth:
To be honest, I don't think much of this _Conventional Commits_ convention. I would never use them in my projects. I find most of these abbreviations and identifiers very ugly.
There have been various commit conventions (e.g. [gitmoji.dev](https://gitmoji.dev/)), but they haven't really caught on and I've never seen them in large projects. The repository also has only [5.1k likes](https://github.com/conventional-commits/conventionalcommits.org/stargazers) -- which is not really much and speaks for a good standard.

This "specification" also seems to me to be absolutely incomplete and unspecific. 
1. They specify only _feat_ and _fix_. Others are optional, they recommend to use the [angular keywords](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type), but doesn't enforce them.
   ~ [ref](https://github.com/conventional-commits/conventionalcommits.org/blob/2446255d79182684d0a752e07e5a427e7bb2b581/content/v1.0.0/index.md?plain=1#L33-L38)
2. Since the precedence rules of the types are not regulated there either, this grammar is ambiguous. A bugfix within the CI could be [A] "fix: config path in CI" but also [B] "ci: Fix config path" or [C] "fix(ci): config path". 
   What use is a specification with so much leeway?
3. I especially miss the important, long-established rule:
   - max. line length (esp. summary line) [ref](https://git-scm.com/docs/git-commit#_discussion)
   - use the imperative mood [ref](https://git.kernel.org/pub/scm/git/git.git/tree/Documentation/SubmittingPatches?h=v2.36.1#n181)
   - Capitalize the subject line and do not end it with a period [ref](https://cbea.ms/git-commit)


I am clearly against this convention.
In general, I welcome a convention, and if we decide on one, I will also follow it within the cores. However, I would prefer a more meaningful and complete specification of our own.

phorward:
> I am clearly against this convention. In general, I welcome a convention, and if we decide on one, I will also follow it within the cores. However, I would prefer a more meaningful and complete specification of our own.

We have to discuss this. This was only a proposal. I don't like "conventional commits" either, as pointed out above, but well, it is a convention that seems to work and is quite easy to understand and adapt.

sveneberth:
In ViUR-meeting 2023-03-31 we decided to create a own specification with custom keywords.

PR #697, opened 2023-03-29
fix: continue thumbnailing when image is broken

XeoN-GHMB:
continue thumbnailing if a image is broken


xnopasaranx (changes requested):
Looks good to me, linter has a problem with the line length though, so a linebreak would be nice.

phorward (approved):

sveneberth (approved):

PR #699, opened 2023-03-31
fix: Keep filename synchronous in both skeleton and blob

ArneGudermann:
Currently you can not rename a file in viur. If you change the filename the core can't find the file anymore because the name is only changed in the database not in the GCloud Storage. Now we move the file to the new name and delete the old file. Because GCloud Storage can't rename a file. 
(https://cloud.google.com/storage/docs/copying-renaming-moving-objects#move)


sveneberth (changes requested):

sveneberth (approved):

phorward (approved):
Thanks @ArneGudermann and @sveneberth, this is a nice solution!

PR #702, opened 2023-04-05
fix: `CredentialBone` without escaping

ArneGudermann:
This PR remove the string escaping in the credentialBone because it change the credetial value. 
An example we have an Base32 Secret like `GZRTOMBZREHFTDQNJVGQYRYMRSTOOLCGNSWKNQ==`
and we store it in the Bone. 
Later  we have want  compare the token in some OTP function or so and now we get `GZRTOMBZREHFTDQNJVGQYRYMRSTOOLCGNSWKNQ&#061;&#061;` from the Database.

### DIFF
--- a/core/bones/credential.py
+++ b/core/bones/credential.py
@@ -14,10 +14,10 @@ class CredentialBone(StringBone):
     def __init__(
         self,
         *,
-        maxLength: int = 100*1024,  # Limit the Bone length on 100KB
+        maxLength: int = 100 * 1024,  # Limit the Bone length on 100KB
         **kwargs
     ):
-        super().__init__(maxLength=maxLength,**kwargs)
+        super().__init__(maxLength=maxLength, **kwargs)
         if self.multiple or self.languages:
             raise ValueError("Credential-Bones cannot be multiple or translated!")
 
--- a/core/bones/credential.py
+++ b/core/bones/credential.py
@@ -14,7 +14,7 @@ class CredentialBone(StringBone):
     def __init__(
         self,
         *,
-        maxLength: int = 100 * 1024,  # Limit the Bone length on 100KB
+        maxLength: int = None, # Unlimited length
         **kwargs
     ):
         super().__init__(maxLength=maxLength, **kwargs)
@@ -28,7 +28,7 @@ class CredentialBone(StringBone):
         """
         if value is None:
             return False
-        if len(value) > self.maxLength:
+        if self.maxLength is not None and len(value) > self.maxLength:
             return "Maximum length exceeded"
 
     def serialize(self, skel: 'SkeletonInstance', name: str, parentIndexed: bool) -> bool:
--- a/core/bones/credential.py
+++ b/core/bones/credential.py
@@ -14,7 +14,7 @@ class CredentialBone(StringBone):
     def __init__(
         self,
         *,
-        maxLength: int = None, # Unlimited length
+        maxLength: int = None,  # Unlimited length
         **kwargs
     ):
         super().__init__(maxLength=maxLength, **kwargs)
--- a/core/bones/credential.py
+++ b/core/bones/credential.py
@@ -11,10 +11,25 @@ class CredentialBone(StringBone):
     """
     type = "str.credential"
 
-    def __init__(self, **kwargs):
-        super().__init__(**kwargs)
+    def __init__(
+        self,
+        *,
+        maxLength: int = 100*1024,  # Limit the Bone length on 100KB
+        **kwargs
+    ):
+        super().__init__(maxLength=maxLength,**kwargs)
         if self.multiple or self.languages:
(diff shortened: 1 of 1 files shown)


sveneberth (commented):
This also removes the length limit of 4 KB. Which is also good in my opinion, because if the value is too long, an error should occur and it should not be silently truncated. 
Nevertheless had the maximum length here a sense?

ArneGudermann:
Mhh I dont't know but the `credatialBone` inherit from the `stringBone` an  by default the `stringBone` has an limit length from 254 . Maybe we should change this to a higher number in the  `credatialBone`

sveneberth:
> `credatialBone` inherit from the `stringBone` an by default the `stringBone` has an limit length from 254

That's right, but `singleValueFromClient` was already customized with this line:
https://github.com/viur-framework/viur-core/blob/38e1a775fb8d2e7d1dafd22a10c65bf91f8a46c6/core/bones/credential.py#L39
where `4 * 1024` is the argument for the `maxLength` in [`utils.escapeString`](https://github.com/viur-framework/viur-core/blob/38e1a775fb8d2e7d1dafd22a10c65bf91f8a46c6/core/utils.py#L71).

Now it's unlimited. Which is why my question was _if this is a problem_?

ArneGudermann:
Unlimited is rigth for the core side. But we send in the Bone structure a `maxLength` of 254. And the VI for example limit the input on this value. So we must set the value to something higher like 4 KB. I don't know a good value for this.

core/bones/credential.py
```diff
@@ -12,11 +11,26 @@ class CredentialBone(StringBone):
...
 
+    def isInvalid(self, value):
+        """
+            Returns None if the value would be valid for
+            this bone, an error-message otherwise.
+        """
+        if value is None:
+            return False
+        if len(value) > self.maxLength:
+            return "Maximum length exceeded"
```
sveneberth, line 32: IMO we should this check generally in the `StringBone`.

core/bones/credential.py
```diff
@@ -12,11 +11,26 @@ class CredentialBone(StringBone):
     """
     type = "str.credential"
 
-    def __init__(self, **kwargs):
-        super().__init__(**kwargs)
+    def __init__(
+        self,
+        *,
+        maxLength: int = 100 * 1024,  # Limit the Bone length on 100KB
```
sveneberth, line 17:
```suggestion
        maxLength: int = 0,  # Unlimited lenght
```

This would be analogous with the `StringBone`.
The condition is falsy and the value won't be cutted:
https://github.com/viur-framework/viur-core/blob/38e1a775fb8d2e7d1dafd22a10c65bf91f8a46c6/core/utils.py#L90-L91

sveneberth (changes requested):

core/bones/credential.py
ArneGudermann: `maxLength: int = 0, # Unlimited length`
 This is not a good solution  I think. For me it looks like the `maxLength` is 0 not Unlimited. When someone parse this value in a HTML input it not works as Unlimited. But rather nothing can be inputed.

core/bones/credential.py
ArneGudermann: Ok. But we need the `isInvalid` nevertheless in the `CredentialBone` for the is None check

core/bones/credential.py
phorward: What about `None`? 
```suggestion
        maxLength: int = None, # Unlimited length
```

core/bones/credential.py
```diff
@@ -12,11 +11,26 @@ class CredentialBone(StringBone):
...
             raise ValueError("Credential-Bones cannot be multiple or translated!")
 
+    def isInvalid(self, value):
+        """
+            Returns None if the value would be valid for
+            this bone, an error-message otherwise.
+        """
+        if value is None:
+            return False
+        if len(value) > self.maxLength:
```
phorward, line 31:
```suggestion
        if self.maxLength is not None and len(value) > self.maxLength:
```

core/bones/credential.py
phorward: Can you please turn this discussion into a separate issue? @sveneberth @ArneGudermann

phorward (changes requested):
Please checkout my suggestions, afterwards we can approve it.
Generally we should discuss if `utils.escapeString` or `StringBone` in general accepts `maxLength` with 0 as real 0 and None as "unlimited", or both 0 or None as unlimited.

sveneberth (approved):

core/bones/credential.py
phorward: I've merged develop containing #714 into this branch.

phorward (approved):

phorward (approved):

PR #704, opened 2023-04-11
feat: Improvements on `securitykey.validate`

phorward:
- `useSessionKey=True` is now default
- `pre_condition`-parameter to auto-raise a PreconditionFailed() exception
- Replaced and optimized calls in viur-core
- 100% backwards compatible to existing calls

I don't know why `useSessionKey=True` was always explicitly provided, as you can see below:
before

### DIFF
--- a/core/modules/file.py
+++ b/core/modules/file.py
@@ -553,8 +553,7 @@ class File(Tree):
         else:
             size = None
 
-        if not securitykey.validate(skey, useSessionKey=True):
-            raise errors.PreconditionFailed()
+        securitykey.validate(skey, pre_condition=True)
 
         targetKey, uploadUrl = self.initializeUpload(fileName, mimeType.lower(), node, size)
 
@@ -656,7 +655,7 @@ class File(Tree):
         if skelType == "leaf":  # We need to handle leafs separately here
             skey = kwargs.get("skey")
             targetKey = kwargs.get("key")
-            if not skey or not securitykey.validate(skey, useSessionKey=True) or not targetKey:
+            if not securitykey.validate(skey) or not targetKey:
                 raise errors.PreconditionFailed()
             skel = self.addSkel("leaf")
             if not skel.fromDB(targetKey):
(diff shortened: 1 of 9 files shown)


phorward:
This can be seen as an improvement on #656

core/modules/file.py
```diff
@@ -656,7 +655,7 @@ def add(self, skelType: SkelType, node=None, *args, **kwargs):
         if skelType == "leaf":  # We need to handle leafs separately here
             skey = kwargs.get("skey")
```
sveneberth, line 656: If `kwargs` doesn't contains the `skey` it's `None`. But the type hint enforces a `str` (`def validate(key: str, ...)`).

Therefore I suggest this change:

```suggestion
            skey = kwargs.get("skey", "")
```

sveneberth (changes requested):

sveneberth:
Let's talk about a decorator. I use this in ViUR2:

```
def skeyRequired(func=None, **decoratorKwArgs):
	"""Decorator that checks the skey before the method is called.

	Optional callable to pass keyword-arguments for the securitykey.validate-call.

	Example:
		>>> from server import exposed
		>>> @exposed
		>>> @skeyRequired
		>>> def yourMethod(self):
		>>>     return {"foo": "bar"}
	"""

	def outerWrapper(f):
		@functools.wraps(f)
		def wrapper(*args, **kwargs):
			if not securitykey.validate(kwargs.pop("skey", ""), **decoratorKwArgs):
				raise errors.PreconditionFailed("Missing or invalid skey")
			return f(*args, **kwargs)

		return wrapper

	if isinstance(func, (types.MethodType, types.FunctionType)):
		return outerWrapper(func)  # @skeyRequired
	else:
		return outerWrapper  # @skeyRequired() or @skeyRequired(**anyKwargs)
```

phorward:
In today's viur-meeting, we decided to solve it by a decorator. I'm with you, but anyway would like to merge this PR as is and integrate the feature set into the code-base.

phorward:
Closed as superseded by #764

PR #706, opened 2023-04-12
fix: Improvements and clarifications on version string

phorward:
This fixes the currently always succeeding version checking in viur-vi. This was NOT intended!

### DIFF
--- a/core/config.py
+++ b/core/config.py
@@ -43,7 +43,7 @@ class Conf(dict):
 
 # Some values used more than once below
 __project_id = google.auth.default()[1]
-__project_version = os.getenv("GAE_VERSION")
+__app_version = os.getenv("GAE_VERSION")
 
 # Determine our basePath (as os.getCWD is broken on appengine)
 __project_base_path = Path().absolute()
@@ -152,7 +152,7 @@ conf = Conf({
     "viur.file.derivers": {},
 
     # Name of this version as deployed to the appengine
-    "viur.instance.app_version": __project_version,
+    "viur.instance.app_version": __app_version,
 
     # The base path of the core, can be used to find file in the core folder
     "viur.instance.core_base_path": __core_base_path,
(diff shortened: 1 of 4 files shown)


ArneGudermann (approved):

phorward:
hello, after reconsideration, I've updated this PR for more clarification on how the version string must be created. It's best to have the first 3 parts as integers, where the fourth part is optional and contains the appendix. Furthermore, the /vi/getVersion function returns the entire version in case a user is authenticated.

sveneberth (changes requested):

sveneberth:
As decided in the ViUR-meeting please remove the user exception and expose always the patch level.

core/version.py
```diff
@@ -1 +1,6 @@
-__version__ = "3.4.0-beta3"
+# This is the place where the viur-core version number is defined;
+# For pre-releases, postfix with ".betaN" or ".rcN" where `N` is an incremented number for each pre-release.
+# This will mark it as a pre-release as well on PyPI.
+
+__version__ = "3.4.0.rc1"
+assert __version__.count(".") == 3  # semantic version number is mandatory!
```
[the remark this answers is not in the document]
phorward: You're right, but the condition was also wrong and I did an improvement. See latest commit 6b28973092b755fcbe23a8228b574bf4787bbe94

core/render/vi/__init__.py
```diff
@@ -105,7 +105,7 @@ def getVersion(*args, **kwargs):
     Returns viur-core version number
     """
     current.request.get().response.headers["Content-Type"] = "application/json"
-    if conf["viur.instance.is_dev_server"]:
+    if conf["viur.instance.is_dev_server"] or current.user.get():
```
phorward, line 108: Hello, we had a discussion today, that the version string should generally be exposed fully. I would like to discuss this later in a separate pull request.

core/config.py
```diff
@@ -43,7 +43,7 @@ def __setitem__(self, key, value):
 
 # Some values used more than once below
 __project_id = google.auth.default()[1]
-__version = os.getenv("GAE_VERSION")
+__project_version = os.getenv("GAE_VERSION")
```
sveneberth, line 46: The project itself has no version. Can we call it `app_version`, as in the conf dictionary?

```suggestion
__app_version = os.getenv("GAE_VERSION")
```

sveneberth (changes requested):

core/config.py
phorward: I've updated the PR. please re-re-review.

sveneberth (approved):

PR #712, opened 2023-04-14
feat: custom actions for user maintenance and debug triggers

phorward:
This provides user maintenance triggers to either kick users from the system or login as any user. This feature is only available for "root" users.

The PR enables to use translate()-objects in adminInfos as well.


core/modules/user.py
```diff
@@ -942,6 +988,30 @@ def getAuthMethods(self, *args, **kwargs):
...
+        match action:
+            case "takeover":
+                self.authenticateUser(skel["key"])
+
+            case "kick":
+                session.killSessionByUser(skel["key"])
+
+            case _:
+                raise errors.NotAcceptable(f"Action {action!r} not implemented")
+
```
sveneberth, line 1012:
```suggestion
current.request.get().response.headers["Content-Type"] = "application/json"
```

sveneberth (changes requested):

ArneGudermann (approved):

core/modules/user.py
```diff
@@ -942,6 +988,30 @@ def getAuthMethods(self, *args, **kwargs):
 
         return json.dumps(res)
 
+    @exposed
+    def trigger(self, action, key, skey):
```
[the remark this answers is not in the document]
sveneberth: @ArneGudermann This cannot work. If you add a type hint to a exposed method it must be unambiguous to be type-cast-able by `processTypeHint`.

This results now in 
```py
ValueError: TypeHint str | viur.datastore.types.Key not supported
```

core/modules/user.py
```diff
@@ -946,6 +992,30 @@ def getAuthMethods(self, *args, **kwargs):
 
         return json.dumps(res)
 
+    @exposed
+    def trigger(self, action: str, key: str | db.Key, skey: str):
```
sveneberth, line 996:
```suggestion
    def trigger(self, action: str, key: str, skey: str):
```

sveneberth (changes requested):

phorward:
Just as hint, the adminInfo can be subclasses this way now to inherit the actions:

```python
class User(User):

    adminInfo = User.adminInfo | {
        "columns": ["lastname", "firstname", "name", "responsibility"],
        "filter": {"orderby": "lastname"},
        "disabledActions": [
            "add",
            "clone",
            "delete"
        ],
    }
```

core/modules/user.py
sveneberth: @phorward Hmm, This isn't resolved?!

core/modules/user.py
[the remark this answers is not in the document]
ArneGudermann: You are right. But the  `processTypeHint` function should still support `|` in a Typehint.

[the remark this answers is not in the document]
ArneGudermann: Maybe the `processTypeHint` function should provide `|` in the Typehint.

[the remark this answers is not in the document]
sveneberth: Let's discuss this in this issue: https://github.com/viur-framework/viur-core/issues/718

core/modules/user.py
```diff
@@ -946,6 +992,33 @@ def getAuthMethods(self, *args, **kwargs):
 
         return json.dumps(res)
 
+    @exposed
+    def trigger(self, action: str, key: str, skey: str):
+        current.request.get().response.headers["Content-Type"] = "application/json"
+
+        # only "root" users are allowed to trigger maintenance functions!
+        if not ((cuser := current.user.get()) and "root" in cuser["access"]):
```
sveneberth, line 1000: This should be configurable in the adminInfo.

sveneberth (changes requested):

core/modules/user.py
phorward: But how?

sveneberth: I'm not sure about the meaning of access within custom access. Does one of the access roles have to be satisfied or all of them?


But something like this should do it?
```suggestion
        access = self.adminInfo[action]["access"]
        if not ((cuser := current.user.get()) and any(role in cuser["access"] for in access):
```

sveneberth (changes requested):

phorward:
@sveneberth I've implemented your recommendations and merged, updated and tested against current develop.
Can we now merge?

sveneberth (approved):
With the last vi version it works smoothly.

PR #714, opened 2023-04-14
Improve `StringBone`

sveneberth:
* Allow `None` as valid `maxLength` argument (which means unlimited)
* Validate length of input in `isInvalid`
* Don't do `utils.escapeString` in `singleValueFromClient` twice
* Add docstring
* Refactor an ugly if

### DIFF
--- a/core/bones/string.py
+++ b/core/bones/string.py
@@ -1,9 +1,8 @@
 import logging
 from typing import Dict, List, Optional, Set
 
-from viur.core import db, utils, current
+from viur.core import current, db, utils
 from viur.core.bones.base import BaseBone, ReadFromClientError, ReadFromClientErrorSeverity
-
 
 
 class StringBone(BaseBone):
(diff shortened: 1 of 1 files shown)


core/bones/string.py
```diff
@@ -42,11 +50,20 @@ def isEmpty(self, value):
...
+        this bone, an error-message otherwise.
+        """
+        if self.maxLength is not None and len(value) > self.maxLength:
+            return "Maximum length exceeded"
+        return None
+
     def singleValueFromClient(self, value, skel, name, origData):
         value = utils.escapeString(value, self.maxLength)
         err = self.isInvalid(value)
         if not err:
```
ArneGudermann, line 65:
```suggestion
        if not (err := self.isInvalid(value)):
```
We can shortcut this too.

phorward (approved):

ArneGudermann (approved):

PR #725, opened 2023-04-26
Fix: comparision in `SelectBone.singleValueFromClient`

sveneberth:
Compare the value in case of `Enum`-values as string too. The data we get from client are strings, these aren't equal to ints (like in the user status) and wouldn't be valid for this bone.


phorward (changes requested):
Please recreate this PR against main.

sveneberth:
Replaced by https://github.com/viur-framework/viur-core/pull/726

PR #730, opened 2023-04-28
chore: Rename `__systemIsIntitialized_` into `__system_initialized`

phorward:
This renames the misspelled global `__systemIsIntitialized_`  into `__system_initialized_`.

### DIFF
--- a/core/bones/base.py
+++ b/core/bones/base.py
@@ -16,25 +16,25 @@ from viur.core import db
 from viur.core.config import conf
 
 
-__systemIsIntitialized_ = False
+__system_initialized_ = False
 """
-Initializes the global variable __systemIsIntitialized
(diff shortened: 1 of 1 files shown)


core/bones/base.py
```diff
@@ -16,36 +16,36 @@
 from viur.core.config import conf
 
 
-__systemIsIntitialized_ = False
+__system_initialized_ = False
```
sveneberth, line 19:
```suggestion
__system_initialized = False
```

sveneberth (changes requested):

core/bones/base.py
phorward: Ja okay...

sveneberth (approved):

PR #731, opened 2023-04-28
chore: Rename `__undefindedC__` into `__undefined__`

phorward:
Rename the misspelled `__undefindedC__` into `__undefined__`.

### DIFF
--- a/core/skeleton.py
+++ b/core/skeleton.py
@@ -15,7 +15,7 @@ from viur.core.bones import BaseBone, DateBone, KeyBone, RelationalBone, Relatio
 from viur.core.bones.base import ReadFromClientError, ReadFromClientErrorSeverity, getSystemInitialized
 from viur.core.tasks import CallableTask, CallableTaskBase, QueryIter, CallDeferred
 
-__undefined__ = object()
+__undefined = object()
 
 
 class MetaBaseSkel(type):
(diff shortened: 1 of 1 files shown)


core/skeleton.py
```diff
@@ -15,7 +15,7 @@
 from viur.core.bones.base import ReadFromClientError, ReadFromClientErrorSeverity, getSystemInitialized
 from viur.core.tasks import CallableTask, CallableTaskBase, QueryIter, CallDeferred
 
-__undefindedC__ = object()
+__undefined__ = object()
```
sveneberth, line 18: If we're renaming this, can we please call it the common name? (https://github.com/search?l=Python&q=%22sentinel+%3D+object%28%29%22&type=Code)

```suggestion
sentinel = object()
```

sveneberth (changes requested):

core/skeleton.py
phorward: Ok, we can omit the `__*__` notation. But sentinel isn't a better name. It sais nothing. Why not `UndefinedKindName`? This is equal to `None`.

sveneberth: Okay, just `__undefined`? But only with the dunder prefix. _kindname_ should be clear from the context (assignment, comparison).

```suggestion
__undefined = object()
```

phorward: Ok, please re-review.

sveneberth (approved):

ArneGudermann (approved):

PR #732, opened 2023-04-28
chore: Rename all `rawValue`-parameters to just `value`

phorward:
This was inconsistent.

### DIFF
--- a/core/bones/base.py
+++ b/core/bones/base.py
@@ -444,7 +444,7 @@ class BaseBone(object):
     def singleValueFromClient(self, value, skel, name, origData):
         """
         Prevents the BaseBone from reading data using the fromClient method.
-        If needed, use the RawValueBone instead.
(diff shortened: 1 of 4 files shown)


core/bones/base.py
```diff
@@ -444,7 +444,7 @@ def parseSubfieldsFromClient(self) -> bool:
     def singleValueFromClient(self, value, skel, name, origData):
         """
         Prevents the BaseBone from reading data using the fromClient method.
-        If needed, use the rawValueBone instead.
+        If needed, use the RawValueBone instead.
```
sveneberth, line 447: The bone is called just `RawBone`.

```suggestion
        If needed, use the RawBone instead.
```

sveneberth (changes requested):

sveneberth (approved):

PR #735, opened 2023-05-03
fix: enable to serialize complex custom config structures

skoegl:
Fix /vi/config endpoint for complex data structure serialization


ArneGudermann (approved):

phorward (approved):
This is also part of #712, but required earlier now with this PR.

PR #736, opened 2023-05-03
feat: Extend User module to built-in role system

phorward:
This Pull Request drafts a role-system built into the user module as an additional way for user rights management. The roles are provided by the system, and a role definition can be attached for every module.

- [x] Take current state of customer projects
- [x] Rename role "user" into "custom"
- [x] Allow to define multiple roles


phorward:
Pull request is feature-completed.
This pull request resolves #587.

XeoN-GHMB (approved):

PR #737, opened 2023-05-05
fix: Rename `Skeleton.__undefined` into `Skeleton._undefined`

ArneGudermann:
This is a fix for #731 you can't use `__` before a var because this indicates a private class  var in Python. 
You get an error like:
`NameError: name '_Skeleton___undefined' is not defined`
https://docs.python.org/3/tutorial/classes.html#private-variables

### DIFF
--- a/core/skeleton.py
+++ b/core/skeleton.py
@@ -15,7 +15,7 @@ from viur.core.bones import BaseBone, DateBone, KeyBone, RelationalBone, Relatio
 from viur.core.bones.base import ReadFromClientError, ReadFromClientErrorSeverity, getSystemInitialized
 from viur.core.tasks import CallableTask, CallableTaskBase, QueryIter, CallDeferred
 
-__undefined = object()
+_undefined = object()
 
 
 class MetaBaseSkel(type):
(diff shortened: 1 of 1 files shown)


phorward (approved):

phorward:
> This is a fix for #731 you can't use `__` before a var because this indicates a private class var in Python. You get an error like: `NameError: name '_Skeleton___undefined' is not defined` https://docs.python.org/3/tutorial/classes.html#private-variables

Hmm but why did it work with `__undefindedC_` (#731)?
Is it because of the trailing `_`?

sveneberth:
> Hmm but why did it work with `__undefindedC_` (#731)? Is it because of the trailing `_`?

There are two trailing `__` https://github.com/phorward/viur-core/commit/4f39b296c3ee17feaa7df29da2a7a616c22a8942#diff-01f7ba1fdaa209a47d76b4651f82262bfb9ff0063de35b9e8b34e888f0c18e30L18

@ArneGudermann and I assume that Python assumes it's a Python member (like `__dict__`) that can be accessed.

sveneberth (approved):

phorward (approved):

PR #739, opened 2023-05-05
feat: Delete old pending `FileSkeletons`

sveneberth:
Over time, pending FileSkeletons (datastore entities) can accumulate when upload urls have been requested but nothing has been uploaded. Therefore this adds the PeriodicTask `start_delete_pending_files` which deletes pending files older than 90 days.

To do the deletion safely on skeleton level a customization of the `DeleteEntitiesIter` was necessary. I see no reason not to use this for skeletons as long as you call the appropriate `.delete()` method.

### DIFF
--- a/core/modules/file.py
+++ b/core/modules/file.py
@@ -810,7 +810,7 @@ def doCleanupDeletedFiles(cursor=None):
 @PeriodicTask(60 * 4)
 def start_delete_pending_files():
     """
-    Start deletion of pending FileSkels that are older than 90 days.
+    Start deletion of pending FileSkels that are older than 7 days.
     """
     DeleteEntitiesIter.startIterOnQuery(
         FileBaseSkel().all()
--- a/core/modules/file.py
+++ b/core/modules/file.py
@@ -815,5 +815,5 @@ def start_delete_pending_files():
     DeleteEntitiesIter.startIterOnQuery(
         FileBaseSkel().all()
         .filter("pending =", True)
-        .filter("creationdate <", datetime.now() - timedelta(days=90))
+        .filter("creationdate <", datetime.now() - timedelta(days=7))
     )
--- a/core/modules/file.py
+++ b/core/modules/file.py
@@ -1,26 +1,28 @@
 import base64
 import email.header
-import google.auth
+import html
 import json
 import logging
 import string
-import html
-from PIL import Image, ImageCms
 from base64 import urlsafe_b64decode
 from datetime import datetime, timedelta
-from google.auth import compute_engine
-from google.auth.transport import requests
-from google.cloud import iam_credentials_v1, storage
-from google.oauth2.service_account import Credentials as ServiceAccountCredentials
 from io import BytesIO
 from quopri import decodestring
 from typing import Any, List, Tuple, Union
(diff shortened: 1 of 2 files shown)


ArneGudermann:
Why 90 day a Uploadurl is expired after 1 hour (https://github.com/viur-framework/viur-core/blob/a54556277b9e719b50863a2d20771ed8375058de/core/modules/file.py#L467)  .
Can we say 1 day or so ?

sveneberth:
> Why 90 day a Uploadurl is expired after 1 hour

But only for signed URLs from this method (`signUploadURL`), which isn't used anywhere in the core.
URLs created by `getUploadURL` seems to be unlimited?!

Tbh, the 90 doesn't have a specific reason. It should be a time that is enough to process an upload, but not unlimited.
I can't imagine a case where you need a validity of several days. For user uploads, you would actually always pick up the URL immediately before the upload.
So for me, 1 day would also be okay. What do the others (/ping @viur-framework/developers and especially @phorward) think?

phorward:
In the viur meeting, we decided to set this value to 7 days.

core/modules/file.py
```diff
@@ -803,3 +805,15 @@ def doCleanupDeletedFiles(cursor=None):
     newCursor = query.getCursor()
     if newCursor:
         doCleanupDeletedFiles(newCursor)
+
+
+@PeriodicTask(60 * 4)
+def start_delete_pending_files():
+    """
+    Start deletion of pending FileSkels that are older than 90 days.
```
sveneberth, line 813:
```suggestion
    Start deletion of pending FileSkels that are older than 7 days.
```

core/modules/file.py
```diff
@@ -803,3 +805,15 @@ def doCleanupDeletedFiles(cursor=None):
...
+
+@PeriodicTask(60 * 4)
+def start_delete_pending_files():
+    """
+    Start deletion of pending FileSkels that are older than 7 days.
+    """
+    DeleteEntitiesIter.startIterOnQuery(
+        FileBaseSkel().all()
+        .filter("pending =", True)
+        .filter("creationdate <", datetime.now() - timedelta(days=7))
```
phorward, line 818: I'm not sure, but shall be use `utcNow()` here, as `creationdate` is a timezone-aware value?
```suggestion
        .filter("creationdate <", utils.utcNow() - timedelta(days=7))
```

core/tasks.py
```diff
@@ -746,7 +746,7 @@ def _qryStep(cls, qryDict: Dict[str, Any]) -> None:
                         logging.exception(e)
                         doCont = False
                     if not doCont:
-                        logging.error("Exiting queryItor on cursor %s" % qry.getCursor())
+                        logging.error("Exiting queryIter on cursor %s" % qry.getCursor())
```
phorward, line 749: Please use f-strings!
```suggestion
                        logging.error(f"Exiting queryIter on cursor {qry.getCursor()!r}")
```

phorward (changes requested):
Hello @sveneberth, looks OK to me, can you review the two comments?

phorward (approved):

PR #742, opened 2023-05-08
feat:add to_json jinjaFunction in html renderer

ArneGudermann:
This PR add a `to_json` Jinja function to  the HTML renderer. This is needed in the future for the `renderBone` Templates. 
Like so:
```jinja
<sl-bone  bonestructure='{{to_json(skel["structure"][bonename])}}' ></sl-bone>
```
We cant use the `tojson` build in Filter from Jinja to handle this. It can't convert Key objects and so on.

### DIFF
--- a/core/render/html/env/viur.py
+++ b/core/render/html/env/viur.py
@@ -1,19 +1,19 @@
-from collections import OrderedDict
-
+import json
 import logging
 import os
 import string
 import urllib
 import urllib.parse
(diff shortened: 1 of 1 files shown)


sveneberth:
> We cant use the `tojson` build in Filter from Jinja to handle this

We can. Take a look at https://github.com/viur-framework/viur-core/pull/744

PR #743, opened 2023-05-10
fix: `File.getUploadURL` with HttpExceptions

### LINKED ISSUE #727: Invalid Mime-type? Cannot upload customers files with new core
The file I was trying to upload was downloaded from a viur2 production system, it is an *.oft file. I don't know what the correct mimetype should be here, but I doubt that we are doing everything correctly when checking. Why do we check the mimetype at all? Do we reject any type of file or is this just for the mimetype field?

```
[2023-04-27 13:44:13,608] /.../viur/core/request.py:299 [ERROR] ViUR has caught an unhandled exception!
[2023-04-27 13:44:13,608] /.../viur/core/request.py:300 [ERROR] Invalid Mime-Type
  File "/home/user/.local/share/virtualenvs/project-viur3-5lETOfwa/lib/python3.10/site-packages/viur/core/request.py", line 275, in processRequest
    self.findAndCall(path)
  File "/home/user/.local/share/virtualenvs/project-viur3-5lETOfwa/lib/python3.10/site-packages/viur/core/request.py", line 623, in findAndCall
    res = caller(*newArgs, **newKwargs)
  File "/home/user/.local/share/virtualenvs/project-viur3-5lETOfwa/lib/python3.10/site-packages/viur/core/modules/file.py", line 520, in getUploadURL
    assert len(mimeType.split("/")) == 2, "Invalid Mime-Type"
AssertionError: Invalid Mime-Type
[2023-04-27 13:44:13] [500] [POST] /json/file/getUploadURL
```

phorward:
...instead of AssertionError. Also includes some refactoring of the code here.

Proposal to resolve #727.

### DIFF
--- a/core/modules/file.py
+++ b/core/modules/file.py
@@ -518,7 +518,7 @@ class File(Tree):
         authSig = kwargs.get("authSig")
 
         # Validate the contentType from the client seems legit
-        mimeType = mimeType.lower()
+        mimeType = mimeType.strip().lower()
         if not (
             mimeType
             and mimeType.count("/") == 1
--- a/core/modules/file.py
+++ b/core/modules/file.py
@@ -520,7 +520,8 @@ class File(Tree):
         # Validate the contentType from the client seems legit
         mimeType = mimeType.lower()
         if not (
-            len(mimeType.split("/")) == 2
+            mimeType
+            and mimeType.count("/") == 1
             and all([ch in string.ascii_letters + string.digits + "/-.+" for ch in mimeType])
         ):
             raise errors.NotAcceptable("Invalid mimeType provided")
--- a/core/modules/file.py
+++ b/core/modules/file.py
@@ -524,7 +524,7 @@ class File(Tree):
             and mimeType.count("/") == 1
             and all(ch in string.ascii_letters + string.digits + "/-.+" for ch in mimeType)
         ):
-            raise errors.UnprocessableEntity(f"Invalid mimeType {mimeType} provided")
+            raise errors.UnprocessableEntity(f"Invalid mime-type {mimeType} provided")
 
         if authData and authSig:
             # First, validate the signature, otherwise we don't need to proceed further
@@ -542,7 +542,7 @@ class File(Tree):
                         validMimeType.endswith("*") and mimeType.startswith(validMimeType[:-1])):
                         break
                 else:
-                    raise errors.UnprocessableEntity(f"Invalid mimeType {mimeType} provided")
+                    raise errors.UnprocessableEntity(f"Invalid mime-type {mimeType} provided")
 
             node = authData["node"]
             maxSize = authData["maxSize"]
--- a/core/modules/file.py
+++ b/core/modules/file.py
@@ -512,21 +512,29 @@ class File(Tree):
         return db.encodeKey(fileSkel["key"]), uploadUrl
 
     @exposed
-    def getUploadURL(self, fileName, mimeType, size=None, skey=None, *args, **kwargs):
+    def getUploadURL(self, fileName: str, mimeType: str, size: int = None, skey: str = None, *args, **kwargs):
         node = kwargs.get("node")
         authData = kwargs.get("authData")
         authSig = kwargs.get("authSig")
-        # Validate the the contentType from the client seems legit
+
+        # Validate the contentType from the client seems legit
         mimeType = mimeType.lower()
-        assert len(mimeType.split("/")) == 2, "Invalid Mime-Type"
-        assert all([x in string.ascii_letters + string.digits + "/-.+" for x in mimeType]), "Invalid Mime-Type"
+        if not (
+            len(mimeType.split("/")) == 2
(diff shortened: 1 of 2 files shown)


core/modules/file.py
```diff
@@ -512,30 +512,41 @@ def initializeUpload(self,
...
-        assert all([x in string.ascii_letters + string.digits + "/-.+" for x in mimeType]), "Invalid Mime-Type"
+
+        # Validate the contentType from the client seems legit
+        mimeType = mimeType.strip().lower()
+        if not (
+            mimeType
+            and mimeType.count("/") == 1
+            and all([ch in string.ascii_letters + string.digits + "/-.+" for ch in mimeType])
+        ):
+            raise errors.NotAcceptable("Invalid mimeType provided")
```
sveneberth, line 527:
```suggestion
            raise errors.NotAcceptable(f"Invalid mimeType {mimeType} provided")
```

core/modules/file.py
```diff
@@ -512,30 +512,41 @@ def initializeUpload(self,
...
-        mimeType = mimeType.lower()
-        assert len(mimeType.split("/")) == 2, "Invalid Mime-Type"
-        assert all([x in string.ascii_letters + string.digits + "/-.+" for x in mimeType]), "Invalid Mime-Type"
+
+        # Validate the contentType from the client seems legit
+        mimeType = mimeType.strip().lower()
+        if not (
+            mimeType
+            and mimeType.count("/") == 1
+            and all([ch in string.ascii_letters + string.digits + "/-.+" for ch in mimeType])
```
sveneberth, line 525:
```suggestion
            and all(ch in string.ascii_letters + string.digits + "/-.+" for ch in mimeType)
```

core/modules/file.py
sveneberth, line 527:
```suggestion
            raise errors.NotAcceptable(f"Invalid mimeType {mimeType} provided")
```

sveneberth (changes requested):
Additionally to my suggestions `406 Not Acceptable` is in the entire method the wrong HTTP-Status. It means the mimetype of the request itself, not of the content.

core/modules/file.py
phorward: Didn't know this is valid, but it seems to work...

core/modules/file.py
```diff
@@ -544,25 +555,25 @@ def getUploadURL(self, fileName, mimeType, size=None, skey=None, *args, **kwargs
...
             maxSize = None  # The user has some file/add permissions, don't restrict fileSize
+
         if maxSize:
-            try:
-                size = int(size)
-                assert size <= maxSize
-            except:  # We have a size-limit set - but no size supplied
-                raise errors.PreconditionFailed()
+            if size > maxSize:
+                raise errors.Forbidden("Size limit exceeds maximum size")
```
ArneGudermann, line 563: Can this cause an error if no size is specified?

sveneberth (changes requested):
> Additionally to my suggestions `406 Not Acceptable` is in the entire method the wrong HTTP-Status. It means the mimetype of the request itself, not of the content.

Please change this as well

core/modules/file.py
xnopasaranx: I am not entirely sure, but it seems to me that size is required. So if not specified, it should cause an error. Size needs to be exact according to initializeUpload doc str. It gets passed to the blob API in order to create a resumable upload session.

xnopasaranx:
The correct error status to throw for an invalid mimetype would be 415: Unsupported Media Type

core/modules/file.py
sveneberth: > Can this cause an error if no size is specified?

Yup, if the size isn't provides as kwarg, it's None. Even with the type hint. 
Therefore the case `size is None` is missing.

ArneGudermann: Maybe we ca use something like
```suggestion
            try:
                size = int(size)
            except:
                raise errors.NotAcceptable(f"Invalid size {size} provided")
            if size > maxSize:
                raise errors.Forbidden("Size limit exceeds maximum size")
``` 
I don't know a better solution like `errors.NotAcceptable`

sveneberth (changes requested):
https://github.com/phorward/viur-core/pull/4

core/modules/file.py
phorward: Ok, why not turn the size-parameter by default to 0 instead of None? As it's type-hinted with `int` it should be rejected in case it doesn't contain a valid int.

sveneberth: > Ok, why not turn the size-parameter by default to 0 instead of None? As it's type-hinted with `int` it should be rejected in case it doesn't contain a valid int.

I would prefer `-1`. An empty file (`size = 0`) would be theoretically possible. Even though it probably makes no sense. But a negative value clearly indicates a different meaning.

phorward:
> [phorward#4](https://github.com/phorward/viur-core/pull/4)

Thanks, I've merged this.

sveneberth (approved):

sveneberth (approved):

PR #749, opened 2023-05-16
fix: `DateBone.fromClient()` should regard tzinfo

phorward:
Hi there,

I came across this error with a project I'm porting from version 3.3.5 to 3.4. The `DateBone.fromClient()` function is called with a ISO-string `2023-02-20T12:26:01+01:00` and fails, because the timezone is already provided. 

Furthermore, the case with `now` was not part of the localization done below. The behavior is now corrected, the bug was introduced by #667.

### DIFF
--- a/core/bones/date.py
+++ b/core/bones/date.py
@@ -112,8 +112,6 @@ class DateBone(BaseBone):
         elif not self.date and self.time:
             try:
                 value = datetime.fromisoformat(value)
-                if not self.naive:
-                    value = time_zone.localize(value)
             except:
                 try:
                     if str(rawValue).count(":") > 1:
--- a/core/bones/date.py
+++ b/core/bones/date.py
@@ -165,11 +165,12 @@ class DateBone(BaseBone):
                 except:
                     value = False  # its invalid
 
-        if value is False:
+        if not value:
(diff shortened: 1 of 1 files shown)


XeoN-GHMB (dismissed):

core/bones/date.py
```diff
@@ -166,15 +164,20 @@ def singleValueFromClient(self, value: str, skel: 'viur.core.skeleton.SkeletonIn
...
-            if value and not self.naive:
-                value = time_zone.localize(value)
-        if value is False:
+
+        if not value:
             return self.getEmptyValue(), [
-                ReadFromClientError(ReadFromClientErrorSeverity.Invalid, "Invalid value entered")]
+                ReadFromClientError(ReadFromClientErrorSeverity.Invalid, "Invalid value entered")
+            ]
+
```
sveneberth, line 172: As already talked about: This case was missing anyway

```suggestion
        if value.tzinfo and self.naive:
            return self.getEmptyValue(), [
                ReadFromClientError(ReadFromClientErrorSeverity.Invalid, "Datetime must be naive")
            ]
```

sveneberth (changes requested):

Issue #760, opened 2023-06-01
PasswordBone - die Übersetzung von "The password is too short. It requires for at least 8 characters." fehlt


ClaudiaSollinger:
PasswordBone - Check des Passwortes

Übersetzung von "The password is too short. It requires for at least 8 characters." fehlt

sveneberth:
Hello @ClaudiaSollinger, can you please explain this problem in more detail?

You mean this translate string, right?
https://github.com/viur-framework/viur-core/blob/b9aedcb1ae7207e8b614c2f7882ebc82890381f1/core/bones/password.py#L53

Like any other translation in the core, there's only the `defaultText` set. If you need it in English or any other language, you have to add the translation to your project.

PR #763, opened 2023-06-07
fix: Missing german translation for "password too short" message

ClaudiaSollinger:
The German translation of one "error"-Prompt for passwordBones was missing. Added it


sveneberth (approved):

phorward:
fixes #760

PR #764, opened 2023-06-12
feat: Rewrite of session-based securitykeys

phorward:
This is a draft for solving the problem of one session-based securitykey, making multiple requests in parallel impossible. This pull request partly replaces #704.

**Features:**
- a new securitykey is created every time it is requested
- securitykeys are bound to a session, if `session`-flag is True (default)
- unused securitykeys from a session are either being dropped on session kill, or when they became invalid
- Entire code refactoring for both securitykey and session module
- `/json/skey' accepts `amount`-flag for batch-mode skey requesting (decreases amount of requests)

### DIFF
--- a/core/render/json/__init__.py
+++ b/core/render/json/__init__.py
@@ -2,13 +2,12 @@ from .default import DefaultRender as default
 from .user import UserRender as user
 from viur.core import securitykey, current, errors, exposed
 import json
-import typing
 
 __all__ = [default]
 
 
 @exposed
-def skey(duration: typing.Optional[int] = None, *args, **kwargs) -> str:
+def skey(duration: int = 60, *args, **kwargs) -> str:
     """
     Creates or returns a valid skey.
 
@@ -20,12 +19,8 @@ def skey(duration: typing.Optional[int] = None, *args, **kwargs) -> str:
     """
     current.request.get().response.headers["Content-Type"] = "application/json"
 
-    if duration is not None:
-        if not current.user.get():
-            raise errors.Unauthorized("Durations can only be used by authenticated users")
-
-        if not 0 < duration <= 60:
-            raise errors.Forbidden("Invalid duration provided")
+    if not 0 < duration <= 60:
+        raise errors.Forbidden("Invalid duration provided")
 
     return json.dumps(securitykey.create(duration=duration))
 
--- a/core/render/json/__init__.py
+++ b/core/render/json/__init__.py
@@ -7,22 +7,30 @@ __all__ = [default]
 
 
 @exposed
-def skey(duration: int = 60, *args, **kwargs) -> str:
+def skey(amount: int = 1, *args, **kwargs) -> str:
     """
-    Creates or returns a valid skey.
+    Creates CSRF-security-keys for transactions.
 
-    When a user is authenticated, a duration can be provided,
-    which returns a fresh, session-agnostic skey.
-    Otherwise, a session-based skey is returned.
+    All returned keys are associated with the session, therefore they cannot be used across sessions.
+    The keys get a maximum lifetime of the session lifetime, afterward they become invalid.
+
+    :param amount: Optional amount of securitykeys to create in a batch.
+        `amount > 1` can only be used by authenticated users, for a maximum of 100 keys.
 
     See module securitykey for details.
     """
     current.request.get().response.headers["Content-Type"] = "application/json"
 
-    if not 0 < duration <= 60:
+    if amount == 1:
+        return json.dumps(securitykey.create())
+
+    if not 0 < amount <= 100:
         raise errors.Forbidden("Invalid duration provided")
 
-    return json.dumps(securitykey.create(duration=duration))
+    if not current.user.get():
+        raise errors.Forbidden("Batch securitykey creation is only available to authenticated users")
+
+    return json.dumps([securitykey.create() for _ in range(amount)])
 
 
 def _postProcessAppObj(obj):  # Register our SKey function
--- a/core/render/json/__init__.py
+++ b/core/render/json/__init__.py
@@ -25,7 +25,7 @@ def skey(amount: int = 1, *args, **kwargs) -> str:
         return json.dumps(securitykey.create())
 
     if not 0 < amount <= 100:
-        raise errors.Forbidden("Invalid duration provided")
+        raise errors.Forbidden("Invalid amount provided")
 
     if not current.user.get():
         raise errors.Forbidden("Batch securitykey creation is only available to authenticated users")
--- a/core/securitykey.py
+++ b/core/securitykey.py
@@ -1,27 +1,18 @@
 """
-    This module provides one-time keys.
+    Implementation of one-time CSRF-security-keys.
 
-    There are two types of security keys:
-
-    1. If :meth:create is called without arguments, it returns the current session CSRF token. Repeated calls to
-        :meth:create will return the same CSRF token (for the same session) until that token has been redeemed.
-        This security key will be valid as long the session is active, and it's not possible to store data along with
-        that key. These are usually used as a CSRF token.
-        This has been changed from ViUR2 - where it was possible to create a arbitrary number of security keys per
-        session.
-    2. If :meth:create is called with a duration (and optional kwargs values), it will create a security key
-        that is *not* bound to the current session, but it's possible to store custom data (provided by kwargs).
-        As these are not bound to the session, each call to :meth:create will yield a new token.
-        These are used if it's expected that the token may be redeemed on a different device (e.g. when sending an
-        email address confirmation)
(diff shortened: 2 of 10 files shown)


core/session.py
```diff
@@ -69,18 +65,17 @@ def load(self, req: BrowseHandler):
             If the client supplied a valid Cookie, the session is read from the datastore, otherwise a new,
             empty session will be initialized.
         """
-        if self.cookieName in req.request.cookies:
-            cookie = str(req.request.cookies[self.cookieName])
+        if self.cookie_name in req.request.cookies:
+            cookie = str(req.request.cookies[self.cookie_name])
```
ArneGudermann, line 69:
```suggestion
        if self.cookie_key := str(req.request.cookies.get(self.cookie_name)):
```
I think we can simplify this.

core/session.py
```diff
@@ -69,18 +65,17 @@ def load(self, req: BrowseHandler):
...
                     # This session is too old
                     self.reset()
                     return False
 
                 self.session = data["data"]
-                self.staticSecurityKey = data["staticSecurityKey"]
-                self.securityKey = data["securityKey"]
-                self.cookieKey = cookie
+                self.static_security_key = data["static_security_key"]
+                self.cookie_key = cookie
```
ArneGudermann, line 78:
```suggestion
```
This is obsolete now.

core/session.py
```diff
@@ -69,18 +65,17 @@ def load(self, req: BrowseHandler):
             If the client supplied a valid Cookie, the session is read from the datastore, otherwise a new,
             empty session will be initialized.
         """
-        if self.cookieName in req.request.cookies:
-            cookie = str(req.request.cookies[self.cookieName])
+        if self.cookie_name in req.request.cookies:
+            cookie = str(req.request.cookies[self.cookie_name])
             if data := db.Get(db.Key(self.kindName, cookie)):  # Loaded successfully
                 if data["lastseen"] < time.time() - conf["viur.session.lifeTime"]:
```
ArneGudermann, line 71:
```suggestion
            if data := db.Get(db.Key(self.kindName, self.cookie_key)):  # Loaded successfully
                if data["lastseen"] < time.time() - conf["viur.session.lifeTime"]:
```

ArneGudermann (changes requested):
Hey @phorward thank you  for this PR. I have a littel suggetion. Can you please take a look.

core/session.py
```diff
@@ -199,56 +192,27 @@ def reset(self) -> None:
...
+            db.Delete(db.Key(self.kindName, self.cookie_key))
+            from viur.core import securitykey
+            securitykey.clear_session_skeys(self.cookie_key)
 
-        self.cookieKey = utils.generateRandomString(42)
-        self.staticSecurityKey = utils.generateRandomString(13)
-        self.securityKey = utils.generateRandomString(13)
+        self.cookie_key = utils.generateRandomString(42)
+        self.static_security_key = utils.generateRandomString(13)
+        self.session.clear()
```
sveneberth, line 202: Since `viur.datastore.Entity` does not have a custom `clear` method, I would also recommend doing to reset the key to make sure the new session gets a new key. Maybe there are some tasks or instances that hold references to the old session. Then I think (I haven't invested time in researching this) it might be possible to take over the new session incorrectly.

```py
self.session.key = None
```

core/securitykey.py
```diff
@@ -26,88 +26,99 @@
...
-    entity["until"] = utils.utcNow() + timedelta(seconds=int(duration))
+    entity["viur_session"] = current.session.get().cookie_key if session else None
+    entity["viur_until"] = utils.utcNow() + datetime.timedelta(seconds=int(duration))
     db.Put(entity)
 
     return key
 
 
-def validate(key: str, useSessionKey: bool) -> Union[bool, db.Entity]:
+def validate(key: str, session: bool = True) -> typing.Union[bool, db.Entity]:
```
sveneberth, line 69: I would like to have a more expressive name in the sense of "enforce that skey must be bound to the current session" or "only accept skeys from this session", but I have no really good suggestion for a short parameter name :(




```suggestion
def validate(key: str, enforce_session_bounded: bool = True) -> typing.Union[bool, db.Entity]:
```

sveneberth (changes requested):

core/render/json/__init__.py
```diff
@@ -20,12 +19,8 @@ def skey(duration: typing.Optional[int] = None, *args, **kwargs) -> str:
     """
     current.request.get().response.headers["Content-Type"] = "application/json"
 
-    if duration is not None:
-        if not current.user.get():
-            raise errors.Unauthorized("Durations can only be used by authenticated users")
-
-        if not 0 < duration <= 60:
-            raise errors.Forbidden("Invalid duration provided")
+    if not 0 < duration <= 60:
```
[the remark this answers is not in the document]
phorward: Yeah, well I'm not sure. We should discuss this in the viur-meeting this friday.

core/securitykey.py
phorward: I've first named this `session_bounded`. We should discuss this on Friday as well.

core/securitykey.py
```diff
@@ -26,88 +26,99 @@
...
     """
-    if useSessionKey:
-        session = current.session.get()
-        if key == "staticSessionKey":
-            skey_header_value = current.request.get().request.headers.get("Sec-X-ViUR-StaticSKey")
-            if skey_header_value and session.validateStaticSecurityKey(skey_header_value):
-                return True
-
-        elif session.validateSecurityKey(key):
+    if session and key == "staticSessionKey":
```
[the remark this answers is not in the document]
phorward: 0bfc59ecaebcee4c22abff6c7587024c18bd3c58 cleans up the use of the static session key and its naming.

core/session.py
```diff
@@ -30,37 +30,33 @@ class Session:
...
             all data currently stored to be lost. Only keys listed in these variables will be copied into the new
             session.
-
     """
     kindName = "viur-session"
-    sameSite = "lax"  # Either None (don't issue sameSite header), "none", "lax" or "strict"
-    sessionCookie = True  # If True, issue the cookie without a lifeTime (will disappear on browser close)
-    cookieName = f"""viur_cookie_{conf["viur.instance.project_id"]}"""
+    same_site = "lax"  # Either None (don't issue same_site header), "none", "lax" or "strict"
+    is_session_cookie = True  # If True, issue the cookie without a lifeTime (will disappear on browser close)
```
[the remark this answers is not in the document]
phorward: 0400c1e0c96f5e98efd3ab8495ed58161802b764 renames the flag.

core/session.py
phorward: Well, the `session` property of the `Session` class is awkward, as it is a `db.Entity` which is only held in memory to change session properties and which is written into the `data` property of a `db.Entity` named `dbSession` in `Session.save()`. Therefore the key of `session.key` is never set to anything, and therefore is not required to be set to `None`. I don't know why this was solved this way, maybe it's only bad code that simply can be replaced by a dict.

core/render/json/__init__.py
```diff
@@ -2,32 +2,35 @@
...
     current.request.get().response.headers["Content-Type"] = "application/json"
 
-    if duration is not None:
-        if not current.user.get():
-            raise errors.Unauthorized("Durations can only be used by authenticated users")
+    if amount == 1:
+        return json.dumps(securitykey.create())
+
+    if not 0 < amount <= 100:
+        raise errors.Forbidden("Invalid duration provided")
```
phorward, line 28:
```suggestion
        raise errors.Forbidden("Invalid amount provided")
```

core/securitykey.py
phorward: It will be renamed to `session_bounded`.

phorward: [3340b47](https://github.com/viur-framework/viur-core/pull/764/commits/3340b474e64c07dc1ab7e5e686e48229b1c1ac03) renames the parameter into `session_bound`.

core/render/json/__init__.py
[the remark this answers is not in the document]
phorward: Meanwhile this feature has been dropped from the skey-retrieval function.

core/securitykey.py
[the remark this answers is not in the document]
sveneberth: But things got mixed up here.

`SECURITYKEY_STATIC` now contains the header name. The static key for the session skey, which was previously `"staticSessionKey"`, is now also the header name `"Sec-X-ViUR-StaticSessionKey"`. This is btw. a breaking change...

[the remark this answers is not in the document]
phorward: yeah, and who uses it?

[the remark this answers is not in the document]
sveneberth: I've created #796

PR #785, opened 2023-07-12
feat: Compute `creationdate` and `changedate` with the new `compute`-feature

ArneGudermann:
This PR change the calculation of the `creationdate` and  `changedate` from `performMagic` to the newer `compute`.
I the future we can discuss to remove the `performMagic` function completly.

### DIFF
--- a/core/skeleton.py
+++ b/core/skeleton.py
@@ -13,6 +13,7 @@ from typing import Any, Callable, Dict, Iterable, List, Optional, Set, Tuple, Ty
 from viur.core import conf, db, email, errors, utils, current
 from viur.core.bones import BaseBone, DateBone, KeyBone, RelationalBone, RelationalUpdateLevel, SelectBone, StringBone
 from viur.core.bones.base import ReadFromClientError, ReadFromClientErrorSeverity, getSystemInitialized
+from viur.core.bones.base import Compute, ComputeMethod, ComputeInterval
 from viur.core.tasks import CallableTask, CallableTaskBase, QueryIter, CallDeferred
 
 _undefined = object()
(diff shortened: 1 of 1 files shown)


phorward (dismissed):
Looks good, did you test it?

ArneGudermann:
> Looks good, did you test it?

Yes

ArneGudermann:
Now it works.

sveneberth:
@ArneGudermann Can you explain why we need to throw away the microseconds here? Doesn't the `viur-datastore` already do that?

ArneGudermann:
> @ArneGudermann Can you explain why we need to throw away the microseconds here? Doesn't the `viur-datastore` already do that?

Yes you're right. I cahnge it. You mean this ?
https://github.com/viur-framework/viur-datastore/blob/37080a63bb39c0db2ec5298c698e692989b85b26/src/viur/datastore/transport.pyx#L188-L190

sveneberth:
> > @ArneGudermann Can you explain why we need to throw away the microseconds here? Doesn't the `viur-datastore` already do that?
> 
> You mean this ? https://github.com/viur-framework/viur-datastore/blob/37080a63bb39c0db2ec5298c698e692989b85b26/src/viur/datastore/transport.pyx#L188-L190

Okay. I was wrong. The datastore client doesn't throw it away, but the core in the `FileBone` does:

https://github.com/viur-framework/viur-core/blob/19f1351ffd5e9fc4269235289b78b48fa2481f71/core/bones/date.py#L242

ArneGudermann:
> Okay. I was wrong. The datastore client doesn't throw it away, but the core in the `FileBone` does:
> 
> https://github.com/viur-framework/viur-core/blob/19f1351ffd5e9fc4269235289b78b48fa2481f71/core/bones/date.py#L242
> 
> So I don't understand why we have to do this here again.

 `FileBone`?
I dont now why throw it away. I implement the old way.

phorward:
Did this PR depend on #786 ?

phorward:
> > Okay. I was wrong. The datastore client doesn't throw it away, but the core in the `FileBone` does:
> > https://github.com/viur-framework/viur-core/blob/19f1351ffd5e9fc4269235289b78b48fa2481f71/core/bones/date.py#L242
> > 
> > So I don't understand why we have to do this here again.
> 
> `FileBone`? I dont now why throw it away. I implement the old way. May be it's historical. The datastore support microseconds in the timestamp. https://cloud.google.com/datastore/docs/reference/data/rest/Shared.Types/Value

The reason for setting the microseconds to 0 is, that additional, project-specific program code might rely on a datetime-object which is the exact time to seconds, without microseconds. In the past, I had a use-case where this was a big problem and why it was added. microseconds should only be set when the DateBone is in a yet not existing time-with-microseconds-mode.

I would like to take this PR for now as is. We can later change the behavior to respect microseconds when this is configured so.

phorward (approved):

sveneberth (approved):

PR #793, opened 2023-07-19
fix: Add missing fallback for `NumericBone.refresh()` destroying valid data

ArneGudermann:
This PR fix the problem when a `numericBone` is refreshed the value would be overwritten with `None`. If the value is already a `int/float`.


xnopasaranx (approved):
missing return statement makes for an easy PR approval.

phorward (approved):

PR #798, opened 2023-07-21
Make recipients for `sendEMailToAdmins` configurable

sveneberth:
It is now possible to set recipients for admin emails via config `conf["viur.email.admin_recipients"]`. Otherwise all root users are still addressed. By using `normalize_to_list` and its enhancement, a method as value is also possible. This allows e.g. dynamic recipient selections like own `UserSkel` queries. Also `conf["viur.email.recipientOverride"]` is now callable.

### DIFF
--- a/core/email.py
+++ b/core/email.py
@@ -2,7 +2,7 @@ import logging, os
 import json, base64
 from urllib import request
 from abc import ABC, abstractmethod
-from typing import Any, Union, List, Dict
+from typing import Any, Callable, Union, List, Dict
 from viur.core.config import conf
 from viur.core import utils, db
 from viur.core.tasks import CallDeferred, QueryIter, PeriodicTask, DeleteEntitiesIter
@@ -119,10 +119,14 @@ def sendEmailDeferred(emailKey: db.Key):
         logging.exception(e)
 
 
-def normalizeToList(value: Union[None, Any, List[Any]]) -> List[Any]:
+def normalize_to_list(value: Union[None, Any, List[Any], Callable[[], List]]) -> List[Any]:
     """
-        Convert the given value to a list.
+    Convert the given value to a list.
+
+    If the value parameter is callable, it will be called first to get the actual value.
     """
+    if callable(value):
+        value = value()
     if value is None:
         return []
     if isinstance(value, list):
@@ -141,7 +145,7 @@ def sendEMail(*,
               headers: Dict[str, str] = None,
               attachments: List[Dict[str, Any]] = None,
               context: Union[db.DATASTORE_BASE_TYPES, List[db.DATASTORE_BASE_TYPES], db.Entity] = None,
-              **kwargs) -> Any:
+              **kwargs) -> bool:
     """
     General purpose function for sending e-mail.
     This function allows for sending e-mails, also with generated content using the Jinja2 template engine.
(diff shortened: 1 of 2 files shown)


XeoN-GHMB (approved):

core/email.py
```diff
@@ -141,7 +145,7 @@ def sendEMail(*,
               headers: Dict[str, str] = None,
```
phorward, line 145: `dests`, `cc` and `bcc` now can be a `Callable[[], List]]` as well?

core/email.py
```diff
@@ -141,7 +145,7 @@ def sendEMail(*,
...
               attachments: List[Dict[str, Any]] = None,
```
phorward, line 146: `attachments` now can be a `Callable[[], List]]` as well?

core/email.py
```diff
@@ -246,29 +250,31 @@ def sendEMail(*,
...
-                success = True
-                return ret
-            else:
-                logging.warning("There are no root-users.")
-
-    except Exception:
-        raise
+        users = []
+        if conf["viur.email.admin_recipients"] is not None:
+            users = normalize_to_list(conf["viur.email.admin_recipients"])
```
phorward, line 267: This is already done by sendEMail
```suggestion
            users = conf["viur.email.admin_recipients"]
```

core/email.py
```diff
@@ -246,29 +250,31 @@ def sendEMail(*,
...
+        elif "user" in dir(conf["viur.mainApp"]):
+            for user_skel in conf["viur.mainApp"].user.viewSkel().all().filter("access =", "root").fetch():
+                users.append(user_skel["name"])
+
+        if users:
+            ret = sendEMail(dests=users, stringTemplate=os.linesep.join((subject, body)), *args, **kwargs)
+            success = True
```
phorward, line 274: Hmm what the heck is the purpose of this variable?
```suggestion
```
why isn't it just the Exception part?

phorward (changes requested):
Hello @sveneberth,
can you please apply my advices regarding typing and the general try...except..finally structure.

core/email.py
sveneberth: Because the case `success = False` can occur also without an exception (no users available). To unify this there is this variable. And as this has to be checked after both cases, this is in the `finally` block.

core/email.py
sveneberth: Yes, this is a side effect of the extension of the `normalize_to_list` method. 
But I didn't want to over-engineer this and make the callable option parameter-controlled. 
However, since I don't see the point in passing attachments as a method, I didn't document it further, but just left it as an "Easter-Egg".

For me we can extend the method signature, but for me it would be ok to leave it as it is, because this should not be the primary data type.

core/email.py
sveneberth: Yes, the same known and accepted side effect.

core/email.py
sveneberth: If `users` is a `callable`, it will already be called at this point. If the result is empty, this method can fail-fast in https://github.com/viur-framework/viur-core/blob/72277865f4e5b0ab035b94acca2c76f38ef0a5db/core/email.py#L272

Also, if the callable `users` returns a string, it will be converted to a list in the `normalize_to_list` call inside `sendEMail`.

So I would keep it to be most flexible.

phorward (approved):
Thanks for clarification.

PR #800, opened 2023-07-27
feat: Implement a new `Module`/`Method` concept with new decorators

### LINKED ISSUE #715: Extend `@exposed` decorator to simplify and generalize recurring code checks
This issue is a replacement for #618 to collect several scenarios and use-cases where an extended `@exposed`-decorator can be useful to configure a function's behavior, resulting in secure, less error prone and shorter code.

## `restrict`

To avoid function definitions like this:
```python
    @exposed
    def do_something(self, action: str):
        if not ((cuser := current.user.get()) and "root" in cuser["access"]):
            raise errors.Unauthorized()

        # ...
```
a `restrict`-attribute can be used like this:
```python
    @exposed(restrict="root")
    def do_something(self, action: str):
        # ...
```
or when thinking further, using a lambda:
```python
    @exposed(restrict=lambda user: "root" in user["access"] or "admin" in user["access"])
    def do_something(self, action: str):
        # ...
```

## `response`

To avoid function definitions like this:
```python
    @exposed
    def do_something(self, action: str):

        match action:
            # ...

            case _:
                raise errors.NotImplemented(f"Action {action!r} not implemented")

        return json.dumps("OKAY")
```
a `response`-attribute could be used like this.
```python
    @exposed(response="json")
    def do_something(self, action: str):
        match action:
            # ...

            case _:
                raise errors.NotImplemented(f"Action {action!r} not implemented")

        return "OKAY"
```

## `skey`

To avoid function definitions like this:
```python
    @exposed
    def do_something(self, action: str, skey: str):
        if not securitykey.validate(skey):
            raise errors.PreconditionFailed()
        # ...
```
an `skey`-attribute can be used,
```python
    @exposed(skey=True)
    def do_something(self, action: str):
        # ...
```
optionally with an lambda, evaluating the result of the `securitykey.validate()` function
```python
    @exposed(skey=lambda skey: skey["user"] == current.user.get()["key"])
    def do_something(self, action: str):
        # ...
```

phorward:
Hello!

This pull request is an alternative implementation based on #782, but wrapping any exposed function in a special `Exposed`-class, which holds the routing configuration. It is used to conceptually resolve #715 in a better and less hackish implementation, as any function that can be routed is now an `Exposed`, and must not be checked against any attached attributes set to it.

This PR is a draft, it still contains unfinished parts and shall be 100% backward compatible.


core/module.py
```diff
@@ -152,6 +152,53 @@ def describe(self) -> Union[Dict, None]:
...
+
+                    if parameter.default is not inspect.Parameter.empty:
+                        _params[parameter.name]["default"] = str(parameter.default)
+
+                return_hint = annotations.get("return")
+                attributes |= {
+                    "name": f.__name__,
+                    "docs": f.__doc__,
+                    "args": _params,
+                    "return": "-" if return_hint is None else str(return_hint)
```
theVAX, line 193: "-" could be interpreted as True. Shouldn't the client reading the data then choose output appropriate for the usecase

src/viur/core/request.py
```diff
@@ -491,8 +491,8 @@ def test(a: List[str])
 
     def findAndCall(self, path: str, *args, **kwargs) -> None:
         """
-            Does the actual work of sanitizing the parameter, determine which @exposed (or @internalExposed) function
-            to call (and with witch parameters)
+            Does the actual work of sanitizing the parameter, determine which exposed-function to call
+            (and with which parameters)
```
sveneberth, line 495:
```suggestion
            (and with wizard parameters)
```
? 🧙🪄

src/viur/core/request.py
```diff
@@ -568,73 +570,84 @@ def findAndCall(self, path: str, *args, **kwargs) -> None:
...
+        if not isinstance(caller, Method):
+            # try to find "index" function
+            if (index := caller.get("index")) and isinstance(index, Method):
+                caller = index
             else:
                 raise errors.MethodNotAllowed()
-        # Check for forceSSL flag
+
+        # Check for internal exposed
+        if caller.exposed is False and not self.internalRequest:
```
sveneberth, line 581: `None` should not be a problem, shouldn't it?

```suggestion
        if not caller.exposed and not self.internalRequest:
```

src/viur/core/request.py
```diff
@@ -568,73 +570,84 @@ def findAndCall(self, path: str, *args, **kwargs) -> None:
...
+                caller_args = self.args
+                caller_kwargs = self.kwargs
+
+            if (conf["viur.debug.traceExternalCallRouting"] and not self.internalRequest
+                    or conf["viur.debug.traceInternalCallRouting"]):
+                logging.debug(f"Calling {caller=} with {args=}, {kwargs=}")
+
+            res = caller(*caller_args, **caller_kwargs)
+
+            if not isinstance(res, bytes):  # Convert result into str if not bytes!
```
sveneberth, line 646: `encode()` returns `bytes`.
```suggestion
            if not isinstance(res, bytes):  # Convert the result to bytes if it is not already!
```

src/viur/core/module.py
```diff
@@ -1,20 +1,201 @@
...
 from viur.core.config import conf
 
 
+class Method:
+    """
+    Abstraction wrapper for any public available method.
+    """
+
+    @classmethod
+    def ensure(cls, func):
```
sveneberth, line 15:
```suggestion
    def ensure(cls, func: typing.Callable | "Method") -> "Method":
```

src/viur/core/module.py
```diff
@@ -1,20 +1,201 @@
...
+        unwanted results.
+        """
+        if isinstance(func, Method):
+            return func
+
+        return cls(func)
+
+    def __init__(self, func: typing.Callable):
+        # Attributes
+        self.exposed = None  # None = unexposed, True = exposed, False = internal exposed
```
sveneberth, line 28: Can we use a `Enum` here? This is very hacky.

src/viur/core/module.py
```diff
@@ -1,20 +1,201 @@
...
+
+        # Guards
+        self.skey = None
+        self.access = None
+
+        # Content
+        self._func = func
+        self._instance = None
+
+    def __get__(self, obj, objtype):
```
sveneberth, line 41: [Here](https://docs.python.org/3/howto/descriptor.html) they're using this syntax. Can it be optional?
```suggestion
    def __get__(self, obj, objtype=None):
```

src/viur/core/decorators.py
```diff
@@ -0,0 +1,141 @@
...
+        meth.skey = skey_config
+        return meth
+
+    if func is None:
+        return decorator
+
+    return decorator(func)
+
+
+def SKEY_ALLOW_EMPTY_FOR_KEY(args, kwargs):
```
sveneberth, line 137:
```suggestion
def SKEY_ALLOW_EMPTY_FOR_KEY(args: tuple, kwargs: dict) -> bool:
```

sveneberth (changes requested):
Furthermore I get this error:

```
  File "/project/deploy/main.py", line 164, in <module>
    app = core.setup(modules, render)
  File "/project/deploy/viur/core/__init__.py", line 220, in setup
    conf["viur.mainApp"] = buildApp(modules, render, default)
  File "/project/deploy/viur/core/__init__.py", line 151, in buildApp
    module_instance = module_cls(
  File "/project/deploy/viur/core/modules/user.py", line 902, in __init__
    setattr(self, name, provider(name, f"{modulePath}/{name}", self))
TypeError: TimeBasedOTP.__init__() takes 3 positional arguments but 4 were given
```


and 

```
  File "/project/deploy/main.py", line 156, in <module>
    import modules, render
  File "/project/deploy/modules/__init__.py", line 20, in <module>
    _import = __import__(_module, globals(), locals(), level=1)
  File "/project/deploy/modules/formmailer.py", line 1, in <module>
    from viur.core.modules.formmailer import Formmailer
  File "/project/deploy/viur/core/modules/formmailer.py", line 2, in <module>
    from viur.core import errors, utils, securitykey, email, skey
ImportError: cannot import name 'skey' from 'viur.core' (/project/deploy/viur/core/__init__.py)
```

src/viur/core/__init__.py
```diff
@@ -151,81 +104,96 @@ def buildApp(modules: Union[ModuleType, object], renderers: Union[ModuleType, Di
...
     if hasattr(modules, "index"):
-        if issubclass(modules.index, Module):
-            root = modules.index("index", "")
-        else:
-            root = modules.index()  # old style for backward compatibility
+        root = modules.index("index", "")
     else:
-        root = ExtendableObject()
-    modules._tasks = TaskHandler
+        root = Module("index", "")
```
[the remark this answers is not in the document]
phorward: We don't support "old fashioned" anymore. All must be inherited from `Module`-class.

src/viur/core/module.py
phorward: I've thought of it first, but then I came to the result that a function should either be "exposed" or not in the future. I really want to get rid of the "internal exposed". I think this will die out with all newer developments we do.

src/viur/core/module.py
phorward: Better it is...

src/viur/core/module.py
```diff
@@ -168,3 +407,36 @@ def describe(self) -> Union[Dict, None]:
...
+
+        # Map module under SEO-mapped name, if available.
+        if self.seo_language_map:
+            for lang in conf["viur.availableLanguages"] or [conf["viur.defaultLanguage"]]:
+                # Map the module under each translation
+                if translated_module_name := self.seo_language_map.get(lang):
+                    translated_module = target.setdefault(translated_module_name, {})
+
+                    # Map module functions to the previously determined target
+                    for name, func in functions.items():
```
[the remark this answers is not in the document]
phorward: you're right... I've forgot to solve this during the PR. Will do soon.

src/viur/core/module.py
```diff
@@ -168,3 +407,36 @@ def describe(self) -> Union[Dict, None]:
...
+                        func.register(translated_module, name, lang)
+
+            conf["viur.languageModuleMap"][module_name] = self.seo_language_map
```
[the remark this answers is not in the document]
phorward: Same here.

src/viur/core/render/html/env/viur.py
```diff
@@ -86,11 +86,12 @@ def execRequest(render: Render, path: str, *args, **kwargs) -> Any:
...
-        or ((not "exposed" in dir(caller)
-             or not caller.exposed))
-        and (not "internalExposed" in dir(caller)
-             or not caller.internalExposed)):
+
+    viur_flags = getattr(caller, "viur_flags", {})
+    if not (hasattr(caller, '__call__') and (
+        ("exposed" in viur_flags and viur_flags["exposed"]) or
+        ("internal_exposed" in viur_flags and viur_flags["internal_exposed"])
+    )):
```
phorward, line 94: Here is something that still must be reworked by  me...

src/viur/core/request.py
phorward: No, it aren't wizard parameters. The docstring is generally wrong, I've just fixed a typo.

src/viur/core/request.py
phorward: This line is correct.

exposed is
- True: the function is generally exposed
- False: the function is internally exposed
- None: the function is not exposed

src/viur/core/utils.py
```diff
@@ -248,8 +248,9 @@ def seoUrlToFunction(module: str, function: str, render: Optional[str] = None) -
         targetObject = targetObject[render]
     if function in targetObject:
         func = targetObject[function]
-        if getattr(func, "seoLanguageMap", None) and lang in func.seoLanguageMap:
-            pathComponents.append(func.seoLanguageMap[lang])
+        viur_flags = getattr(func, "viur_flags", {})
+        if seoLanguageMap := viur_flags.get("seoLanguageMap", {}) and lang in seoLanguageMap:
+            pathComponents.append(seoLanguageMap[lang])
```
phorward, line 253: Again something I need to do!

src/viur/core/module.py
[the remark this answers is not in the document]
phorward: fixed

src/viur/core/module.py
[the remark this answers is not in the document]
phorward: fixed

XeoN-GHMB (approved):

phorward:
> Furthermore I get this error:
> ...
> and
> ...

Errors had been fixed meanwhile.

skoegl (approved):

PR #807, opened 2023-08-09
Fix docs configuration after #804

sveneberth:
The [last build](https://readthedocs.org/projects/viur-core/builds/21561968/) failed


phorward (approved):
Thanks!

PR #808, opened 2023-08-10
feat: Implement a `PeriodicTask` to check the remaining SIB email quota

### LINKED ISSUE #741: Add SIB Quota Warn `PeriodicTask`

sveneberth:
Resolves #741

### DIFF
--- a/src/viur/core/email.py
+++ b/src/viur/core/email.py
@@ -389,7 +389,8 @@ class EmailTransportSendInBlue(EmailTransport):
         This task does not have to be enabled.
         It automatically checks if the apiKey is configured.
 
-        There are three thresholds: 1000, 500, 100
+        There are three default thresholds: 1000, 500, 100
+        Others can be set via conf["viur.email.sendInBlue.thresholds"].
         An email will be sent for the lowest threshold that has been undercut.
         """
         if conf.get("viur.email.sendInBlue.apiKey") is None:
@@ -413,7 +414,8 @@ class EmailTransportSendInBlue(EmailTransport):
 
         logging.info(f"SIB E-Mail credits: {credits}")
 
-        for idx, limit in list(enumerate((1000, 500, 100), 1))[::-1]:
+        thresholds = conf.get("viur.email.sendInBlue.thresholds", (1000, 500, 100))
+        for idx, limit in list(enumerate(thresholds, 1))[::-1]:
             if credits < limit:
                 sendEMailToAdmins(
                     f"SendInBlue email budget for {conf['viur.instance.project_id']}: "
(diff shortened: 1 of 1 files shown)


src/viur/core/email.py
```diff
@@ -375,3 +380,45 @@ def validateQueueEntity(entity: db.Entity):
...
+        for plan in data["plan"]:
+            if plan["type"] == "payAsYouGo":
+                credits = plan["credits"]
+                break
+        else:
+            credits = -1
+
+        logging.info(f"SIB E-Mail credits: {credits}")
+
+        for idx, limit in list(enumerate((1000, 500, 100), 1))[::-1]:
```
ArneGudermann, line 416: Maybe we can make this configurable like. 
```suggestion
        for idx, limit in list(enumerate(conf["viur.email.sendInBlue.thresholds"], 1))[::-1]:
```

phorward (changes requested):
Hello @sveneberth,
thanks for creating this pull request. Should we generally still use the naming covention "SendInBlue" or the new, ugly name "Brevo", which seems that it should sound like "bravo", but shows how silly the company owner's there are by replacing a good name by a stupid and ugly one.

src/viur/core/email.py
sveneberth: This probably makes sense. Depending on how high the email volume of the website other levels are necessary.

sveneberth:
Hey reviewers,

I applied @ArneGudermann suggestion and added configurable thresholds.

Furthermore I ensure we don't waste the remaining credits with warning mails and send emails only once for each limit.

@phorward The rebranding completely passed me by, thanks for the info.
First of all, I would leave it constant at SIB. But soon we should rebrand it once, otherwise it can be confusing for new developers. But we should check before if the old api keys work with the new api URL and so on.

In general I would change the transport class to an instances.
I think it is cleaner to configure an instance than the `conf` like this:
```
conf["viur.email.transport"] = EmailTransportSendInBlue(
	api_key="*************************",
	thresholds=(500,300,100),
)
```

phorward:
> In general I would change the transport class to an instances. I think it is cleaner to configure an instance than the `conf` like this:
> 
> ```python
> conf["viur.email.transport"] = EmailTransportSendInBlue(
> 	api_key="*************************",
> 	thresholds=(500, 300, 100),
> )
> ```

Do you want to do it within this pull request, or is it separated?

sveneberth:
> Do you want to do it within this pull request, or is it separated?

No, no in this PR anymore. I have to see if I can do that without breaking changes or if I have to do it with ViUR4.

phorward (approved):

PR #809, opened 2023-08-11
feat: Implement natural sorting in `StringBone`

sveneberth:
This enable a more natural sorting for umlauts according to DIN 5007 Variant 2 (or a custom one).

It does not use [`natsort`](https://pypi.org/project/natsort/), because `natsort` needs a German locale set for umlauts, which (afaik) does not work in the gae.


src/viur/core/bones/string.py
```diff
@@ -196,28 +210,25 @@ def buildDBSort(
...
-                order = (prop, db.SortOrder.Descending)
-            elif "orderdir" in rawFilter and rawFilter["orderdir"] == "2":
-                order = (prop, db.SortOrder.InvertedAscending)
-            elif "orderdir" in rawFilter and rawFilter["orderdir"] == "3":
-                order = (prop, db.SortOrder.InvertedDescending)
+                prop = f"{name}.{lang}"
             else:
-                order = (prop, db.SortOrder.Ascending)
-            inEqFilter = [x for x in dbFilter.queries.filters.keys() if  # FIXME: This will break on multi queries
-                          (">" in x[-3:] or "<" in x[-3:] or "!=" in x[-4:])]
```
sveneberth, line 220: This part has only been refactored.

ArneGudermann (dismissed):

phorward (approved):

src/viur/core/bones/string.py
```diff
@@ -229,6 +241,29 @@ def buildDBSort(
...
+        and sets ö and oe, etc. equal.
+        """
+        if value is None:
+            return None
+        assert isinstance(value, str)
+        if not self.caseSensitive:
+            value = value.lower()
+
+        # DIN 5007 Variant 2
+        return value.translate(str.maketrans({
```
phorward, line 257: This is useful! [translate](https://docs.python.org/3/library/stdtypes.html#str.translate) should be used on several ugly and stupid replace-cascades.

phorward (approved):

ArneGudermann (approved):

PR #812, opened 2023-08-15
feat: Add bones parameter for renderEditForm

ArneGudermann:
This PR extends the function `renderEditForm` with the parameter `bones`. If this is set, only the bones that are in the list will be rendered.

### DIFF
--- a/src/viur/core/render/html/env/viur.py
+++ b/src/viur/core/render/html/env/viur.py
@@ -574,6 +574,8 @@ def renderEditForm(render: Render,
     rowTpl = render.getEnv().get_template(render.getTemplateFileName("editform_row"))
     sections = OrderedDict()
 
+if both := set(ignore).intersection(bones):
+    raise ValueError(f"You have specified the same bones {', '.join(both)} in *ignore* AND *bones*!")
     for boneName, boneParams in skel["structure"].items():
         category = str("server.render.html.default_category")
         if "params" in boneParams and isinstance(boneParams["params"], dict):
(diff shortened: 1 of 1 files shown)


src/viur/core/render/html/env/viur.py
```diff
@@ -570,6 +573,7 @@ def renderEditForm(render: Render,
     sectionTpl = render.getEnv().get_template(render.getTemplateFileName("editform_category"))
     rowTpl = render.getEnv().get_template(render.getTemplateFileName("editform_row"))
     sections = OrderedDict()
+
```
sveneberth, line 576: Can we add a check like this before?

```suggestion

if both := set(ignore).intersection(bones):
    raise ValueError(f"You have specified the same bones {', '.join(both)} in *ignore* AND *bones*!")
```

sveneberth (changes requested):

src/viur/core/render/html/env/viur.py
```diff
@@ -570,6 +573,9 @@ def renderEditForm(render: Render,
...
+if both := set(ignore).intersection(bones):
+    raise ValueError(f"You have specified the same bones {', '.join(both)} in *ignore* AND *bones*!")
```
sveneberth, line 578:
```suggestion
    if both := set(ignore).intersection(bones):
        raise ValueError(f"You have specified the same bones {', '.join(both)} in *ignore* AND *bones*!")
```

sveneberth (approved):

phorward (approved):

PR #815, opened 2023-08-16
feat: Add `secret` module to access values from GC secret manager

phorward:
Hello, I would like to add `google-cloud-secret-manager` and the `secret` module to viur-core, as several projects already use it and it could be useful for some standard extensions soon.


sveneberth (changes requested):
A dependency should only be added if it is used by the package. Otherwise it is just unnecessary ballast. 

If you provide a low-level function that returns the value of a secret with one call and then uses this library, I would have no problem with that. But I think this is wrong here in the core. But we could add it in the base as a preset.

phorward:
> If you provide a low-level function that returns the value of a secret with one call and then uses this library, I would have no problem with that. But I think this is wrong here in the core. But we could add it in the base as a preset.

OK you're right. Maybe a utility function satisfies for that.

phorward:
@sveneberth I've modified the pull request to provide a `secret` module which cointains a `get` function for Google Cloud secret manager access.

src/viur/core/secret.py
```diff
@@ -0,0 +1,27 @@
+import logging
+from google.cloud import secretmanager
+from viur.core.config import conf
+
+"""
+This module provides utiliy functions for accessing values stored in the Google Cloud Secret Manager.
```
sveneberth, line 6:
```suggestion
This module provides utility functions for accessing values stored in the Google Cloud Secret Manager.
```

src/viur/core/secret.py
sveneberth, line 1: I'm not completely satisfied with the module name `secret` as it is almost called as python own [`secrets` module](https://docs.python.org/3/library/secrets.html)
But would be okay for me.

src/viur/core/secret.py
```diff
@@ -0,0 +1,27 @@
...
+"""
+
+# Global secret manager client instance
+__client = secretmanager.SecretManagerServiceClient()
+
+
+def get(secret: str, version: str = "latest") -> str:
```
sveneberth, line 13: Since the versions are incremented `int`s and we are using string interpolation, this should be fine too.


```suggestion
def get(secret: str, version: int | str = "latest") -> str:
```

src/viur/core/secret.py
```diff
@@ -0,0 +1,27 @@
...
+    """
+    Retrieves a secret stored in Google Cloud Secret Manager for use within the application.
+
+    Add a secret online under https://console.cloud.google.com/security/secret-manager.
+    Service accounts requires the role "Secret Manager Secret Accessor" in IAM.
+    """
+    name = f"""projects/{conf["viur.instance.project_id"]}/secrets/{secret}/versions/{version}""".replace(".", "_")
```
sveneberth, line 20: What's the reason for the `replace(".", "_")`? Dots (`.`) are not allowed in secret names, so just don't use them!

src/viur/core/secret.py
```diff
@@ -0,0 +1,27 @@
...
+
+    try:
+        return __client.access_secret_version(request={"name": name}).payload.data.decode()
+    except Exception as e:
+        logging.exception(e)
+
+    return None
```
sveneberth, line 27: I think silent error is a stupid idea. Exceptions are there to send signals and not to be ignored. If I explicitly call a function, I also want to get feedback in case of an error.

Imagine you use a zip function (which creates a zip archive), which has the parameter `password`. As argument you pass the return value of the `secret.get` method.
If your secret is not set or the permissions are set incorrectly the zip archive might be created without password protection (because `None` might be the input for "no password protection"). You will only see later in the logs that your software has already distributed open zips instead of aborting the process with an exception.

```suggestion
    return __client.access_secret_version(request={"name": name}).payload.data.decode()
```

sveneberth (changes requested):

src/viur/core/secret.py
phorward: yeah, but this is in the viur.core namespace. And it should not be bound to Google.

phorward:
@sveneberth I've updated the PR, can you please re-review?

XeoN-GHMB (approved):

sveneberth (approved):

PR #825, opened 2023-09-04
fix: Add check for  `Method` class

ArneGudermann:
Since #800 you can't make a method `exposed` and to a `jinjaGlobalFunction`. This PR fix this.


phorward:
#827 replaces this PR.

PR #828, opened 2023-09-05
refactor: Some clean-up on #800

phorward:
This PR cleans up code from #800

### DIFF
--- a/src/viur/core/modules/user.py
+++ b/src/viur/core/modules/user.py
@@ -284,7 +284,7 @@ class UserPassword(UserAuthentication):
     @exposed
     @force_ssl
     @skey(allow_empty=True)
-    def login(self, *, name: str = "", password: str = "", **kwargs):
+    def login(self, *, name: str | None = None, password: str | None = None, **kwargs):
         if current.user.get():  # User is already logged in, nothing to do.
             return self._user_module.render.loginSucceeded()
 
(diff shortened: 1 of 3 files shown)


src/viur/core/modules/user.py
```diff
@@ -281,11 +281,10 @@ class LostPasswordStep2Skel(skeleton.RelSkel):
             required=True,
         )
 
-
     @exposed
     @force_ssl
     @skey(allow_empty=True)
-    def login(self, *, name=None, password=None, **kwargs):
+    def login(self, *, name: str = "", password: str = "", **kwargs):
```
sveneberth, line 287: What makes the difference?

Both values wouldn't pass l.291 (`if not name or not password:`)

phorward: Empty str and None is both False.
```python
>>> name = ""
>>> password = ""
>>> not name or not password
True
```

sveneberth: Yes, that is known. That's why I'm asking where the difference or the concrete advantage lies. I think `None` makes more sense for the absence of a parameter than an empty string, which says that the value was sent but is wrong or empty.

sveneberth: Okay. That was more a question of interest. For me, both variants would be okay.

phorward: Would you be happier with the `str | None = None` notation?

sveneberth (approved):

PR #842, opened 2023-09-08
refactor: Refactor `Router` + collect context-variables to `Router.context`

phorward:
Initial commit, refactor of BrowseHandler to Request. This is a first attempt to refactor this module without doing too big changes.

Contexts are already used in several ViUR projects to turn requests into a specific data environment. Context-parameters are preceded by an "@". This can be changed to a dynamic or customizable version in a next release.

This behavior is wanted:
- Request.kwargs contains all kwargs (because of backward compatibility)
- caller(kwargs) is a filtered kwargs without contexts
- Request.contexts contains only contexts

### DIFF
--- a/src/viur/core/request.py
+++ b/src/viur/core/request.py
@@ -521,8 +521,15 @@ class Request:
                 logging.debug("Caching disabled by X-Viur-Disable-Cache header")
                 self.disableCache = True
 
+        # Copy contexts into self.contexts if available
+        if contexts := {k: v for k, v in self.kwargs.items() if k.startswith("@")}:
+            kwargs = {k: v for k, v in self.kwargs.items() if k not in contexts}
+            self.contexts |= contexts
+        else:
+            kwargs = self.kwargs
+
         # Now call the routed method!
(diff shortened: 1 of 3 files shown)


XeoN-GHMB (dismissed):

src/viur/core/request.py
```diff
@@ -76,7 +74,7 @@ def validate(request: 'BrowseHandler') -> typing.Optional[typing.Tuple[int, str,
         return 403, "Forbidden", "Request rejected due to fetch metadata"
 
 
-class BrowseHandler():  # webapp.RequestHandler
+class Request:
```
sveneberth, line 77: I would call it `Router`, otherwise you've nested the `Request` of the `viur-core` and the `Request` of webob.
```suggestion
class Router:
```

sveneberth (changes requested):

src/viur/core/request.py
phorward: But the context variable is called `current.request`, therefore I chose this name. Nevermind, I want to finalize rc1 now.

PR #845, opened 2023-09-08
fix: template name should not be static.

sveneberth:
Look for the attribute `second_factor_choice_template` in the current renderer or the parent module (`user` module) and use it's value. By default this should be `"user_second_factor_choice"`.

### DIFF
--- a/src/viur/core/modules/user.py
+++ b/src/viur/core/modules/user.py
@@ -1196,7 +1196,7 @@ class User(List):
             # We have only one second factor we don't need the choice template
             return second_factor_providers[0].start(userKey)
         # In case there is more than one second factor, let the user select a method.
-        return self.render.second_factor_choice(tpl="second_factor_choice", second_factors=second_factor_providers)
+        return self.render.second_factor_choice(second_factors=second_factor_providers)
 
     def secondFactorSucceeded(self, secondFactor, userKey):
         session = current.session.get()
--- a/src/viur/core/render/html/user.py
+++ b/src/viur/core/render/html/user.py
@@ -11,6 +11,7 @@ class Render(DefaultRender):  # Render user-data to xml
     passwdRecoverInfoTemplate = "user_passwdrecover_info"
     otp_add_template = "user_secondfactor_add"
     otp_add_success_template = "user_secondfactor_add_success"
+    second_factor_choice_template = "user_second_factor_choice"
 
     def _choose_template(self, tpl: None | str, fallback_attribute: str) -> str:
         if tpl:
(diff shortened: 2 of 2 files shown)


phorward (dismissed):

src/viur/core/modules/user.py
```diff
@@ -1196,7 +1196,7 @@ def continueAuthenticationFlow(self, caller, userKey):
             # We have only one second factor we don't need the choice template
             return second_factor_providers[0].start(userKey)
         # In case there is more than one second factor, let the user select a method.
-        return self.render.second_factor_choice(tpl="second_factor_choice", second_factors=second_factor_providers)
+        return self.render.second_factor_choice(second_factors=second_factor_providers)
```
phorward, line 1199: Why isn't there a template configuration in the User-module, so `tpl=self.second_factor_choice_template` should be right?

```suggestion
        return self.render.second_factor_choice(tpl=self.second_factor_choice_template, second_factors=second_factor_providers)
```

phorward (changes requested):

XeoN-GHMB (dismissed):

src/viur/core/modules/user.py
sveneberth: Because it's predefined in the `user.Render` and could be optionally added to the `user` module.

Inside the user module you call the `second_factor_choice` method of the `user.Render` module. This method uses `User.Render._choose_template` to choose the template:  https://github.com/viur-framework/viur-core/blob/49eb6b7ad432e6937fa8dc0cb9823b1839714923/src/viur/core/render/html/user.py#L79
If you don't provide the `tpl`, the first two lines https://github.com/viur-framework/viur-core/blob/49eb6b7ad432e6937fa8dc0cb9823b1839714923/src/viur/core/render/html/user.py#L16-L18 will be skipped.
Now we check if the `user` module (which is the `parent` of the renderer) has the attribute `second_factor_choice_template` (passed as `fallback_attribute`): https://github.com/viur-framework/viur-core/blob/49eb6b7ad432e6937fa8dc0cb9823b1839714923/src/viur/core/render/html/user.py#L19-L20
This is not the case by default. But if you want, you can add `second_factor_choice_template = "my_custom_template" to your `user` module and it would use this template. 
And as a final fallback, the default will be used in the renderer:
https://github.com/viur-framework/viur-core/blob/49eb6b7ad432e6937fa8dc0cb9823b1839714923/src/viur/core/render/html/user.py#L21

---

This behavior may be a bit confusing and not straightforward, but this is how the other methods currently work, like login: 
https://github.com/viur-framework/viur-core/blob/49eb6b7ad432e6937fa8dc0cb9823b1839714923/src/viur/core/modules/user.py#L298
https://github.com/viur-framework/viur-core/blob/49eb6b7ad432e6937fa8dc0cb9823b1839714923/src/viur/core/render/html/user.py#L28-L31

So the behavior should be the same. In a future release we can of course rework it.

phorward: OK, thanks for clarification.

phorward (dismissed):

ArneGudermann (approved):

PR #846, opened 2023-09-08
fix some mixed up and blurred methods and concepts.

sveneberth:
- The methods `second_factor_add` and `second_factor_add_success` are generic. So their template name should also be generic and not include `"otp"`.
- Since the above methods are general, it must be known for which concrete 2nd factor they are currently rendered. So the `name` and `action_name` are missing.
- Rendering the method `second_factor_choice` without specifying `second_factors` makes no sense. So the parameter should not be keyword-only.
- Add `add_url` in `second_factor_add` where data should be POST to.

### DIFF
--- a/src/viur/core/modules/user.py
+++ b/src/viur/core/modules/user.py
@@ -647,6 +647,7 @@ class UserSecondFactorAuthentication(UserAuthentication, abc.ABC):
     def __init__(self, moduleName, modulePath, _user_module):
         super().__init__(moduleName, modulePath, _user_module)
         self.action_url = f"{self.modulePath}/{self.ACTION_NAME}"
+        self.add_url = f"{self.modulePath}/add"
         self.start_url = f"{self.modulePath}/start"
 
     @abc.abstractmethod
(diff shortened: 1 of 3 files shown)


src/viur/core/render/html/user.py
```diff
@@ -9,8 +11,8 @@ class Render(DefaultRender):  # Render user-data to xml
     verifySuccessTemplate = "user_verify_success"
     verifyFailedTemplate = "user_verify_failed"
     passwdRecoverInfoTemplate = "user_passwdrecover_info"
-    otp_add_template = "user_secondfactor_add"
-    otp_add_success_template = "user_secondfactor_add_success"
+    second_factor_add_template = "user_secondfactor_add"
```
phorward, line 14: IMHO, the choice-template must be set also here, see https://github.com/viur-framework/viur-core/pull/845/files#r1320525611

phorward (changes requested):
Please check comment in #845  with my annotation here.

XeoN-GHMB (approved):

src/viur/core/render/html/user.py
sveneberth: Yeah, that's done in #845. 

The PRs reflect just the consequence of my adjustments of the templates for the base. Therefore, the different templates are packed into different PRs.

src/viur/core/modules/user.py
```diff
@@ -904,17 +905,26 @@ def add(self, otp=None):
 
         if otp is None:
             return self._user_module.render.second_factor_add(
-                tpl=self.otp_add_template,
+                tpl=self.second_factor_add_template,
```
sveneberth, line 908: Btw: here we have to pass the `tpl` parameter because we are not inside the `user` module. But we are using the `user.Render`, where the `parent` is the `user` module. So `_choose_template` won't use the attribute of that class (where it logically belongs).

phorward (approved):

PR #848, opened 2023-09-11
fix: Clean-up user/view and user/edit with "self"

phorward:
This is only a suggestion. Please comment for any ambiguity or improvement. We can also remove /user/edit/self completely, or make "self" a mandatory parameter, so that "/user/view" or "/user/edit" are not allowed without parameters.


XeoN-GHMB (approved):

ArneGudermann (approved):

sveneberth (approved):
I would prefer it if `"self"` had to be explicitly used as a key argument instead of implying it as a fallback.
But if this was already the case before, we should keep it that way for now for compatibility reasons and change it in a later release.

PR #850, opened 2023-09-12
fix: Replace forward_argument by forward_payload in `/user/verify`

ArneGudermann:


phorward (changes requested):
Hello @ArneGudermann, are you sure this code will work, when no skey is provided? 
And is the skey really optional, when empty?
In such case, kwargs["skey"] will be unset, and raise a KeyError.

My proposal for the entire pull request:
```python
    @exposed
    @skey(forward_payload="data", session_bound=False)
    def verify(self, data, *args, **kwargs):
```
Sure this isn't tested, but can you do it this way?

XeoN-GHMB (approved):

ArneGudermann:
@phorward `allow_empty` is needed in this case. Because of this https://github.com/viur-framework/viur-core/blob/b3d156ee5e3d87fe7afcfca38dc15aef141cc6e2/src/viur/core/module.py#L212 And your code doesn't work because `data` is a requiered argument in this case and this is not sent along. so we have to take the way over the kwargs.

phorward:
> @phorward `allow_empty` is needed in this case. Because of this
> 
> https://github.com/viur-framework/viur-core/blob/b3d156ee5e3d87fe7afcfca38dc15aef141cc6e2/src/viur/core/module.py#L212

This still doesn't make sense. `kwargs["skey_data"]` as provided by your updated commit is still not set when no skey is providing, resulting in a error 500.

> And your code doesn't work because `data` is a requiered argument in this case and this is not sent along. so we have to take the way over the kwargs.

As I've said, the code was not tested. I think you should remove the `allow_empty=True` from the @skey configuration, to ensure `skey_data` is made available. This, in turn, does not require to use  `kwargs["skey_data"]`  and you can just pass the payload from the skey via "data" as parameter "data" to your function, yielding in 

```python
    @exposed
    @skey(forward_payload="data", session_bound=False)
    def verify(self, data, *args, **kwargs):
```

phorward (dismissed):

phorward:
Ok now I've tested together with @ArneGudermann and it works as expected.

PR #853, opened 2023-09-12
feat: Add `run_in_transaction` decorator.

ArneGudermann:
This PR adds a  `run_in_transaction` decorator. Now you can call transaction like : 

```python
@exposed
	def test(self, key):
		@run_in_transaction
		def trans(key):
			e = db.Get(db.Key.from_legacy_urlsafe(key))
			assert e
			e["new"] = "123"
			return e

		return trans(key)
```
I think it looks cleaner.
This is only an Idea What do you think of it?

### DIFF
--- a/src/viur/core/decorators.py
+++ b/src/viur/core/decorators.py
@@ -133,3 +133,20 @@ def skey(
         return decorator
 
     return decorator(func)
+
+
+def run_in_transaction(func: Callable) -> Method:
+    """
+    Decorator, which enforces the Method to run in a transaction.
+    .. code-block:: python
+        from viur.core.decorators import run_in_transaction
+        def outer_method(key)
+            @run_in_transaction
+            def transaction_method(key):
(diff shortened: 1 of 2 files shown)


sveneberth:
I still don't understand why we should have a method for this and not have a context manager implementation like [Google's Datastore client](https://cloud.google.com/python/docs/reference/datastore/latest/transactions).

```python
@exposed
	def test(self, key):
		with db.Transaction():
			e = db.Get(db.Key.from_legacy_urlsafe(key))
			assert e
			e["new"] = "123"
			return e
```
is so much more simple and pythonic

ArneGudermann:
The context of the transaction is currently  handled by the viur-datastore. But we can exposed it to the core.

sveneberth:
> Why not simply...
```python
skel = MySkeleton(auto_save=True)
skel["count"] += 1
```

PR #855, opened 2023-09-12
fix: `html.Render.getTemplateFileName()` should be deterministic

sveneberth:
This bug was introduced in https://github.com/viur-framework/viur-core/pull/658 We can't use a `set` here because it's unordered. So it's random which template is chosen in case of a `style` or `lang` argument.

Just make sure we don't add duplicates.


phorward (approved):
Thanks!

Issue #811, opened 2023-08-14
Replace `str.replace`-cascades by `str.translate` calls


phorward:
This is useful! [translate](https://docs.python.org/3/library/stdtypes.html#str.translate) should be used on several ugly and stupid replace-cascades.

_Originally posted by @phorward in https://github.com/viur-framework/viur-core/pull/809#discussion_r1293277163_

PR #873, opened 2023-09-22
refactor: replace usage of `str.replace` with `str.translate` on several places

cheeaza:
This PR replaces some ugly str.replace-cascades with the str.translate method as mentioned in https://github.com/viur-framework/viur-core/issues/811

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,12 +1,11 @@
-/*.pyc
 *.pyc
 *.bak
-.idea/
-/docs/build/
+.idea
+.vscode
 dist
 viur_core.egg-info
 .env
-/tests/htmlcov/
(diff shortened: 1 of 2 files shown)


sveneberth (changes requested):
Hey @cheeaza, it would be more efficient to move the translation dicts at module level to instantiate them only once at instance warmup. And I think we should use [`str.maketrans`](https://docs.python.org/3.3/library/stdtypes.html?highlight=maketrans#str.maketrans) with the normal unicode characters, which is more readable and maintainable than the ordinals. If you do this during warm-up, the runtime is negligible.

Last but not least, please fix the linter issues :x: --> :heavy_check_mark:

PR #878, opened 2023-09-27
feat: `UserPrimaryAuthentication` with unified `can_handle()`

phorward:
This is both a feature and refactoring. Any `UserAuthentication` provides a can_handle()-function which is evaluated on appropiate positions.

Also refactors:

- `get_config()` and `can_handle()` should accept `user` parameter and not `possible_user` which is confusing
- `continueAuthenticationFlow` first performs user and handling check, then does session-stuff
- Renamed `caller` into `provider` and `userKey` into `user_key`

### DIFF
--- a/src/viur/core/modules/user.py
+++ b/src/viur/core/modules/user.py
@@ -221,8 +221,15 @@ class UserAuthentication(Module):
         super().__init__(moduleName, modulePath)
         self._user_module = userModule
 
+    def can_handle(self, user: db.Entity) -> bool:
+        return True
 
-class UserPassword(UserAuthentication):
+
+class UserPrimaryAuthentication(UserAuthentication, abc.ABC):
+    """Abstract class for all primary authentication methods."""
+
+
+class UserPassword(UserPrimaryAuthentication):
     registrationEnabled = False
     registrationEmailVerificationRequired = True
     registrationAdminVerificationRequired = True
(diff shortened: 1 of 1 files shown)


XeoN-GHMB (dismissed):

src/viur/core/modules/user.py
```diff
@@ -221,8 +221,15 @@ def __init__(self, moduleName, modulePath, userModule):
         super().__init__(moduleName, modulePath)
         self._user_module = userModule
 
+    def can_handle(self, user: db.Entity) -> bool:
+        return True
 
-class UserPassword(UserAuthentication):
+
+class UserPrimaryAuthentication(UserAuthentication, abc.ABC):
+    """Abstract class for all primary authentication methods."""
```
ArneGudermann, line 229: Why does this class exist without a member or a method ?
```suggestion
class UserPrimaryAuthentication(UserAuthentication, abc.ABC):
    """Abstract class for all primary authentication methods."""
    registrationEnabled = False
```
We can add for example `registrationEnabled = False` or `def getAuthMethodName`

ArneGudermann (changes requested):

src/viur/core/modules/user.py
phorward: > Why does this class exist without a member or a method ?
> We can add for example `registrationEnabled = False` or `def getAuthMethodName`

Yes, please! Make a proposal or edit the PR. I've added the class only to clearly identify primary authentication providers.

ArneGudermann: ok.

XeoN-GHMB (approved):

PR #881, opened 2023-10-04
fix: Pass `kwargs` to `renderEntry` on json Renderer

ArneGudermann:
This PR fix the Bug that the `kwargs` are not passed to the render `renderEntry`. 
So  this not work. 
https://github.com/viur-framework/viur-core/blob/20f16774f2cb356c5035d3ff6af0282f861498c8/src/viur/core/modules/user.py#L734-L740

### DIFF
--- a/src/viur/core/render/json/default.py
+++ b/src/viur/core/render/json/default.py
@@ -139,7 +139,7 @@ class DefaultRender(object):
                                                       expires=conf["viur.render.json.downloadUrlExpiration"])
         return res
 
-    def renderEntry(self, skel: SkeletonInstance, actionName, params=None):
+    def renderEntry(self, skel: SkeletonInstance, actionName, params=None, **kwargs):
         structure = None
         errors = None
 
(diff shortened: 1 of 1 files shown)


phorward (changes requested):
Hello @ArneGudermann,
well I think this is the wrong solution for the problem.
The `params` where added some years ago to all renderers to pass customized values to the specific renderer. In this case, the code part

https://github.com/viur-framework/viur-core/blob/20f16774f2cb356c5035d3ff6af0282f861498c8/src/viur/core/modules/user.py#L734-L740

Is the problem, and should become

```python
return self._user_module.render.edit( 
    self.OtpSkel(),
    params={
         "name": translate(self.NAME),
         "action_name": self.ACTION_NAME,
         "action_url": f"{self.modulePath}/{self.ACTION_NAME}"
    },
    tpl=self.second_factor_login_template 
)
```
Though, `tpl` is only a parameter relevant for the Jinja2 (html) renderer, this must be ignored in any other renders.

The problem should be solved this way. The pros for going this way is:
- The JSON/XML/msgpack-message format is always the same
- kwargs cannot be misused to override values like "action" or "skel" in any way

Things to do:
- Fix the call in `TimeBasedOTP` to use params, as suggested above
- Fix at least JSON-renderer to accept tpl-flag, just for throwing it away.

I hope this clarifies things a bit.

ArneGudermann:
Okay I close this PR and make a new one.

PR #888, opened 2023-10-10
fix: Readd self for kwargs in request `_route`

ArneGudermann:
Readd `self` for kwargs. Fix https://github.com/viur-framework/viur-core/pull/884


phorward (approved):
wtf, I'm sorry!

PR #890, opened 2023-10-10
Fix: Second Factor `start`

ArneGudermann:
Fix / Followup for #881. Now the `action_url` and so on are send over the `params` parameter.


XeoN-GHMB (approved):

phorward (approved):
Yeah, that is the way it should be done.

PR #900, opened 2023-10-11
fix: Built a standardized way for the return of errors in 2Factor

ArneGudermann:


src/viur/core/modules/user.py
```diff
@@ -773,14 +775,14 @@ def otp(self, *args, **kwargs):
         if res is None:
             otp_user_conf["attempts"] = attempts + 1
             session.markChanged()
-
+            skel = self.OtpSkel()
```
phorward, line 778: You can re-use `skel` defined in line 760 for this.
```suggestion
```

src/viur/core/modules/user.py
```diff
@@ -773,14 +775,14 @@ def otp(self, *args, **kwargs):
...
+            skel.errors = [ReadFromClientError(ReadFromClientErrorSeverity.Invalid, "Wrong OTP Token", ["otptoken"])]
```
phorward, line 779: Error shall not be set on first call.
```suggestion
            if attempts > 0:
                skel.errors = [ReadFromClientError(ReadFromClientErrorSeverity.Invalid, "Wrong OTP Token", ["otptoken"])]
```

src/viur/core/modules/user.py
```diff
@@ -1014,7 +1018,7 @@ def authenticator_otp(self, **kwargs):
         if AuthenticatorOTP.verify_otp(otp=otp_token, secret=user["otp_app_secret"]):
             return self._user_module.secondFactorSucceeded(self, user_key)
 
-        skel.errors = [ReadFromClientError(ReadFromClientErrorSeverity.Invalid, "Wrong OTP Token")]
+        skel.errors = [ReadFromClientError(ReadFromClientErrorSeverity.Invalid, "Wrong OTP Token", ["otptoken"])]
```
phorward, line 1021: The handler `AuthenticatorOTP` doesn't seem to hold a counter for a maximum retry call. Can you please fix this, as it's the case in `TimeBasedOTP`.

phorward (changes requested):

src/viur/core/modules/user.py
ArneGudermann: But why ? We start with the `start` method and send the token and the we must send an error back ?

phorward: OK it seems you're right.

phorward (approved):

PR #901, opened 2023-10-11
feat: Add user admin login context

ArneGudermann:
This PR add a check for a contextvar for the new admi and sends back only json.

### DIFF
--- a/src/viur/core/render/vi/user.py
+++ b/src/viur/core/render/vi/user.py
@@ -1,3 +1,4 @@
+from viur.core import current
 from viur.core.render.json.user import UserRender as user
 import string, json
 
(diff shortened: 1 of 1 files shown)


phorward:
Hello @ArneGudermann, I generally want to remove this stupid HTML-redirect-solution.
Unfortunately, I must confess that I made this stupid solution years ago, but I do not know why anymore. Therefore I would like to remove it in general, so there is no need for a context-variable to check for. Can you please verify how the old VI behaves when logging in, both with Username+Password, and with Google, but with a `json.dumps("OKAY")` only?

ArneGudermann:
It testyo

ArneGudermann:
I have tried it and it does not work with the Google login.

phorward:
Ok, sh*t! Then we have to fix it differently. Lets talk about this in the VIUR meeting.

sveneberth:
Suggestion: The vi-admin -- or any other "client" -- could send an `Accept` header. In the case of the vi-admin `Accept: application/json`. The core can send an appropriate response based on this type. For compatibility reasons this could be `text/html` by default (if omitted) and JSON if explicitly provided. In the future, of course, it should be JSON by default.

---

Another suggestion: Switch the response type in general based on a [compatibility flag](https://github.com/viur-framework/viur-core/blob/926d42f9f3062d28bc826b5e43bce5216499f229/src/viur/core/config.py#L88).

phorward (approved):

sveneberth (dismissed):

phorward (approved):

PR #904, opened 2023-10-11
feat: Provide `Module`-specific `Logger`

phorward:
Introduces `log`-member to Modules, so logs can be filtered by module and are also printed so to the local formatter, using 'Module: message'. No need to import logging to all your ViUR modules anymore, just use `self.log`.

### DIFF
--- a/src/viur/core/logging.py
+++ b/src/viur/core/logging.py
@@ -110,6 +110,10 @@ class ViURLocalFormatter(logging.Formatter):
 
         record.pathname = pathname
 
+        # Show logger-name before message when not "root" logger
+        if record.name != "root":
+            record.msg = record.name + ": " + record.msg
+
         # Select colorization mode
         match (os.getenv(f"VIUR_LOGGING_COLORIZATION") or "FULL").upper():
             case "DECENT":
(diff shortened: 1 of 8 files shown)


src/viur/core/logging.py
```diff
@@ -110,6 +110,10 @@ def format(self, record: logging.LogRecord) -> str:
 
         record.pathname = pathname
 
+        # Show logger-name before message when not "root" logger
+        if record.name != "root":
```
sveneberth, line 114: Sorry, but this condition is totally sh*t. This effects all other loggers. Not only the _module-loggers_. 

The logging in the viur-core is general _kinda hackish_. Using the root logger is a bad practise, because usually all loggers of other packages propagates to it. Which results in workaround like [this](https://github.com/viur-framework/viur-core/blob/926d42f9f3062d28bc826b5e43bce5216499f229/src/viur/core/logging.py#L145-L150). 

I have long wanted to rework the logging and work with sub loggers... Maybe this would be a good time to combine this. Give me a bit of time to evaluate and get this done.

sveneberth (changes requested):

src/viur/core/logging.py
phorward: Ok, let's put this thing on hold. You can also create a follow-up PR based on this, if it's easier.

sveneberth: I am working on #1080

phorward:
Replaced by #1080.

Issue #143, opened 2020-10-13
Implement relation depth parameter


sveneberth:
As decided in the ViUR-round at 2020-08-07 we need a parameter to set the depth of refKeys in relations.

PR #905, opened 2023-10-11
feat: Add `expand`-flag to `RelationalBone`

phorward:
Allows for specific renders to expand the values of a RelationalBone to its full dest-skeleton, and not only the specified refKeys.

This is useful to directly fetch entire data structures with just one request, and is useful for modern single-page-apps.

Example usage:
```python
class UserPortal(Skeleton):
    portal = RelationalBone(
        descr="Portal",
        kind="portal",
        expand=True,
        readOnly=True,
        format="$(dest.identifier)",
        refKeys=["key", "identifier", "name"],
        consistency=RelationalConsistency.CascadeDeletion,
    )
```

This expands portal to its full skeleton, including all descriptions, logos, marketing-huiuiui-wheeee-images and texts without spamming the datastore, by extending all bones to refKeys. The userportal entity still only stores the values in refKeys.

### DIFF
--- a/src/viur/core/render/json/default.py
+++ b/src/viur/core/render/json/default.py
@@ -1,7 +1,7 @@
 import json
 from enum import Enum
 
-from viur.core import bones, utils, db, current
+from viur.core import bones, utils, db, current, skeleton
 from viur.core.skeleton import SkeletonInstance
 from viur.core.i18n import translate
 from viur.core.config import conf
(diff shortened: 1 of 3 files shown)


src/viur/core/render/json/default.py
```diff
@@ -85,19 +85,29 @@ def renderSingleBoneValue(self, value: Any,
...
+            if not isinstance(value, dict):
+                return None  # FIXME: Shall I raise?
+
+            dest = value["dest"]
+            rel = value["rel"]
+
+            if bone.expand:
+                skel = skeleton.skeletonByKind(bone.kind)()
+                if not skel.fromDB(dest["key"]):
+                    return None  # FIXME: Shall I raise?
```
sveneberth, line 97: Yeah, of course! Otherwise there wouldn't be a difference between no and invalid data!

sveneberth:
Do we need to think about recursion in case of nested relBones? I think this correlates in some case with #143. I mean in some cases you want to `expand` the _sub_ relBone, in others not. 
Imagine these relational models: `A -> B -> C` and `D -> B -> C`. The cascade of the RelationalBone in `A` (to `B`, which as a relation to `C`) should completely expand `B` and `C`. But the RelationalBone in `D` should only expand `B`, but not `C`.

phorward:
Alternative replacement provided by #1193, issued #1101.

PR #906, opened 2023-10-12
feat: Add conf var to disable describe hache

ArneGudermann:
In some projects we need to claculate the admininfo based on a user so we can't cache it.


phorward (changes requested):
Hello @ArneGudermann,

this solution is not good. I think this is only necessary to some modules in your project.
What about solving it via the `_cached_description`-member of Module? Say

```python
class MyModule(Module):
    _cached_description = False   # disable caching
```
This would be an adequate solution.

phorward (approved):

PR #910, opened 2023-10-13
feat: `utils.is_prefix`-function

phorward:
This function is required for developments coming soon.

### DIFF
--- a/src/viur/core/utils.py
+++ b/src/viur/core/utils.py
@@ -273,6 +273,29 @@ def normalizeKey(key: Union[None, 'db.KeyClass']) -> Union[None, 'db.KeyClass']:
     return db.Key(key.kind, key.id_or_name, parent=parent)
 
 
+def prefixes(name: str, prefix: str, delimiter: str = ".") -> bool:
+    """
+    Utility function to check if a given name matches a prefix,
+    which defines a specialization, delimited by `delimiter`.
+
+    In ViUR, modules, bones, renders, etc. provide a kind or handler
+    to classify or subclassify the specific object. To securitly
+    check for a specific type, it is either required to ask for the
+    exact type or if its prefixed by a path delimited normally by
+    dots.
+
+    Example:
+
+    .. code-block:: python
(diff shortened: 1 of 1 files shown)


src/viur/core/utils.py
```diff
@@ -273,6 +273,29 @@ def normalizeKey(key: Union[None, 'db.KeyClass']) -> Union[None, 'db.KeyClass']:
     return db.Key(key.kind, key.id_or_name, parent=parent)
 
 
+def prefixes(name: str, prefix: str, delimiter: str = ".") -> bool:
```
sveneberth, line 276: I don't find the name very meaningful. `prefixes` sounds like prefixes are being generated here -- even though the docstring explains that, of course.

technically it should be `is_equal_or_prefix`. But maybe `is_prefix` is already enough. 

```suggestion
def is_prefix(name: str, prefix: str, delimiter: str = ".") -> bool:
```

sveneberth (changes requested):

src/viur/core/utils.py
```diff
@@ -273,6 +273,29 @@ def normalizeKey(key: Union[None, 'db.KeyClass']) -> Union[None, 'db.KeyClass']:
...
+    dots.
+
+    Example:
+
+    .. code-block:: python
+        handler = "tree.file.special"
+        utils.prefixes(handler, "tree")  # True
+        utils.prefixes(handler, "tree.node")  # False
+        utils.prefixes(handler, "tree.file")  # True
+        utils.prefixes(handler, "tree.file.special")  # True
```
sveneberth, line 294: The examples must then be adapted:

```suggestion
        utils.is_prefix(handler, "tree")  # True
        utils.is_prefix(handler, "tree.node")  # False
        utils.is_prefix(handler, "tree.file")  # True
        utils.is_prefix(handler, "tree.file.special")  # True
```

sveneberth (changes requested):

sveneberth (approved):

PR #936, opened 2023-11-10
fix: `access` in method description must not be a generator object

sveneberth:
A generator object is not JSON serializeable and `dumpConfig` fails

### DIFF
--- a/src/viur/core/module.py
+++ b/src/viur/core/module.py
@@ -324,7 +324,7 @@ class Method:
             ret["skey"] = self.skey["name"]
 
         if self.access:
-            ret["access"] = (str(access) for access in self.access["access"])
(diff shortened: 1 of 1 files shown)


src/viur/core/module.py
```diff
@@ -324,7 +324,7 @@ def describe(self) -> dict:
             ret["skey"] = self.skey["name"]
 
         if self.access:
-            ret["access"] = (str(access) for access in self.access["access"])
+            ret["access"] = [str(access) for access in self.access["access"]]
```
phorward, line 327:
```suggestion
            ret["access"] = [str(access) for access in self.access["access"]]  # must be a list to be JSON-serializable
```

phorward (approved):

PR #937, opened 2023-11-15
feat: use `CustomJsonEncoder` class in `json.dumps` call

sveneberth:
I want to translate our repo names, the `CustomJsonEncoder` can resolve the `translate` objects. Just in case I've added this class where it could make sense to have translated objects too.


phorward (approved):
It's okay for now. Generally, I would like to replace this stupid CustomEncoder by a general way of how ViUR handles JSON. This will hopefully come soon, when there's more time again.

Issue #932, opened 2023-11-08
feat: Implement `minLength` parameter in `StringBone`


sveneberth:
That would be nice to ensure complete GTINs, ISBNs and so one are entered.

PR #940, opened 2023-11-21
feat: Add `min_length` to `StringBone`

ArneGudermann:
Resolve: https://github.com/viur-framework/viur-core/issues/932

### DIFF
--- a/src/viur/core/bones/string.py
+++ b/src/viur/core/bones/string.py
@@ -16,6 +16,7 @@ class StringBone(BaseBone):
         *,
         caseSensitive: bool = True,
         maxLength: int | None = 254,
+        min_length: int | None = None,
         natural_sorting: bool | Callable = False,
         **kwargs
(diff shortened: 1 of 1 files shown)


sveneberth (approved):
Nice, so far this LGTM. Thank you.

Not sure if this fits into this PR. But we have now a camelcased and a snakecased *length argument. Can we unify them?

I think
```py
        max_length: int | None = 254,
```
and sth like 
```py
if "maxLength" in kwargs:
  warning.warn("maxLength is deprecated, please use max_length")
  max_length = kwargs.pop("maxLength")
```
could be a backward compatible solution

and maxLength

ArneGudermann:
Yes, I noticed that too, I'm doing a new PR for it.

phorward:
Renaming perfomed in #942

phorward (approved):

PR #941, opened 2023-11-21
fix: Invalid `maxLength` check in `StringBone`

ArneGudermann:
This PR fixes the problem that the value was always shortened to the maximum length before it was checked for the maximum length and so you simply lose data without getting an error message

### DIFF
--- a/src/viur/core/bones/string.py
+++ b/src/viur/core/bones/string.py
@@ -114,7 +114,7 @@ class StringBone(BaseBone):
         this bone, otherwise the empty value and an error-message.
         """
         if not (err := self.isInvalid(value)):
-            return utils.escapeString(value, self.maxLength), None
+            return utils.escapeString(value, None), None
         return self.getEmptyValue(), [ReadFromClientError(ReadFromClientErrorSeverity.Invalid, err)]
 
     def buildDBFilter(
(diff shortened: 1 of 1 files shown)


sveneberth (commented):
I see your intention, but both solution have their right for existence.

Imagine you want to send data to an external API that has a length limit. You set your bone to this limit to ensure that it will not exceed it. But then the user enters a word with the maximum length and umlauts. To prevent the API from rejecting this value, you have to guess a suitable lower limit for your bone, unquote the value (btw. why do we quote values for the database, it might be time to restart this discussion :see_no_evil:) or slice it.

phorward:
Let's discuss this on friday.

phorward:
I would like to first merge #942, then #940  and afterwards update this PR for discussion on friday. Is that ok?

src/viur/core/bones/string.py
```diff
@@ -104,15 +104,17 @@ def isInvalid(self, value):
...
     def singleValueFromClient(self, value, skel, bone_name, client_data):
-        value = utils.escapeString(value, self.maxLength)
-
+        """
+        Returns None and the escaped value if the value would be valid for
+        this bone, otherwise the empty value and an error-message.
+        """
         if not (err := self.isInvalid(value)):
-            return value, None
+            return utils.escapeString(value, self.maxLength), None
```
phorward, line 117:
```suggestion
            return utils.escapeString(value, None), None
```

phorward (changes requested):

sveneberth (approved):

PR #949, opened 2023-11-28
fix: Readd is_dev_server check for secure cookies

ArneGudermann:
Revert https://github.com/viur-framework/viur-core/pull/931
If you set the cookie to `secure` but use [requests](https://github.com/psf/requests) to establish a connection via `http`, the python [http](https://github.com/python/cpython/tree/main/Lib/http) lib throws the cookie away.

Code:
https://github.com/python/cpython/blob/05a370abd6cdfe4b54be60b3b911f3a441026bb2/Lib/http/cookiejar.py#L1135-L1139

### DIFF
--- a/src/viur/core/session.py
+++ b/src/viur/core/session.py
@@ -117,7 +117,7 @@ class Session:
             "Path=/",
             "HttpOnly",
             f"SameSite={self.same_site}" if self.same_site else None,
-            "Secure",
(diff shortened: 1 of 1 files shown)


sveneberth (changes requested):
While this fix the problem with `requests`, this doesn't address the problem from #931.
I created https://github.com/viur-framework/viur-core/pull/961 as alternative solution

PR #951, opened 2023-11-29
feat: Re-add `unescape_string`-function

ArneGudermann:


phorward (dismissed):
Yay! Finally, this function is re-added again, after it was [mysteriously removed a few years ago for very important and perfectly understandable reasons](https://github.com/viur-framework/viur-core/commit/438fd45a54acc078d10ecfcf27451ed259b98f44)............................................................................................................................................................................................................................................................................................................................................................................

phorward:
> I'm not sure, why we need this `max_length` part here in this method too?

@ArneGudermann this function is a utility-function not used in viur-core yet. Therefore, it is a valid question.

ArneGudermann:
I wanted it to be an exact counterpart of the `unescape_string` function. But I can also just take it out

phorward:
> I wanted it to be an exact counterpart of the `unescape_string` function. But I can also just take it out. Or should we just set the default limit to unlimeted?

no, the question is, is it useful?

ArneGudermann:
I don't have an useful expample for it now.

sveneberth:
> Or should we just set the default limit to unlimeted?

I would definitely find that much better.

---

And then I no longer really see the advantage in writing
```py
unescape_string(value, 128)
```
instead of
```py
unescape_string(value)[:128]
```

---

I would simply remove it. The string length will never increase with this method — in contrast to the escape method.

phorward:
#992 is a replacement for this PR

phorward:
Replaced by #992

Issue #933, opened 2023-11-08
`@property` in module causes problems during warmup


sveneberth:
I've `@property`-ies in my modules. During warmup, they are accessed by [`Module._update_methods()`](https://github.com/viur-framework/viur-core/blob/008c68692ed3fdd5bfcf3ee68bfbc03670f7b87c/src/viur/core/module.py#L517C11-L517C11). But at this point the module instance is not initialized and attribute access on attributes set in `__init__` fails. Therefore, the entire warmup (of this module) fails.

---

This MRE makes no real logical sense, but it's (not) working :)

```py
class Example(List):

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self._items = ["foo", "bar"]

    @property
    def first_item(self):
        return self._items[0]
```

```py
Traceback (most recent call last):
  ...
  File "/.../deploy/main.py", line 193, in <module>
    app = core.setup(modules, render)
  File "/.../deploy/viur/core/__init__.py", line 253, in setup
    conf.main_app = buildApp(modules, render, default)
  File "/.../deploy/viur/core/__init__.py", line 194, in buildApp
    module_instance = module_cls(
  File "/.../deploy/modules/example.py", line 19, in __init__
    super().__init__(*args, **kwargs)
  File "/.../deploy/viur/core/module.py", line 515, in __init__
    self._update_methods()
  File "/.../deploy/viur/core/module.py", line 529, in _update_methods
    prop = getattr(self, key)
  File "/.../deploy/modules/example.py", line 24, in first_item
    return self._items[0]
AttributeError: 'Example' object has no attribute '_items'
```

## Workaround (but not nice)
```py
    @property
    def first_item(self):
        if conf.main_app is None:
            return
        return self._items[0]
```

sveneberth:
still a problem ...

phorward:
Fixed by #1020

PR #952, opened 2023-11-29
fix: Modules with property crash on `_update_methods`

ArneGudermann:
Resolves: #933


sveneberth (approved):
This looks simple and logical. I approve this without testing it. I hope I don't regret it later :grin:

phorward (approved):
Let's give it a try.

Issue #790, opened 2023-07-18
Extend `DateBone` to allow for time values with microseconds


phorward:
There are several issues that confuse or point out a todo for a `DateBone` supporting microseconds. There was also a request from a customer project. Therefore this issue is created.

              > > Okay. I was wrong. The datastore client doesn't throw it away, but the core in the `FileBone` does:
> > https://github.com/viur-framework/viur-core/blob/19f1351ffd5e9fc4269235289b78b48fa2481f71/core/bones/date.py#L242
> > 
> > So I don't understand why we have to do this here again.
> 
> `FileBone`? I dont now why throw it away. I implement the old way. May be it's historical. The datastore support microseconds in the timestamp. https://cloud.google.com/datastore/docs/reference/data/rest/Shared.Types/Value

The reason for setting the microseconds to 0 is, that additional, project-specific program code might rely on a datetime-object which is the exact time to seconds, without microseconds. In the past, I had a use-case where this was a big problem and why it was added. microseconds should only be set when the DateBone is in a yet not existing time-with-microseconds-mode.

I would like to take this PR for now as is. We can later change the behavior to respect microseconds when this is configured so.

_Originally posted by @phorward in https://github.com/viur-framework/viur-core/issues/785#issuecomment-1640241907_

phorward:
In #954, I've seen [this comment](https://github.com/viur-framework/viur-core/pull/954#issuecomment-1848386839):

> In the viur-meeting from Dec 8, 2023, we decided to resolve this issue differently, and split the `DateBone` into several bones:
> 
> - `DateBone` - used for dates
> - `TimeBone` - used for times
> - `DateTimeBone(microseconds=True)` - used for date and time
> - `DeltaBone` - used for durations
> 
> 
> This requires more investigation and work, may be part of viur-core > 3.6.

Therefore... we should discuss either about it or leave it for now.

phorward:
We implement DateBone with a microsecond=True.

PR #954, opened 2023-11-29
feat: Extend DateBone to allow for time  with microseconds

ArneGudermann:
Resolves : #790

### DIFF
--- a/src/viur/core/bones/date.py
+++ b/src/viur/core/bones/date.py
@@ -21,6 +21,7 @@ class DateBone(BaseBone):
     :param time: If True, the bone will contain time information.
     :param localize: If True, the user's timezone is assumed for input and output. This is only valid if both 'date'
           and 'time' are set to True. By default, UTC time is used.
+    :param bool allow_microsecond If True the bones stores the date with microsecond.
     """
     # FIXME: the class has no parameters; merge with __init__
     type = "date"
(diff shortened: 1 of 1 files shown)


phorward (changes requested):
Hello @ArneGudermann, thanks for working on this!

I'm a little unhappy with the naming convention of the new configuration. `allow_microsecond` is only an additional configuration of the `time=True` case, therefore it should only be allowed in combination with `time=True`. This could be done with an Enum, for example, so `time=True` becomes a backward compatibility only. 

Let's think further about it.... why might time=True not always store microseconds, as the information is there, and microsecond is just set 0 when not provided (e.g. in old-style VI date-time-formattings `dd.mm.yyyy hh:mm:ss`.

Or, maybe split `DateBone` into `DateBone`, `TimeBone` and `DateTimeBone`? 

Let's discuss about this next Friday.

sveneberth:
> Or, maybe split `DateBone` into `DateBone`, `TimeBone` and `DateTimeBone`?

that sounds like proven concepts, as in SQL.

---

By the way, I also have a use case for a (computable) `TimeDeltaBone` :)

phorward:
> By the way, I also have a use case for a (computable) `TimeDeltaBone` :)

Hahah, I've created a `DeltaBone` in a customer's project, decades ago.. that's exactly this thing.

phorward:
In the viur-meeting from Dec 8, 2023, we decided to resolve this issue differently, and split the `DateBone` into several bones:

- `DateBone` - used for dates
- `TimeBone` - used for times
- `DateTimeBone(microseconds=True)` - used for date and time
- `DeltaBone` - used for durations

This requires more investigation and work, may be part of viur-core > 3.6.

PR #960, opened 2023-12-01
fix: Get rid of the catch all index behavior in vi renderer

### LINKED ISSUE #787: get rid of the catch all index behavior at least in api calls
The Call localhost:8080/vi/news/any_unknown_function_name  returns the same result as localhost:8080/vi/news/list or localhost:8080/vi/news ... In my opinion it should return an 404 error.

sveneberth:
Only `/vi/` should redirect, in case of `/vi/usre` (typo) it should raise an error (404) instead of redirection.

Resolves #787 
Alternative to #955

### DIFF
--- a/src/viur/core/render/vi/__init__.py
+++ b/src/viur/core/render/vi/__init__.py
@@ -141,6 +141,8 @@ def canAccess(*args, **kwargs) -> bool:
 
 @exposed
 def index(*args, **kwargs):
+    if args or kwargs:
+        raise errors.NotFound
     if not conf.instance.project_base_path.joinpath("vi", "main.html").exists():
         raise errors.NotFound()
     if conf.instance.is_dev_server or current.request.get().isSSLConnection:
--- a/src/viur/core/render/vi/__init__.py
+++ b/src/viur/core/render/vi/__init__.py
@@ -142,7 +142,7 @@ def canAccess(*args, **kwargs) -> bool:
 @exposed
 def index(*args, **kwargs):
     if args or kwargs:
-        raise errors.NotFound
+        raise errors.NotFound()
     if not conf.instance.project_base_path.joinpath("vi", "main.html").exists():
         raise errors.NotFound()
     if conf.instance.is_dev_server or current.request.get().isSSLConnection:


src/viur/core/render/vi/__init__.py
```diff
@@ -141,6 +141,8 @@ def canAccess(*args, **kwargs) -> bool:
 
 @exposed
 def index(*args, **kwargs):
+    if args or kwargs:
+        raise errors.NotFound
```
phorward, line 145: Isn't it required to instantiate this?
```suggestion
        raise errors.NotFound()
```

src/viur/core/render/vi/__init__.py
```diff
@@ -141,6 +141,8 @@ def canAccess(*args, **kwargs) -> bool:
...
     if not conf.instance.project_base_path.joinpath("vi", "main.html").exists():
```
phorward, line 146: Alternative solution:
```suggestion
    if args or kwargs or not conf.instance.project_base_path.joinpath("vi", "main.html").exists():
```

phorward (changes requested):

src/viur/core/render/vi/__init__.py
sveneberth: That's implicit done.


> The sole argument to [raise](https://docs.python.org/3.11/reference/simple_stmts.html#raise) indicates the exception to be raised. This must be either an exception instance or an exception class (a class that derives from [BaseException](https://docs.python.org/3.11/library/exceptions.html#BaseException), such as [Exception](https://docs.python.org/3.11/library/exceptions.html#Exception) or one of its subclasses). If an exception class is passed, it will be implicitly instantiated by calling its constructor with no arguments:
> 
> ```python
> raise ValueError  # shorthand for 'raise ValueError()'
> ```

_~ from the [Python docs](https://docs.python.org/3.11/tutorial/errors.html?highlight=shorthand#raising-exceptions)_

src/viur/core/render/vi/__init__.py
sveneberth: But then you lose the information about the reason that caused this exception. If the trace exception conf option is activated, you can see the line of the exception and know whether it is an invalid path or whether there is no vi in your deploy folder.

phorward (dismissed):

src/viur/core/render/vi/__init__.py
phorward: OK thanks for the info, but to me it always looks wrong. The next time I come over this line I would correct it. Therefore, we just fix this.

PR #963, opened 2023-12-01
fix: Fix `CallableTask` `TaskVacuumRelations`

sveneberth:
It was totally broken and got some refactoring.

### DIFF
--- a/src/viur/core/skeleton.py
+++ b/src/viur/core/skeleton.py
@@ -1451,41 +1451,30 @@ class RebuildSearchIndex(QueryIter):
     @classmethod
     def handleFinish(cls, totalCount: int, customData: Dict[str, str]):
         QueryIter.handleFinish(totalCount, customData)
+        if not customData["notify"]:
+            return
         try:
-            if customData["notify"]:
-                txt = f"Rebuild search index finished for {customData['module']}\n\n" \
-                      f"ViUR finished to rebuild the search index for module {customData['module']}.\n" \
-                      f"{totalCount} records updated in total on this kind."
-                email.sendEMail(dests=customData["notify"], stringTemplate=txt, skel=None)
-        except:  # OverQuota, whatever
-            pass
+            txt = f"Rebuild search index finished for {customData['module']}\n\n" \
+                  f"ViUR finished to rebuild the search index for module {customData['module']}.\n" \
+                  f"{totalCount} records updated in total on this kind."
+            email.sendEMail(dests=customData["notify"], stringTemplate=txt, skel=None)
+        except Exception as exc:  # noqa; OverQuota, whatever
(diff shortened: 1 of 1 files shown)


ArneGudermann (dismissed):

src/viur/core/skeleton.py
```diff
@@ -1451,41 +1451,30 @@ def handleEntry(cls, skel: SkeletonInstance, customData: Dict[str, str]):
...
+            return
         try:
-            if customData["notify"]:
-                txt = f"Rebuild search index finished for {customData['module']}\n\n" \
-                      f"ViUR finished to rebuild the search index for module {customData['module']}.\n" \
-                      f"{totalCount} records updated in total on this kind."
-                email.sendEMail(dests=customData["notify"], stringTemplate=txt, skel=None)
-        except:  # OverQuota, whatever
-            pass
+            txt = f"Rebuild search index finished for {customData['module']}\n\n" \
```
phorward, line 1457: I would like to include the project_id here as well!
```suggestion
            txt = f"""{conf["viur.instance.project_id"]}: Rebuild search index finished for {customData["module"]}\n\n""" \
```

src/viur/core/skeleton.py
```diff
@@ -1496,56 +1485,51 @@ def execute(self, module, *args, **kwargs):
...
+    elif notify:
         try:
-            if notify:
-                txt = ("Vaccum Relations finished for %s\n\n" +
-                       "ViUR finished to vaccum viur-relations.\n" +
-                       "%d records processed, %d entries removed") % (module, newTotalCount, newRemovedCount)
-                email.sendEMail(dests=[notify], stringTemplate=txt, skel=None)
-        except:  # OverQuota, whatever
-            pass
+            txt = (f"Vacuum Relations finished for {module}\n\n"
```
phorward, line 1526:
```suggestion
            txt = (f"""{conf["viur.instance.project_id"]}: Vacuum relations finished for {module}\n\n"""
```

phorward (changes requested):
Can we use this PR to add the project_id as well to the given tasks?
I'm often working on the same project with different instances, this will help to distinguish the emails better.

sveneberth:
> Can we use this PR to add the project_id as well to the given tasks email notification subjects? I'm often working on the same project with different instances, this will help to distinguish the emails better.

Hmm. I don't really have anything against that. But that's already in the sender address. Isn't that enough for you?

phorward:
> Hmm. I don't really have anything against that. But that's already in the sender address. Isn't that enough for you?

I really don't know what you mean...
image
..especially while this project uses same SIB credentials (the emails came from two GAE projects)

phorward (approved):

PR #964, opened 2023-12-01
fix: `exclude_from_indexes` has to be a set

sveneberth:
Otherwise this can lead to problems like https://github.com/viur-framework/viur-datastore/issues/43

### DIFF
--- a/src/viur/core/cache.py
+++ b/src/viur/core/cache.py
@@ -156,7 +156,7 @@ def wrapCallable(f, urls: List[str], userSensitive: int, languageSensitive: bool
         dbEntity["path"] = path
         dbEntity["content-type"] = currReq.response.headers['Content-Type']
         dbEntity["accessedEntries"] = list(accessedEntries)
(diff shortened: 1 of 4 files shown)


ArneGudermann (approved):

src/viur/core/cache.py
```diff
@@ -156,7 +156,7 @@ def wrapF(self, *args, **kwargs) -> Union[str, bytes]:
         dbEntity["path"] = path
         dbEntity["content-type"] = currReq.response.headers['Content-Type']
         dbEntity["accessedEntries"] = list(accessedEntries)
-        dbEntity.exclude_from_indexes = ["data", "content-type"]  # We can save 2 DB-Writs :)
+        dbEntity.exclude_from_indexes = {"data", "content-type"}  # We can save 2 DB-Writs :)
```
phorward, line 159:
```suggestion
        dbEntity.exclude_from_indexes = {"data", "content-type"}  # save two DB-writes.
```

phorward (changes requested):

PR #967, opened 2023-12-05
Implement `InstancedModule`

sveneberth:
InstancedModule is a base class for modules, which can be passed as instance to the viur-core.

Normally, the viur-core expects classes from modules and instantiates them during setup.
However, there are cases where you want to pass already instantiated
classes, i.e. module instances, to the core. For example, to integrate
plugins. This is where this base class comes into play. If it is included
as an ADDITIONAL base class to the module class, the viur-core accepts an
instance of this class. The class itself is then no longer
instantiated in the core.

In order to preserve the well-known ViUR module concept, this class
overwrites the __init__ method so that it behaves like a classic object
constructor. The __call__ method then takes on the role of the
conventional __init__ module.
However, as each render requires its own module instance, as the render
itself is then referenced to this instance via the "render" attribute,
several instances are required. Therefore, a copy of the module instance
is created using the internal method `_viur_clone`. This can be changed
if necessary. A shadow-copy is created here by default.

Here an example:

```py
from viur.core.prototypes import List
from viur.core.prototypes.instanced_module import InstancedModule


class MyModule(InstancedModule, List):
    def __init__(self, configuration):
        self.configuration = configuration


my_module = MyModule(
    configuration=...
)
```

As you can see, the configuration parameter is set from outside and not
in the class. This may seem a bit strange here, but now imagine that the
class definition `MyModule` is not within the same file but in an imported
pip package. Then this instantiation is much easier than using a subclass.


src/viur/core/email.py
```diff
@@ -429,7 +429,7 @@ def check_sib_quota() -> None:
         entity["credits"] = credits
         entity["email"] = data["email"]
 
-        thresholds = sorted(conf.email.sendinblue_thresholds, reverse=True)
+        thresholds = sorted(conf.email_sendinblue_thresholds, reverse=True)
```
phorward, line 432: This looks invalid, as it was recently replaced by #989

phorward (changes requested):

src/viur/core/email.py
sveneberth: Oh, yes that shouldn't be there...

src/viur/core/prototypes/instanced_module.py
```diff
@@ -0,0 +1,77 @@
...
+        """
+        object.__init__(self)
+
+    def __call__(self, moduleName: str, modulePath: str, *args, **kwargs) -> "InstancedModule":
+        """Do the instance initialisation
+
+        Take the role of Module.__init__ (and call it).
+        """
+        is_default_renderer = modulePath == f"/{moduleName}"
+        if is_default_renderer:
```
ArneGudermann, line 63:
```suggestion
        if modulePath == f"/{moduleName}":
```

ArneGudermann (changes requested):
Hey @sveneberth thank you for this PR. I have only one suggestion and one Question for now. Did you notice any performance differences when you mounted a module via `InstancedModule`?

sveneberth:
> Did you notice any performance differences when you mounted a module via `InstancedModule`?

I have not done any benchmarking in this direction, but I have not noticed any subjective changes in performance. In the end, it would only affect the warmup. Everything else behaves the same later on.

phorward (approved):

ArneGudermann (approved):

PR #969, opened 2023-12-06
feat: Improve translations & provide `Translation` module

sveneberth:
- Add module + skeleton to manage translations
- Don't copy translated values for alias language in memory multiple times
- Fix some bugs, e.g.:
  - inconsistent variable substitution
  - problem with some argument combinations in jinja translation extension
- Add documentation (need refactoring, grammar fixes)
- Add config option and logic to auto add needed and missing translation
- Add `force_lang` parameter as alternative to the current request lang


src/viur/core/i18n.py
```diff
@@ -1,21 +1,109 @@
...
+from pathlib import Path
+
+from viur.core import current, db, languages, tasks
 from viur.core.config import conf
-from viur.core import db, utils, languages, current
 
 systemTranslations = {}
+"""Memory storage for translation methods"""
+
+KINDNAME = "viur-translations"
```
phorward, line 89: I would like to prefer singular names `viur-translation` in this case.
Yes I know, that other modules use plural, but we should start at a point.
```suggestion
KINDNAME = "viur-translation"
```

src/viur/core/i18n.py
```diff
@@ -139,71 +295,175 @@ def parse(self, parser):
...
-        for lang, translation in tr["translations"].items():
-            trDict[lang] = translation
-            if lang in invertMap:
-                for v in invertMap[lang]:
-                    trDict[v] = translation
-
-        systemTranslations[tr["key"]] = trDict
+    # TODO: iter() would be more memory efficient, but unfortunately takes much longer than run()
+    # for entity in db.Query(KINDNAME).iter():
+    for entity in db.Query(KINDNAME).run(10_000):
```
phorward, line 355: Hmmm... does it make sense to make this configurable?

src/viur/core/i18n.py
```diff
@@ -139,71 +295,175 @@ def parse(self, parser):
...
+        if "tr_key" not in entity:
+            logging.error(f"translations entity {entity.key} has no tr_key set --> Call migration")
```
phorward, line 357:
```suggestion
            logging.warning(f"translations entity {entity.key} has no tr_key set --> Call migration")
```

src/viur/core/i18n.py
```diff
@@ -139,71 +295,175 @@ def parse(self, parser):
...
+        "_default_text_": default_text or None,
+    }
+
+
+@tasks.CallDeferred
+@tasks.retry_n_times(20)
+def migrate_translation(
+    key: db.Key,
+) -> None:
+    """Migrate entities in the required.
```
phorward, line 441: In the required what??

src/viur/core/modules/translation.py
```diff
@@ -0,0 +1,173 @@
...
+
+    @classmethod
+    def toDB(cls, skelValues: SkeletonInstance, **kwargs) -> db.Key:
+        # Ensure we have only lowercase keys
+        skelValues["tr_key"] = skelValues["tr_key"].lower()
+        return super().toDB(skelValues, **kwargs)
+
+    @classmethod
+    def preProcessSerializedData(cls, skelValues: SkeletonInstance, entity: db.Entity) -> db.Entity:
+        # Backward-compatibility: re-add the key for viur-core < v3.6
```
phorward, line 119:
```suggestion
        # Backward-compatibility: re-add the key for viur-core < v3.6
        # TODO: Remove in ViUR4
```

phorward (changes requested):
Hello @sveneberth, I'm mostly fine with this, can you take a look into the comments?

src/viur/core/i18n.py
sveneberth: In general, I agree with you. But that would mean that existing data would have to be copied to the new kind.  I would therefore not want to change that.

Or we could change it and then recommend monkeypatching the value for existing projects that already have translations in the DB:
```py
viur.core.i18n.KINDNAME = "viur-translations"
```
But we should test this first to see if it works in time during warmup.

src/viur/core/i18n.py
sveneberth: Actually, the value doesn't really make sense. It could also be sys.max or 999999999 if you like.
The method simply needs a high value to be _"unlimited"_.

src/viur/core/i18n.py
sveneberth:
```suggestion
    """Migrate entities, if required.
```

src/viur/core/i18n.py
phorward: Oh... okay then miss this out.

sveneberth:
:partying_face:

Issue #630, opened 2023-01-24
Refactor `Skeleton` interface


phorward:
Ok, this will become a huger and recently updated issue with several things pointing to several things in VIURs current `Skeleton` interface. The reason for this issue is, that the current interface is uneven, messy and misleading in several ways.

## CRUD-functions

It's called CRUD: Create, Read, Update, Delete.

### Current state

For Skeletons, this is:
- Create: `toDB()` with `skel["key"]` set to `None`
- Read: `fromDB()` 
- Update: `toDB()` with a `skel["key"]` set
- Delete: `delete()` 
- 
Okay - why is it fromDB(), toDB() but not... deleteDB()? This is uneven and messy.

### Proposed state

- Create: `write()` with `skel["key"]` set to `None`
- Read: `read()`
- Update: `write()` with a `skel["key"]` set
- Delete: `delete()`

## Hooks

In many situations, it is necessary to do some pre- or post-processing of data before it is written, read or deleted.

### Current state

- `toDB()`:
   - `preProcessSerializedData()` - Do something on the already serialized `db.Entity` without using the bones anymore (mostly useless!)
   - `preProcessBlobLocks()` - Use is only relevant to blobs?
   - `postSavedHandler()` oh nice, I can do something *after* the skeleton has been saved. Why isn't it called `postToDBHandler()`? 
   - Ok how to use the bones preliminary before writing data? Right, you override `toDB()`, do your preparation and afterwards make a super call. This is the current way to go.
- `fromDB()`: No chance to do something here. Wow.
- `delete()`:
  - `postDeletedHandler()` is a hook on a `SkeletonInstance` which has been previously deleted.

### Proposed state

It should be equivalent to module's hooks functionalities and naming:

- `write()`:
  - `on_write()` - do something before saving with the skeleton
  - `serialize()` - do something before saving with the `db.Entity`
    default/super should be a `Skeleton.serialize()`, which doesn't exist as well!
- `read()`:
  - `unserialize()` - do something after reading with the `db.Entity`
    `Skeleton.unserialize()` doesn't exist anymore, as bones are unserialized on-demand. Anyway, the hook should exist for override.
  - `on_read()` - do something after reading with the skeleton
- `delete()`:
  - `on_delete()` - do something after deleting the dataset. This is mostly equal to `postDeletedHandler()`, except the signature.

PR #972, opened 2023-12-07
feat: add hooks on_item_added_and_edited and `on_item_added_and_edited_and_deleted`

sveneberth:
I've often methods / actions that should be called after and skeleton has changes and exist (`on_item_added_and_edited`) or a skeleton has changes in any way (`on_item_added_and_edited_and_deleted`). 
With these methods it's very easy to call these actions.

--- 

I know the name is long, but expressive. And to be honest, I don't have a better one. `on_changed` would be okay for all three cases, but how should the other one be called?

### DIFF
--- a/src/viur/core/prototypes/list.py
+++ b/src/viur/core/prototypes/list.py
@@ -500,6 +500,8 @@ class List(SkelModule):
         flushCache(kind=skel.kindName)
         if user := current.user.get():
             logging.info("User: %s (%s)" % (user["name"], user["key"]))
+        self.on_item_added_and_edited(skel)
+        self.on_item_added_and_edited_and_deleted(skel)
 
     def onEdit(self, skel: SkeletonInstance):
         """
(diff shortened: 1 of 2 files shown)


ArneGudermann (approved):

phorward:
Hello @sveneberth,

Thanks for creating this pull request, and pointing out the need for such hooks. Nevertheless, I am not convinced by this implementation, as I find that this concept collides with the envisaged "ActionSkels"-paradigm we already talked about, but also with ideas regarding an improvement of the model (Skeleton-)level, see #630.

You could also think about an event bus that you can hook into for certain events. However, this all needs to be planned properly.

I will block this pull request for now, as I don't see any more hooks for actions in the VIUR system and especially the prototypes, but would rather reduce them and replace them with a more general, leaner and universal solution that fits all demans.

Please respect this for the time being. We can discuss about this in the viur-meeting tomorrow, but I see much more general work required first, to achieve this issue.

akelch (approved):

phorward:
Closing this for now, as it is not relevant and should be resolved differently.

PR #975, opened 2023-12-07
feat: Add `File.read()` method

ArneGudermann:
This PR add a read Method for the `File` module so you can easy read files  from the Cloudstore.

### DIFF
--- a/src/viur/core/modules/file.py
+++ b/src/viur/core/modules/file.py
@@ -1,8 +1,7 @@
 import base64
-import email.header
 import html
+import io
 import json
-import logging
 import string
 from base64 import urlsafe_b64decode
 from datetime import datetime, timedelta
@@ -11,6 +10,7 @@ from quopri import decodestring
 from typing import Any, List, Tuple, Union
 from urllib.request import urlopen
 
+import email.header
 import google.auth
 from PIL import Image, ImageCms
 from google.auth import compute_engine
@@ -18,9 +18,10 @@ from google.auth.transport import requests
 from google.cloud import iam_credentials_v1, storage
 from google.oauth2.service_account import Credentials as ServiceAccountCredentials
 
+import logging
 from viur.core import conf, current, db, errors, utils
-from viur.core.decorators import *
 from viur.core.bones import BaseBone, BooleanBone, KeyBone, NumericBone, StringBone
+from viur.core.decorators import *
 from viur.core.prototypes.tree import SkelType, Tree, TreeSkel
 from viur.core.skeleton import SkeletonInstance, skeletonByKind
 from viur.core.tasks import CallDeferred, DeleteEntitiesIter, PeriodicTask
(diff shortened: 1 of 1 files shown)


src/viur/core/modules/file.py
```diff
@@ -443,6 +444,34 @@ def write(self, filename: str, content: Any, mimetype: str = "text/plain", width
...
+            else:
+                path = f"""{skel["dlkey"]}/source/{skel["name"]}"""
+
+        blob = bucket.blob(path)
+        try:
+            data = blob.download_as_bytes()
+            return io.BytesIO(data)
+        except Exception as e:
+            logging.error(f"Could not read Bolb {e=}")
+            raise errors.NotFound()
```
phorward, line 473: IMHO this function is only used internally, should it really raise `NotFound`?
```suggestion
        data = blob.download_as_bytes()
        return io.BytesIO(data)
```

phorward (changes requested):

src/viur/core/modules/file.py
phorward: @ArneGudermann what about this proposal?

ArneGudermann: I missed it. Now ok ?

phorward (dismissed):

src/viur/core/modules/file.py
```diff
@@ -443,6 +444,31 @@ def write(self, filename: str, content: Any, mimetype: str = "text/plain", width
...
+        :param key: Key of the LeafSkel that contains the "dlkey" and the "name".
+        :param path: The path of the file in the Cloudstorage Bucket.
+
+        :return: Returns the file as a BytesIO buffer and the content-type
+        """
+        if not key and not path:
+            raise ValueError("Please provide a key or a path")
+        if key:
+            skel = self.viewSkel("leaf")
+            if not skel.fromDB(db.keyHelper(key, self.leafSkelCls().kindName)):
```
sveneberth, line 463:
```suggestion
            if not skel.fromDB(db.keyHelper(key, skel.kindName)):
```

sveneberth (changes requested):

phorward (approved):

src/viur/core/modules/file.py
```diff
@@ -443,6 +444,31 @@ def write(self, filename: str, content: Any, mimetype: str = "text/plain", width
 
         return skel.toDB()
 
+    def read(self, key: db.Key | int | str | None = None, path: str | None = None) -> Tuple[io.BytesIO, str]:
+        """
+        Read a file from the Cloudstorage.
```
sveneberth, line 449:
```suggestion
        Read a file from the Cloud Storage.
```

src/viur/core/modules/file.py
```diff
@@ -443,6 +444,31 @@ def write(self, filename: str, content: Any, mimetype: str = "text/plain", width
...
+
+        If a key and a path are provided, the key is preferred.
+        This means that the entry in the db is searched first and if this is not found, the path is used.
+
+        :param key: Key of the LeafSkel that contains the "dlkey" and the "name".
+        :param path: The path of the file in the Cloudstorage Bucket.
```
sveneberth, line 455:
```suggestion
        :param path: The path of the file in the Cloud Storage Bucket.
```

sveneberth (changes requested):
Please correct the [product name](https://cloud.google.com/storage). Otherwise people can think we mean a different service.

sveneberth (approved):

Issue #976, opened 2023-12-08
core 3.6 does not deliver vi/settings correctly


XeoN-GHMB:
https://github.com/viur-framework/viur-core/blob/48c210c03b6db4dc2d65fc314df7fca90075e953/src/viur/core/render/vi/__init__.py#L154
this should be something like this:

`fields = {key: values for key, values in conf.admin.items()}`

phorward:
Fixed by #977

PR #977, opened 2023-12-08
fix: `vi`-render returns wrong config (regression introduced by #833)

ArneGudermann:
Fix: #976
Regression fix for #833

### DIFF
--- a/src/viur/core/render/vi/__init__.py
+++ b/src/viur/core/render/vi/__init__.py
@@ -152,9 +152,7 @@ def index(*args, **kwargs):
 
 @exposed
 def get_settings():
-    fields = {key: values for key, values in conf.items()
-              if key.startswith("admin.")}
-
+    fields = {key: values for key, values in conf.admin.items()}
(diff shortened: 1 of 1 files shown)


sveneberth:
@ArneGudermann Thanks for this fix, it seems like i missed this part.

But the fix is inconsistent / not complete. Take a look at this:
https://github.com/viur-framework/viur-core/blob/482f7a8716578f5cf77e6128c80129d482569656/src/viur/core/render/vi/__init__.py#L90

In v3.5.x we have conf members like [`admin.color.primary`](https://github.com/viur-framework/viur-core/blob/48c210c03b6db4dc2d65fc314df7fca90075e953/src/viur/core/config.py#L69). 
Now it's [`color_primary`](https://github.com/viur-framework/viur-core/blob/482f7a8716578f5cf77e6128c80129d482569656/src/viur/core/config.py#L265). Therefor I added in #833 the replace. It think this is necessary here as well.

PR #979, opened 2023-12-08
fix: Admin-tool specific settings

phorward:
This is another fix for #977

### DIFF
--- a/src/viur/core/render/vi/__init__.py
+++ b/src/viur/core/render/vi/__init__.py
@@ -87,7 +87,7 @@ def dumpConfig():
         "modules": res,
         # "configuration": dict(conf.admin.items()), # TODO: this could be the short vision, if we use underscores
         "configuration": {
-            k.replace("_", "."): v for k, v in conf.admin.items()
+            {f"admin.{key}": values for key, values in conf.admin.items()}
         }
     }
     current.request.get().response.headers["Content-Type"] = "application/json"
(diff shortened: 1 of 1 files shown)


src/viur/core/render/vi/__init__.py
```diff
@@ -87,7 +87,7 @@ def dumpConfig():
         "modules": res,
         # "configuration": dict(conf.admin.items()), # TODO: this could be the short vision, if we use underscores
         "configuration": {
-            k.replace("_", "."): v for k, v in conf.admin.items()
+            {f"admin.{key}": values for key, values in conf.admin.items()}
```
sveneberth, line 90: That's now a dict nested in a set and `"admin.color_secondary"` instead of `"admin.color.secondary"`


```suggestion
            k.replace("_", "."): v for k, v in conf.admin.items(True)
```

src/viur/core/render/vi/__init__.py
```diff
@@ -152,8 +152,14 @@ def index(*args, **kwargs):
 
 @exposed
 def get_settings():
-    fields = {key: values for key, values in conf.admin.items()}
-    fields["admin.user.google.clientID"] = conf.user.google_client_id
+    """
+    Get public admin-tool specific settings, requires no user to be logged in.
+    This is used by new vi-admin.
+    """
+    fields = {f"admin.{key}": values for key, values in conf.admin.items()}
```
sveneberth, line 159:
```suggestion
    fields = {k.replace("_", "."): v for k, v in conf.admin.items(True)}
```

sveneberth (changes requested):

sveneberth (approved):

PR #984, opened 2023-12-08
feat: Add deprecation handling to `skeleton`

phorward:
Handles `seoKeyBone` usage, which is now `SeoKeyBone`.

### DIFF
--- a/src/viur/core/skeleton.py
+++ b/src/viur/core/skeleton.py
@@ -1565,3 +1565,21 @@ def processVacuumRelationsChunk(
 
 # Forward our references to SkelInstance to the database (needed for queries)
 db.config["SkeletonInstanceRef"] = SkeletonInstance
+
+
+# DEPRECATED ATTRIBUTES HANDLING
+
+__DEPRECATED_NAMES = {
+    # stuff prior viur-core < 3.6
+    "seoKeyBone": ("SeoKeyBone", SeoKeyBone),
+}
+
+def __getattr__(attr: str) -> object:
+    if entry := __DEPRECATED_NAMES.get(attr):
+        func = entry[1]
+        msg = f"{attr} was replaced by {entry[0]}"
+        warnings.warn(msg, DeprecationWarning, stacklevel=2)
+        logging.warning(msg, stacklevel=2)
+        return func
+
+    return super(__import__(__name__).__class__).__getattr__(attr)


sveneberth (changes requested):
Why is the `SeoKeyBone` definition not in `viur.core.bones.seo_key`? Although it is special, the `KeyBone` and every other bone is there too. 

Furthermore, I don't think anyone has used these bones, except within the normal skeleton definition, and we don't need a deprecation here at all.

phorward:
> Why is the `SeoKeyBone` definition not in `viur.core.bones.seo_key`? Although it is special, the `KeyBone` and every other bone is there too.

Because it was initially created there, and I just wanted to have a capitalized class name.

> Furthermore, I don't think anyone has used these bones, except within the normal skeleton definition, and we don't need a deprecation here at all.

It is used by a project where I got this error, therefore I created the pull request.

sveneberth:
> Because it was initially created there, and I just wanted to have a capitalized class name.
> 
> >

The second sentence was actually rather a joke. I think I missed the emoji behind it. All good :)

sveneberth:
> Because it was initially created there, and I just wanted to have a capitalized class name.
> 
> >

But wouldn't now be a good time to change that? I mean, if we have deprecation handling anyway, it shouldn't cause any problems.

phorward:
> But wouldn't now be a good time to change that? I mean, if we have deprecation handling anyway, it shouldn't cause any problems.

I don't want to touch this SEO-fu for now. IMHO, there is huger refactoring required. The reason why there is a need for a SEO-Bone is enough why this is wrong.

Please let's just merge this piece of shit and fix it later.

src/viur/core/skeleton.py
```diff
@@ -1565,3 +1565,22 @@ def processVacuumRelationsChunk(
...
+
+def __getattr__(attr: str) -> object:
+    if entry := __DEPRECATED_NAMES.get(attr):
+        func = entry[1]
+        msg = f"{attr} was replaced by {entry[0]}"
+        warnings.warn(msg, DeprecationWarning, stacklevel=2)
+        logging.warning(msg, stacklevel=2)
+        return func
+
+    return super(__import__(__name__).__class__).__getattr__(attr)
```
sveneberth, line 1586:
```suggestion
    return super(__import__(__name__).__class__).__getattribute__(attr)
```

Same as in https://github.com/viur-framework/viur-core/pull/1065

sveneberth (changes requested):

PR #987, opened 2023-12-12
feat: Provide `UserPassword.on_login()` hook

phorward:
Implements a hook that is called whenever the password authentication was successful. It allows to perform further steps in custom UserPassword authentications.

The login process should be interruptible, so that custom modules can provide additional steps before  `continueAuthenticationFlow` is being called.

In my custom implementation, `on_login` is overwritten as follows:
```py
    def on_login(self, user_entry):
        if user_entry["password_enforce_change"]:
            logging.info("Require password change")
            return self.new_password()

        return super().on_login(self, user_entry)
```

Without this hook, I need to copy the entire `login`-function and inject my custom code before the `continueAuthenticationFlow` call. Please respect this implementation, it works fine.

### DIFF
--- a/src/viur/core/modules/user.py
+++ b/src/viur/core/modules/user.py
@@ -341,6 +341,13 @@ class UserPassword(UserAuthentication):
             skel["password"] = password  # will be hashed on serialize
             skel.toDB(update_relations=False)
 
+        return self.on_login(user_entry)
+
+    def on_login(self, user_entry: db.Entity):
+        """
+        Hook that is called whenever the password authentication was successful.
+        It allows to perform further steps in custom UserPassword authentications.
+        """
         return self._user_module.continueAuthenticationFlow(self, user_entry.key)
 
     @exposed


src/viur/core/modules/user.py
```diff
@@ -341,6 +341,13 @@ def login(self, *, name: str | None = None, password: str | None = None, **kwarg
...
             skel.toDB(update_relations=False)
 
+        return self.on_login(user_entry)
+
+    def on_login(self, user_entry: db.Entity):
+        """
+        Hook that is called whenever the password authentication was successful.
+        It allows to perform further steps in custom UserPassword authentications.
+        """
         return self._user_module.continueAuthenticationFlow(self, user_entry.key)
```
ArneGudermann, line 351:
```suggestion
        self.on_login(user_entry)
        return self._user_module.continueAuthenticationFlow(self, user_entry.key)

    def on_login(self, user_entry: db.Entity):
        """
        Hook that is called whenever the password authentication was successful.
        It allows to perform further steps in custom UserPassword authentications.
        
        :param user_enty: The Datastore User Entity
        """
       pass
```
We can solve it like https://github.com/viur-framework/viur-core/blob/08d6044bae4daabca4ec45cb2c6dae64aac26c8f/src/viur/core/prototypes/list.py#L476

phorward: No! The process should be interruptible, so that custom modules can provide additional steps before  `continueAuthenticationFlow` is being called.

In my custom implementation, `on_login` is overwritten as follows:
```py
    def on_login(self, user_entry):
        if user_entry["password_enforce_change"]:
            logging.info("Require password change")
            return self.new_password()

        return super().on_login(self, user_entry)
```

Without this hook, I need to copy the entire `login`-function and inject my custom code before the `continueAuthenticationFlow` call. Please respect this implementation, it works fine.

ArneGudermann: Ok that is a point.

ArneGudermann (approved):

src/viur/core/modules/user.py
```diff
@@ -341,6 +341,13 @@ def login(self, *, name: str | None = None, password: str | None = None, **kwarg
             skel["password"] = password  # will be hashed on serialize
             skel.toDB(update_relations=False)
 
+        return self.on_login(user_entry)
+
+    def on_login(self, user_entry: db.Entity):
```
sveneberth, line 346: Can we move this in the abstract `UserAuthentication`? I think this can be useful for all methods.

Besides, can we not use action skels for this? You said yourself:

> I don't see any more hooks for actions in the VIUR system

_in https://github.com/viur-framework/viur-core/issues/972#issuecomment-1845301188_

sveneberth (changes requested):

src/viur/core/modules/user.py
phorward: > Can we move this in the abstract `UserAuthentication`? I think this can be useful for all methods.

You're right, I thought about this as well.

> Besides, can we not use action skels for this? You said yourself:
> 
> > I don't see any more hooks for actions in the VIUR system
> 
> _in [#972 (comment)](https://github.com/viur-framework/viur-core/pull/972#issuecomment-1845301188)_

Well... let me clarify:

1. No, I don't want more explicit hooks for special cases, like a hook called `on_item_added_and_edited_and_deleted`
2. Yes, I want more hooks, but these should be universal and dynamical, like my proposed `on()`-hook which is called with event parameter on any action. You've stated that you want code completion for any VIUR modules, so we first have a to work out a solution which is both dynamic and fits all needs.
3. This is not an event hook, but I need a way to influence the behavior of the `UserPassword` primary authentication for custom behaviors - in this case an enforced password change before the login completes. Maybe a hook like this isn't the best way, but it works. I will update this PR soon and add the hook-function into the prototype.

Don't be angry if this comes across as a bit rough, I just want to have a general solution and not something that only works for one special case and not for everything else.

phorward: > > Can we move this in the abstract `UserAuthentication`? I think this can be useful for all methods.
> 
> You're right, I thought about this as well.
> 

@sveneberth can we please merge it here and rewrite this PR for develop? Because in develop, the `UserPrimaryAuthentication` is also made available, which makes things easier. This is a feature needed by a current project, as it copies code from viur-core and extends the feature afterwards, as there is no hooking possible.

sveneberth: Oh, we're here on the road to the main. Okay.

sveneberth (approved):

PR #988, opened 2023-12-12
feat: Allow `None` as defaultValue in BooleanBone

ArneGudermann:
This PR allows BooleanBones to have `None` as `defaultValue`. The Problem is when you have a Question where the user should click Yes or No (`Ture` or `False`) and  the defaultValue is  `False` we can not say if the User don't choose an Option or click No.


sveneberth:
We discussed this at extensive length years ago. But I think we were in agreement that a `BooleanBone` should really only have two states. But I no longer know the exact arguments.
I wouldn't like to change the behavior so easily. Please let's discuss this again with everyone in the next ViUR round.

But I can't quite figure out your use case either. The user selects either True or False okay, but then you'll have an HTML switch that only has two states and is activated or deactivated depending on the default value. If you need more states, you just have to use a `SelectBone` or see if you can solve the problem by overwriting `getEmptyValue()`.

phneutral:
I just stumbled upon the `indeterminate` value that Shoelace provides for its [Checkbox](https://serene-allen-537100.netlify.app/components/checkbox/) and it reminded me of this discussion. 

In CSS you can set a property to `[unset](https://developer.mozilla.org/en-US/docs/Web/CSS/unset)`.

Imho an undetermined initial behaviour could be very valuable (no pun intended) for any bone.

phorward (dismissed):

sveneberth (changes requested):
### Bug 1

This bone:
```py
test_boolean_single_lang1 = BooleanBone(
    descr="test_boolean_single_lang1",
    languages=["de", "en", "fr"],
    defaultValue=True,
)
```

causes this error:
```py
[2024-03-08 17:36:24,205] viur/core/request.py:331 [ERROR] argument of type 'bool' is not iterable
  File "/project/deploy/viur/core/request.py", line 305, in _process
    self._route(path)
  File "/project/deploy/viur/core/request.py", line 561, in _route
    res = caller(*self.args, **kwargs)
  File "/project/deploy/viur/core/module.py", line 299, in __call__
    return self._func(self._instance, *args, **kwargs)
  File "/project/deploy/viur/core/prototypes/list.py", line 272, in add
    return self.render.add(skel)
  File "/project/deploy/viur/core/render/json/default.py", line 219, in add
    return self.renderEntry(skel, action, params)
  File "/project/deploy/viur/core/render/json/default.py", line 162, in renderEntry
    vals = self.renderSkelValues(skel)
  File "/project/deploy/viur/core/render/json/default.py", line 137, in renderSkelValues
    res[key] = self.renderBoneValue(bone, skel, key)
  File "/project/deploy/viur/core/render/json/default.py", line 113, in renderBoneValue
    if boneVal and language in boneVal and boneVal[language]:
TypeError: argument of type 'bool' is not iterable
```


---

### Bug 2

These bones:
```py
test_boolean_single1 = BooleanBone(
    descr="test_boolean_single1",
    defaultValue=True,
)
test_boolean_single2 = BooleanBone(
    descr="test_boolean_single2",
    defaultValue=False,
)
test_boolean_single3 = BooleanBone(
    descr="test_boolean_single3",
    defaultValue=None,
)
# test_boolean_single_lang1 = BooleanBone(
#     descr="test_boolean_single_lang1",
#     languages=["de", "en", "fr"],
#     defaultValue=True,
# )
test_boolean_single_lang2 = BooleanBone(
    descr="test_boolean_single_lang2",
    languages=["de", "en", "fr"],
    defaultValue=False,
)
test_boolean_single_lang3 = BooleanBone(
    descr="test_boolean_single_lang3",
    languages=["de", "en", "fr"],
    defaultValue=None,
)
```

caused this part of value response on `/add`:
```json
"test_boolean_single1": true,
"test_boolean_single2": false,
"test_boolean_single3": null,
"test_boolean_single_lang2": {
  "de": null,
  "en": null,
  "fr": null
},
"test_boolean_single_lang3": {
  "de": null,
  "en": null,
  "fr": null
}
```

As you can see `test_boolean_single_lang2` is `null` instead of `false`.

sveneberth (approved):

PR #990, opened 2023-12-12
feat: Improved and slightly refactored `PasswordBone`

phorward:
- Provides default `descr`
- Cleans up RefSkel and Skeleton in system's User module
- Refactores some parts of PasswordBone for new syntax
- No functional changes

### DIFF
--- a/src/viur/core/modules/user.py
+++ b/src/viur/core/modules/user.py
@@ -80,7 +80,6 @@ class UserSkel(skeleton.Skeleton):
 
     # Properties required by custom auth
     password = PasswordBone(
-        descr="Password",
         required=False,
         readOnly=True,
         visible=False,
@@ -286,11 +285,20 @@ class UserPassword(UserPrimaryAuthentication):
         return "X-VIUR-AUTH-User-Password"
 
     class LoginSkel(skeleton.RelSkel):
-        name = EmailBone(descr="E-Mail", required=True, caseSensitive=False, indexed=True)
-        password = PasswordBone(descr="Password", indexed=True, params={"justinput": True}, required=True)
+        name = EmailBone(
+            descr="Username",
+            required=True,
+            caseSensitive=False,
+        )
+        password = PasswordBone(
+            required=True,
+        )
 
     class LostPasswordStep1Skel(skeleton.RelSkel):
-        name = EmailBone(descr="Username", required=True)
+        name = EmailBone(
+            descr="Username",
+            required=True
+        )
 
     class LostPasswordStep2Skel(skeleton.RelSkel):
         recovery_key = StringBone(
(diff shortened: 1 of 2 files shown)


src/viur/core/modules/user.py
```diff
@@ -286,11 +285,20 @@ def getAuthMethodName(*args, **kwargs):
...
+        )
+        password = PasswordBone(
+            required=True,
+        )
 
     class LostPasswordStep1Skel(skeleton.RelSkel):
-        name = EmailBone(descr="Username", required=True)
+        name = EmailBone(
+            descr="Username",
+            required=True
```
phorward, line 300:
```suggestion
            required=True,
```

src/viur/core/modules/user.py
```diff
@@ -286,11 +285,20 @@ def getAuthMethodName(*args, **kwargs):
         return "X-VIUR-AUTH-User-Password"
 
     class LoginSkel(skeleton.RelSkel):
-        name = EmailBone(descr="E-Mail", required=True, caseSensitive=False, indexed=True)
-        password = PasswordBone(descr="Password", indexed=True, params={"justinput": True}, required=True)
+        name = EmailBone(
+            descr="Username",
```
sveneberth, line 289: That's still an email address. I cannot set this to a custom fantasy name as I would expect from a _username_. I've always found this forced name bone very strange anyway. With ViUR4 we could think about renaming it to _email_. And _name_ can then be computed from _firstname_, _lastname_ and/or _email_.

```suggestion
            descr="E-Mail",
```

src/viur/core/modules/user.py
```diff
@@ -286,11 +285,20 @@ def getAuthMethodName(*args, **kwargs):
...
+            caseSensitive=False,
+        )
+        password = PasswordBone(
+            required=True,
+        )
 
     class LostPasswordStep1Skel(skeleton.RelSkel):
-        name = EmailBone(descr="Username", required=True)
+        name = EmailBone(
+            descr="Username",
```
sveneberth, line 299:
```suggestion
            descr="E-Mail",
```

sveneberth (changes requested):

src/viur/core/modules/user.py
phorward: Ok yes, I think this came in as I copied it from the `LostPasswordStep1Skel` - anyway, I have further plans in future for this field, for customization.

src/viur/core/modules/user.py
```diff
@@ -80,7 +80,6 @@ class UserSkel(skeleton.Skeleton):
 
     # Properties required by custom auth
     password = PasswordBone(
-        descr="Password",
         required=False,
```
sveneberth, line 83: I think we could remove some default arguments. But that's not a big problem either.
```suggestion
```

sveneberth (approved):

sveneberth (approved):

PR #992, opened 2023-12-13
feat: Refactoring and renamed `utils`, adding `utils.string.unescape`

phorward:
- Moves string-related functions into `utils.string`
- Adds `utils.string.unescape` as suggested in #951
- Replaces all usages of moved functions in viur-core
- Improved deprecation handling for both conf and names


src/viur/core/utils/string.py
```diff
@@ -0,0 +1,128 @@
...
+            find = f"&#{find};"
+        else:
+            find = re_match.group(0)
+
+        return __STRING_UNESCAPE_MAPPING.get(find) or re_match.group(0)
+
+    res = re.sub(r"&(\w{2,4}|#0*(\d{2}));", __escape_replace, str(val).strip())
+
+    if max_length:
+        return res[:max_length]
```
ArneGudermann, line 88: I thought that we no longer need `max_length` in the `unescape` function

src/viur/core/utils/string.py
```diff
@@ -0,0 +1,128 @@
...
+        handler = "tree.file.special"
+        utils.string.is_prefix(handler, "tree")  # True
+        utils.string.is_prefix(handler, "tree.node")  # False
+        utils.string.is_prefix(handler, "tree.file")  # True
+        utils.string.is_prefix(handler, "tree.file.special")  # True
+    """
+    return name == prefix or name.startswith(prefix + delimiter)
+
+
+def parse_bool(value: t.Any, truthy_values: t.Iterable[str] = ("true", "yes", "1")) -> bool:
```
ArneGudermann, line 116: Does this function really belong in `string`? You can parse everything to bool.

src/viur/core/utils/__init__.py
```diff
@@ -1,38 +1,23 @@
 import hashlib
 import hmac
 import warnings
-
 import logging
 import secrets
-import string
+import string as py_string
```
ArneGudermann, line 6:
```suggestion
```
We don't need this anymore

src/viur/core/utils/string.py
```diff
@@ -0,0 +1,128 @@
...
+    """
+    Return a string containing random characters of given *length*.
+    It's safe to use this string in URLs or HTML.
+    Because we use the secrets module it could be used for security purposes as well
+
+    :param length: The desired length of the generated string.
+
+    :returns: A string with random characters of the given length.
+    """
+    return "".join(secrets.choice(py_string.ascii_letters + py_string.digits) for _ in range(length))
```
ArneGudermann, line 20:
```suggestion
    return "".join(secrets.choice(string.ascii_letters + string.digits) for _ in range(length))
```

ArneGudermann (changes requested):
Hey @phorward thank you for this PR. I have some suggestions.

src/viur/core/utils/string.py
phorward: Hmmm, well parsing means, to analyze a string and give its content a meaning and structure. Therefore, I think yes. We might also introduce a `utils.parse` module supporting `bool` as a function, as I also would like to introduce a JavaScript-style `parse_float` and `parse_int`. What do you think?

ArneGudermann: Yes `utils.parse` is a nice idea.

src/viur/core/utils/__init__.py
```diff
@@ -1,38 +1,20 @@
...
-import typing
 from base64 import urlsafe_b64encode
 from datetime import datetime, timedelta, timezone
 from typing import Any, Union, Optional
 from urllib.parse import quote
-
 from viur.core import current, db
 from viur.core.config import conf
-
+from . import string
```
sveneberth, line 11: I would prefer to have a
```py
from .string import *
```
with `__all__` configured so I can just import everything from `viur.core.utils`. This would also be no problem if we change the structure again in the future.

sveneberth (changes requested):

src/viur/core/utils/__init__.py
phorward: Hmmm... okay I've now introduced `parse.bool`, which would introduce `bool` as a function - no good idea ;-)

src/viur/core/utils/string.py
phorward: Maybe this approach is too blatant, but check it out!

src/viur/core/utils/__init__.py
sveneberth: Well, okay then…

sveneberth (dismissed):

ArneGudermann (approved):

PR #1001, opened 2023-12-18
feat: Implement `@property` support in `Skeleton`s

sveneberth:
Sometimes you want to perform skeleton related actions like `skel.view_url`. `@property`s  are a nice way for this.

Example:

```py
class ExampleSkel(Skeleton):
    # [...]

    @property
    def test_property(self):
        return f'test_property of {self.__class__.__name__} {self.key} {self["key"]}'
```

### DIFF
--- a/src/viur/core/skeleton.py
+++ b/src/viur/core/skeleton.py
@@ -208,19 +208,54 @@ class SkeletonInstance:
         self.renderAccessedValues[key] = value
         return value
 
-    def __getattr__(self, item):
+    def __getattr__(self, item: str):
+        """
+        Get a special attribute from the SkeletonInstance
+
+        __getattr__ is called when an attribute access fails with an
+        AttributeError. So we know that this is not a real attribute of
+        the SkeletonInstance. But there are still a few special cases in which
+        attributes are loaded from the skeleton class.
+        """
         if item == "boneMap":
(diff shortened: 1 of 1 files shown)


src/viur/core/skeleton.py
```diff
@@ -208,19 +208,54 @@ def __getitem__(self, key):
...
+                    msg, *args = exc.args
+                    msg = f"AttributeError: {msg}"
+                    raise ValueError(msg, *args) from exc
+        # Load the bone instance from the bone map of this SkeletonInstance
         try:
             return self.boneMap[item]
-        except KeyError:
-            raise AttributeError(f"{self.__class__.__name__!r} object has no attribute '{item}'")
+        except KeyError as exc:
+            raise AttributeError(f"{self.__class__.__name__!r} object has no attribute '{item}'") from exc
```
sveneberth, line 258: @ Reviewers:

I changed this too. We get now

> The above exception was the direct cause of the following exception:

which sounds like intended (which it is)

while before with

> During handling of the above exception, another exception occurred:

it sounded like an unexpected error.

---

I didn't want to make an extra PR for it now, because I was working on the method anyway.

ArneGudermann (approved):

phorward (approved):

PR #1006, opened 2023-12-19
fix: handling alpha channel within thumbnail generation if icc profile is present

skoegl:
Now we keep transparency in generated thumbnails if needed


phorward (approved):
Thank you!

Issue #1007, opened 2023-12-21
Assign a `db.Key` as value with `setBoneValue` to a `KeyBone` returns None


sveneberth:
Because `from_legacy_urlsafe` fails (If I saw this correctly ...):
https://github.com/viur-framework/viur-core/blob/1efa29697cdac2dbf82510993ea16adb3cd6a984/src/viur/core/bones/key.py#L50

I tried 
```py
key = db.Key("foo", "bar")
skel.setBoneValue("parententry", key)
```
on a TreeSkel

PR #1008, opened 2023-12-21
fix: Add check for `db.Key` in `KeyBone.singleValueFromClient`

ArneGudermann:
Fix for #1007

### DIFF
--- a/src/viur/core/bones/key.py
+++ b/src/viur/core/bones/key.py
@@ -47,7 +47,10 @@ class KeyBone(BaseBone):
                 return self.getEmptyValue(), [ReadFromClientError(ReadFromClientErrorSeverity.Invalid, e.args[0])]
         else:
             try:
-                key = db.normalizeKey(db.Key.from_legacy_urlsafe(value))
+                if isinstance(value, str):
+                    key = db.normalizeKey(db.Key.from_legacy_urlsafe(value))
+                elif isinstance(value, db.Key):
+                    key = db.normalizeKey(value)
             except:
                 return self.getEmptyValue(), [
                     ReadFromClientError(
--- a/src/viur/core/bones/key.py
+++ b/src/viur/core/bones/key.py
@@ -51,7 +51,8 @@ class KeyBone(BaseBone):
                     key = db.normalizeKey(db.Key.from_legacy_urlsafe(value))
                 elif isinstance(value, db.Key):
                     key = db.normalizeKey(value)
-            except:
+            except Exception as exc:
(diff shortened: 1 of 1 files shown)


phorward (approved):

src/viur/core/bones/key.py
```diff
@@ -47,7 +47,10 @@ def singleValueFromClient(self, value, skel, bone_name, client_data):
                 return self.getEmptyValue(), [ReadFromClientError(ReadFromClientErrorSeverity.Invalid, e.args[0])]
         else:
             try:
-                key = db.normalizeKey(db.Key.from_legacy_urlsafe(value))
+                if isinstance(value, str):
+                    key = db.normalizeKey(db.Key.from_legacy_urlsafe(value))
+                elif isinstance(value, db.Key):
+                    key = db.normalizeKey(value)
             except:
```
sveneberth, line 54:
```suggestion
            except Exception as exc:
                logging.exception(f"Failed to normalize {value}: {exc}")
```

sveneberth (changes requested):

src/viur/core/bones/key.py
phorward: @ArneGudermann can you please add this proposal to your PR in order for merging?

sveneberth (dismissed):

src/viur/core/bones/key.py
```diff
@@ -47,8 +47,12 @@ def singleValueFromClient(self, value, skel, bone_name, client_data):
                 return self.getEmptyValue(), [ReadFromClientError(ReadFromClientErrorSeverity.Invalid, e.args[0])]
         else:
             try:
-                key = db.normalizeKey(db.Key.from_legacy_urlsafe(value))
-            except:
+                if isinstance(value, str):
+                    key = db.normalizeKey(db.Key.from_legacy_urlsafe(value))
+                elif isinstance(value, db.Key):
+                    key = db.normalizeKey(value)
```
phorward, line 53: What happens here when key is either not `isinstance(value, str)` and not `isinstance(value, db.Key)`? In this case, key will left as is, and program flow continues.

I would like suggest:
```suggestion
                if isinstance(value, db.Key):
                    key = db.normalizeKey(value)
                else:
                    key = db.normalizeKey(db.Key.from_legacy_urlsafe(value))
```
to keep closer to the previous behavior.

phorward (changes requested):

phorward (approved):

Issue #1019, opened 2024-01-10
Accept prefix `refKey` value in `RelationalBone`


sveneberth:
Instead of this list
https://github.com/viur-framework/viur-shop/blob/dfccf12bc2fe189c14f9d83f01c3c277577c3c28/src/viur/shop/skeletons/cart.py#L70-L75
it would be more nice to add just `shop_*` as wildcard prefix.

phorward:
fixed by #1022

PR #1022, opened 2024-01-11
feat: Support `*`-wildcard postfix in `refKeys` for `RelationalBones`

ArneGudermann:
proposal for #1019.


src/viur/core/skeleton.py
```diff
@@ -1338,7 +1338,14 @@ def fromSkel(cls, kindName: str, *args: List[str]) -> Type[RefSkel]:
...
         fromSkel = skeletonByKind(kindName)
-        newClass.__boneMap__ = {k: v for k, v in fromSkel.__boneMap__.items() if k in args}
+        bone_map = {}
+        for arg in args:
+            if "*" in arg:  # we have a wild card
+                bone_map |= {k: v for k, v in fromSkel.__boneMap__.items() if k.startswith(arg.replace("*", ""))}
+            else:
+                if bone := fromSkel.__boneMap__.get(arg):
+                    bone_map[arg] = bone
+        newClass.__boneMap__ = bone_map
```
sveneberth, line 1348: That looks complicated. Couldn't we simplify this with [`utils.string.is_prefix()`](https://github.com/viur-framework/viur-core/blob/c34d45fdf0eeeace4d4317ab89cd251ddf6191cf/src/viur/core/utils/string.py#L88): 

```suggestion
        for arg in args:
            bone_map |= {k: v for k, v in fromSkel.__boneMap__.items()
                         if utils.string.is_prefix(k, arg, "*"))}
        newClass.__boneMap__ = bone_map
```

sveneberth (changes requested):

src/viur/core/skeleton.py
ArneGudermann: But now we iter over the bonemap `len(args) `time ? That is not so good i think.

sveneberth: > But now we iter over the bonemap `len(args) `time ? That is not so good i think.

well, okay.


But I think index access would be faster than an _in_ test. And slicing the prefix once would be better than _replace_ on every _if_ check. But I didn't bench marked this ...


```suggestion
        for arg in args:
            if arg[-1] == "*":  # we have a wild card prefix
                prefix = arg[:-1]
                bone_map |= {k: v for k, v in fromSkel.__boneMap__.items() if k.startswith(prefix)}
            else:
                if bone := fromSkel.__boneMap__.get(arg):
                    bone_map[arg] = bone
        newClass.__boneMap__ = bone_map
```

sveneberth (changes requested):
I think there are more changes necessary. We valide the `refKeys` also before filtering/ordering:

https://github.com/search?q=repo%3Aviur-framework%2Fviur-core+self.refKeys++path%3A**%2Fsrc%2Fviur%2Fcore%2Fbones%2Frelational.py&type=code

We should resolve the wildcard in the bone in `getSystemInitialized()`. Maybe this could already be enough:
```py
self.refKeys = list(self._refSkelCache__boneMap__.keys())
```

src/viur/core/skeleton.py
```diff
@@ -1363,8 +1363,15 @@ def fromSkel(cls, kindName: str, *args: list[str]) -> t.Type[RefSkel]:
         """
         newClass = type("RefSkelFor" + kindName, (RefSkel,), {})
         fromSkel = skeletonByKind(kindName)
-        newClass.__boneMap__ = {k: v for k, v in fromSkel.__boneMap__.items() if k in args}
-        newClass.kindName = kindName
+        bone_map = {}
```
sveneberth, line 1366:
```suggestion        
        newClass.kindName = kindName
        bone_map = {}
```

sveneberth (changes requested):

sveneberth (dismissed):

phorward:
@ArneGudermann what do you think, should we use fnmatch here as well? This would also allow to match `*_bones` next to `a*bones` ... I think this would be very nice!

ArneGudermann:
@phorward Now better ?

src/viur/core/skeleton.py
```diff
@@ -1363,8 +1364,15 @@ def fromSkel(cls, kindName: str, *args: list[str]) -> t.Type[RefSkel]:
...
         fromSkel = skeletonByKind(kindName)
-        newClass.__boneMap__ = {k: v for k, v in fromSkel.__boneMap__.items() if k in args}
         newClass.kindName = kindName
+        bone_map = {}
+        for arg in args:
+            if arg[-1] == "*":  # we have a wild card prefix
+                bone_map |= {k: v for k, v in fromSkel.__boneMap__.items() if fnmatch.fnmatch(k, arg)}
+            else:
+                if bone := fromSkel.__boneMap__.get(arg):
+                    bone_map[arg] = bone
```
phorward, line 1374: > @phorward Now better ?

Quite, but it's still inconsistent, and only accepts keys ending with "*". 
```suggestion
        for arg in args:
            bone_map |= {k: fromSkel.__boneMap__[k] for k in fnmatch.filter(fromSkel.__boneMap__.keys(), arg)}
```
What do you think about this?

phorward (changes requested):

phorward (approved):
@sveneberth are you still fine with this?

sveneberth (approved):

PR #1029, opened 2024-01-17
feat: Support compute for multilang and/or multiple bones

sveneberth:


src/viur/core/bones/base.py
```diff
@@ -1212,8 +1212,21 @@ def _compute(self, skel: 'viur.core.skeleton.SkeletonInstance', bone_name: str):
 
         ret = self.compute.fn(**compute_fn_args)
 
+        def unserialize_raw_value(value: list[dict] | dict | None):
+            if self.multiple:
+                return [self.singleValueUnserialize(value)
+                        for value in ret]
+            else:
+                return self.singleValueUnserialize(ret)
```
phorward, line 1220:
```suggestion
            return self.singleValueUnserialize(ret)
```

src/viur/core/bones/base.py
```diff
@@ -1212,8 +1212,21 @@ def _compute(self, skel: 'viur.core.skeleton.SkeletonInstance', bone_name: str):
...
+
         if self.compute.raw:
-            return self.singleValueUnserialize(ret)
+            if self.languages:
+                return {
+                    lang: unserialize_raw_value(ret.get(lang, [] if self.multiple else None))
+                    for lang in self.languages
+                }
+            else:
+                return unserialize_raw_value(ret)
```
phorward, line 1229:
```suggestion
            return unserialize_raw_value(ret)
```

phorward (changes requested):

src/viur/core/bones/base.py
sveneberth: To quote the second zen of python:
> Explicit is better than implicit.

But if the maintainer wants it that way, that's fine for me. :smiley:

phorward: Keep it short and simple. <3

phorward (approved):

PR #1030, opened 2024-01-17
feat: Add and improve check on root node in `Tree` prototype

sveneberth:
`rootNode` is neither snake_case or a boolen (`=1`).

In `viur-shop` I use the bone `is_root_node`: https://github.com/viur-framework/viur-shop/blob/eacfb4a9e5288d3f3b5bf4c0b6445657cb48ff3b/src/viur/shop/skeletons/cart.py#L79
* It's snake_case
* a BooleanBone
* and `is_` is describing a state. You could also think that the bone rootNode points to a rootnode as relation.

### DIFF
--- a/src/viur/core/prototypes/tree.py
+++ b/src/viur/core/prototypes/tree.py
@@ -546,7 +546,8 @@ class Tree(SkelModule):
 
         # Test if we try to move a rootNode
         tmp = skel.dbEntity
-        if tmp.get("rootNode") == 1 or tmp.get("is_root_node"):
+        # TODO: Remove "rootNode"-fallback with VIUR4
+        if skel.dbEntity.get("is_root_node") or skel.dbEntity.get("rootNode"):
             raise errors.NotAcceptable("Can't move a rootNode to somewhere else")
 
         currentParentRepo = skel["parentrepo"]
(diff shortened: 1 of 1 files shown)


src/viur/core/prototypes/tree.py
```diff
@@ -546,7 +546,7 @@ def move(self, skelType: SkelType, key: db.Key | int | str, parentNode: str, *ar
 
         # Test if we try to move a rootNode
         tmp = skel.dbEntity
```
phorward, line 548:
```suggestion
```

src/viur/core/prototypes/tree.py
```diff
@@ -546,7 +546,7 @@ def move(self, skelType: SkelType, key: db.Key | int | str, parentNode: str, *ar
...
-        if "rootNode" in tmp and tmp["rootNode"] == 1:
+        if tmp.get("rootNode") == 1 or tmp.get("is_root_node"):
```
phorward, line 549:
```suggestion
        # TODO: Remove "rootNode"-fallback with VIUR4
        if skel.dbEntity.get("is_root_node") or skel.dbEntity.get("rootNode"):
```

phorward (changes requested):
Nice! But can you please refactor out this stupid `tmp`?

phorward (approved):

PR #1032, opened 2024-01-17
feat: Add `scriptor` access flag

phorward:
- Adds access flag `scriptor`
- Documentation for the default access flags and the default roles
- User default access flag with `scriptor` for users having the `admin`-role


sveneberth (commented):
Well, the new roles are okay for me. But where do we use them? You didn't change a `can*` method of a module/prototype :thinking:

akelch (approved):

ArneGudermann (approved):

phorward:
> Well, the new roles are okay for me. But where do we use them? You didn't change a `can*` method of a module/prototype 🤔

You're right. My initial intention behind this was to display the "scriptor"-button in the Admin. I will have further consideration about this and let you know.

Grashalmbeisser:
There needs to be an distinction between a Flag and an Access right. Since "Scriptor" is not an Acces right but a usage privillege we should differentiate that as a flag not as a right.

phorward:
Upated this PR in convenience with @akelch. Clarified naming convention, so an access flag is something like "root", "admin" or now "scriptor", and an access-right is the stuff with the "add", "edit", "delete", "view" or any other action-related.

ArneGudermann (approved):

sveneberth (approved):

PR #1037, opened 2024-01-22
fix: Remove default `P.html = True` from prototypes

phorward:
Ever wondered why every module you want to access by URL without specifying a renderer throws out an error message like this?

image

Well, this pull requests removes the inconsequent `Prototype.html = True` from the `List` and `Singleton` prototype. The `Tree` prototype was always not by default visible to the html-render.

This is how the same call is rendered now:
image

Normally, all modules being visible to the standard html render have a `Module.html = True` on the bottom. Therefore, this little change wouldn't break too much. I'm also of the opinion, that it is a security issue.

### DIFF
--- a/src/viur/core/prototypes/list.py
+++ b/src/viur/core/prototypes/list.py
@@ -571,5 +571,4 @@ class List(SkelModule):
 
 
 List.admin = True
-List.html = True
 List.vi = True
--- a/src/viur/core/prototypes/singleton.py
+++ b/src/viur/core/prototypes/singleton.py
@@ -293,5 +293,4 @@ class Singleton(SkelModule):
 
 
 Singleton.admin = True
-Singleton.html = True
 Singleton.vi = True


sveneberth:
> Ever wondered why every module you want to access by URL without specifying a renderer throws out an error message like this?

honestly, not really




> The `Tree` prototype was always not by default visible to the html-render.

I think this is due to historical reasons. the prototype is/was mostly only used for the menu and file module, so it wasn't necessary.





> Normally, all modules being visible to the standard html render have a `Module.html = True` on the bottom

No, not really.
Like I don't write `required=False` on a bone every time, I rely on the inherited default value here. So IMO this would also be a breaking change.

---

As long as the HTML renderer is set as the default renderer, I don't really see the benefit of this PR and the security issue

phorward:
> > Normally, all modules being visible to the standard html render have a `Module.html = True` on the bottom
> 
> No, not really. Like I don't write `required=False` on a bone every time, I rely on the inherited default value here. So IMO this would also be a breaking change.
> 
> As long as the HTML renderer is set as the default renderer, I don't really see the benefit of this PR and the security issue

For me, this is a security problem. By default, all modules shall not be accessible by any render, except the admin. In case the viur-base was used, there is a default `view.html` template, so the above call would render data which was not explicitly allowed to be rendered.

This is not a breaking change, it is a bug- and security-fix. Maybe, some projects which don't follow standards need to be fixed.

phorward:
In the viur-meeting from Feb 2, 2024, we decided to apply this pull request.

Additionally, any `P.admin = True`-assignments should be removed, too.
For latter one, I would like to discuss why the `vi`-render is not simply renamed into `admin`, and aliased by `vi` to stay backward-compatible.

theVAX (approved):
looks fine to me 👍

PR #995, opened 2023-12-13
feat: Collect modules recursivly in `vi/config`

sveneberth:
The viur-shop has it's modules as submodules under a main `shop`. As you can see [here](https://github.com/viur-framework/viur-shop/blob/191b6a56e764a709170bcd1e3f1e27f1e3fa4c1b/src/viur/shop/shop.py#L45-L53).

So for example we have this path `/shop/address/list`.
It's perfectly routeable, but the vi doesn't know it because submodules are not collected. This PR adds a recursive walk through all the modules and collects them in `/vi/config`.

PR #1041, opened 2024-01-25
feat: Collect modules recursivly in `read_all_modules`

ArneGudermann:
This PR adds a recursive walk through all the modules and collects them in `read_all_modules`.

for further information take a look at #995

### DIFF
--- a/src/viur/core/modules/moduleconf.py
+++ b/src/viur/core/modules/moduleconf.py
@@ -68,16 +68,36 @@ class ModuleConf(List):
 @StartupTask
 def read_all_modules():
     db_module_names = [m["name"] for m in db.Query("viur-module-conf").run(999)]
-    module_names = dir(conf.main_app.vi)
+    visited_modules = set()
 
-    for module_name in module_names:
-        module = getattr(conf.main_app.vi, module_name)
-        if isinstance(module, Module):
+    def collect_modules(parent, depth: int = 0, prefix: str = "") -> None:
+        """Recursively collects all routable modules for the vi renderer"""
+        if depth > 10:
+            logging.warning(f"Reached maximum recursion limit of {depth} at {parent=}")
+            return
+
+        for module_name in dir(parent):
+            module = getattr(parent, module_name, None)
+            if not isinstance(module, Module):
+                continue
+            if module in visited_modules:
+                # Some modules reference other modules as parents, this will
+                # lead to infinite recursion. We can avoid reaching the
+                # maximum recursion limit by remembering already seen modules.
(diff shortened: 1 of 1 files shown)


src/viur/core/modules/moduleconf.py
```diff
@@ -68,16 +68,36 @@ def get_by_module_name(cls, module_name: str) -> None | SkeletonInstance:
...
+                # maximum recursion limit by remembering already seen modules.
+                if conf.debug.trace:
+                    logging.debug(f"Already visited and added {module=}")
+                continue
+            visited_modules.add(module)
             if module_name not in db_module_names:
                 skel = conf.main_app._moduleconf.addSkel()
-                skel["key"] = db.Key("viur-module-conf", module_name)
-                skel["name"] = module_name
+                skel["key"] = db.Key("viur-module-conf", f"{prefix}{module_name}")
```
phorward, line 93: Make this less static!
```suggestion
                skel["key"] = db.Key(skel.kindName, f"{prefix}{module_name}")
```

src/viur/core/modules/moduleconf.py
```diff
@@ -68,16 +68,36 @@ def get_by_module_name(cls, module_name: str) -> None | SkeletonInstance:
 @StartupTask
 def read_all_modules():
     db_module_names = [m["name"] for m in db.Query("viur-module-conf").run(999)]
```
phorward, line 70: Please make the `"viur-module-conf"` more dynamic. Either by using a skeleton.kindName (which is not the case here - maybe use `skel_cls = self._resolveSkelCls()` and use `skel_cls.kindName`, or using a constant defined on top.)

phorward (changes requested):
Hello @ArneGudermann, can you please make the kind name more dynamic, so there is not the same string coming up multiple times? I'd given some advises.

src/viur/core/modules/moduleconf.py
ArneGudermann: But can the `kindName` change ? Because we set the module here hardcoded.
https://github.com/viur-framework/viur-core/blob/486d500c0e226feac67c28bdd3d2a6d0f27c4fb9/src/viur/core/__init__.py#L123
https://github.com/viur-framework/viur-core/blob/486d500c0e226feac67c28bdd3d2a6d0f27c4fb9/src/viur/core/__init__.py#L127)

sveneberth (changes requested):

sveneberth (approved):

phorward (approved):

PR #1046, opened 2024-01-26
refactor: Modularization of `File`

phorward:
This partly refactors the `File`-module to the following:

- Moved and renamed `utils`-functions into `modules.file` as they belong there
- Removed unused variables and unused functions
- Fixed calls to prior `utils`-functions (e.g. from the renders)
- Fixed `expired`-flag in srcSetFor() as this was inconsistent to `downloadUrlFor`


src/viur/core/render/html/env/viur.py
```diff
@@ -695,23 +698,42 @@ def downloadUrlFor(render: Render,
...
         expires = timedelta(minutes=expires)
+
     if not isinstance(fileObj, (SkeletonInstance, dict)) or "dlkey" not in fileObj or "name" not in fileObj:
-        return None
+        logging.error("Invalid fileObj supplied")
+        return ""
+
     if derived and ("derived" not in fileObj or not isinstance(fileObj["derived"], dict)):
-        return None
+        return ""
```
ArneGudermann, line 710:
```suggestion
        logging.error("No derivation for this fileObj")
        return ""
```

ArneGudermann (changes requested):
Thank you @phorward  for the PR. I have one more question. Shouldn't we include the functions that are now out of the utils in the [__UTILS_NAME_REPLACEMENT](https://github.com/viur-framework/viur-core/blob/96d91c2f5d135c9eeeec68b3b5e20bf4cdefe9f6/src/viur/core/utils/__init__.py#L246) ?

phorward:
> Thank you @phorward for the PR. I have one more question. Shouldn't we include the functions that are now out of the utils in the [__UTILS_NAME_REPLACEMENT](https://github.com/viur-framework/viur-core/blob/96d91c2f5d135c9eeeec68b3b5e20bf4cdefe9f6/src/viur/core/utils/__init__.py#L246) ?

No, these functions where only used in viur-core itself. Therefore, I don't think it's necessary.

sveneberth (approved):

ArneGudermann (approved):

PR #1050, opened 2024-01-31
refactor: `File.getUploadURL()`

### LINKED ISSUE #1017: Handling of filenames with umlauts and other special characters
This problem currently exists in ViUR.
A file named `testupload- ö ä -(123).txt` should be uploaded.

1. The file `testupload- ö ä -(123).txt` is selected or dropped. 
image
2. [`file/getUploadURL`](https://github.com/viur-framework/viur-core/blob/main/src/viur/core/modules/file.py#L529) is being called with this filename, and generates an upload URL to Google Cloud Storage with the cloud bucket and file path `project-id.appspot.com/06eJqwTaFo8JO/source/testupload-%20%C3%B6%20%C3%A4%20-%28123%29.txt`.
image
   This first sanitizes the filename to the above filename form ([source](https://github.com/viur-framework/viur-core/blob/main/src/viur/core/modules/file.py#L504)).
   ```python
        fileName = sanitizeFileName(fileName)

        targetKey = utils.generateRandomString()
        blob = bucket.blob("%s/source/%s" % (targetKey, fileName))
        uploadUrl = blob.create_resumable_upload_session(content_type=mimeType, size=size, timeout=60)
   ```
   Then, it creates a new file leaf skel with the filename `pending` in ([source](https://github.com/viur-framework/viur-core/blob/main/src/viur/core/modules/file.py#L512)):
   ```python
        fileSkel = self.addSkel("leaf")
        fileSkel["name"] = "pending"
        fileSkel["size"] = 0
        fileSkel["mimetype"] = "application/octetstream"
        fileSkel["dlkey"] = targetKey
        fileSkel["parentdir"] = None
        fileSkel["pendingparententry"] = db.keyHelper(node, self.addSkel("node").kindName) if node else None
        fileSkel["pending"] = True
        fileSkel["weak"] = True
        fileSkel["width"] = 0
        fileSkel["height"] = 0
        fileSkel.toDB()
   ```
3. The file is being uploaded to Google Cloud Storage using the provided upload URL.
4. Afterwards, [`file/add`](https://github.com/viur-framework/viur-core/blob/main/src/viur/core/modules/file.py#L677) is being called with the key of the file leaf skel created by `file/getUploadURL`. This now reads the Google Cloud storage object blob and extracts the filename again from it ([source](https://github.com/viur-framework/viur-core/blob/main/src/viur/core/modules/file.py#L710)):
   ```python
            skel["name"] = utils.escapeString(blob.name.replace("%s/source/" % skel["dlkey"], ""))
   ```
   Resulting file entity in Google Cloud Datastore:
   image
   Wow! Such a nicely resolved piece of code.
5. The file is uploaded. The filename is displayed "broken" everywhere it is being used. A simple `string unescaping` is not enough, as the filename is URL-encoded. Why? For security reasons? 
image

In the end, this entire stuff is inconsistent and breaks the filename.

1. In step 2: Why not save the pending file leaf skeleton with the unquoted filename, like `testupload- ö ä -(123).txt (pending)`
2. In step 4: Why is `skel["name"]` being created from `blob.name`? Why is `06eJqwTaFo8JO/source/testupload-%20%C3%B6%20%C3%A4%20-%28123%29.txt` turned into `testupload-%20%C3%B6%20%C3%A4%20-%28123%29.txt` and then the URL-escaped string is `utils.escapeString`ed again. Is this extra-secure? Why not just use the original filename provided before without the `(pending)`-postfix, and this, as normal for `StringBone` content, is required to be `utils.escapeString`ed.
3. General question: Why is ViUR required to save escaped data for any string content? Why is such data not escaped when rendered, so the HTML-renderer is the place where data is going to be escaped, so no code leaking or injection will be possible. In JSON or any other format, the escaping is irellevant and impractical.

phorward:
This pull request is focuesed to (partly) fix #1017.

### DIFF
--- a/src/viur/core/modules/file.py
+++ b/src/viur/core/modules/file.py
@@ -408,6 +408,7 @@ class FileLeafSkel(TreeSkel):
         descr="Download-Key",
         readOnly=True
     )
+
     name = StringBone(
         descr="Filename",
         caseSensitive=False,
@@ -499,6 +500,8 @@ class FileNodeSkel(TreeSkel):
 
 
 class File(Tree):
+    PENDING_POSTFIX = " (pending)"
+
     leafSkelCls = FileLeafSkel
     nodeSkelCls = FileNodeSkel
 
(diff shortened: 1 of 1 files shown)


sveneberth:
viur meeting result:
* limit charset like windows filesystem https://stackoverflow.com/a/31976060/3749896
* refususe invalid filenames

phorward:
> https://stackoverflow.com/a/31976060/3749896

I've created a regex for this, which should work: https://regex101.com/r/faUewi/1

src/viur/core/modules/file.py
```diff
@@ -766,104 +799,121 @@ def download(self, blobKey: str, fileName: str = "", download: str = "", sig: st
...
+                response = current.request.get().response
+                response.headers["Content-Type"] = blob.content_type
+                response.headers["Cache-Control"] = "public, max-age=604800"  # 7 Days
+                if content_disposition:
+                    response.headers["Content-Disposition"] = content_disposition
+                return blob.download_as_bytes()
+
+        # Default fallback - create a signed URL and redirect
+        authRequest = google.auth.transport.requests.Request()
+        expiresAt = datetime.datetime.now() + datetime.timedelta(seconds=60)
```
ArneGudermann, line 856:
```suggestion
    if vaildUntil == "0":
        expiresAt =  datetime.timedelta(days=7)
    else:
        try:
            expiresAt = datetime.datetime.strptime(vaildUntil, "%Y%m%d%H%M")
            if expiresAt-datetime.datetime.now() > datetime.timedelta(days=7):
                expiresAt = datetime.timedelta(days=7)
        except:
            expiresAt = datetime.timedelta(seconds=60)
```

src/viur/core/modules/file.py
```diff
@@ -766,104 +799,121 @@ def download(self, blobKey: str, fileName: str = "", download: str = "", sig: st
...
+
+        content_disposition = "; ".join(
+            item for item in (
+                "attachment" if download else None,
+                f"filename={filename}" if filename else None,
+            ) if item
+        )
+
+        if isinstance(_CREDENTIALS, ServiceAccountCredentials):
             expiresAt = datetime.datetime.now() + datetime.timedelta(seconds=60)
```
ArneGudermann, line 834:
```suggestion
            if vaildUntil == "0":
                expiresAt =  datetime.timedelta(days=7)
            else:
                try:
                    expiresAt = datetime.datetime.strptime(vaildUntil, "%Y%m%d%H%M")
                    if expiresAt-datetime.datetime.now() > datetime.timedelta(days=7):
                        expiresAt = datetime.timedelta(days=7)
                except:
                    expiresAt = datetime.timedelta(seconds=60)
```

ArneGudermann (changes requested):

src/viur/core/modules/file.py
phorward: Already talked to @ArneGudermann, but here for documentation: We'll gonna make this proposal in a separate PR apart of #1050 when #1050 is merged.

ArneGudermann (dismissed):

src/viur/core/modules/file.py
```diff
@@ -766,104 +799,121 @@ def download(self, blobKey: str, fileName: str = "", download: str = "", sig: st
...
             if len(blobs) != 1:
                 logging.error("Invalid number of blobs in folder")
                 logging.error(targetKey)
                 raise errors.PreconditionFailed()
+
             blob = blobs[0]
             skel["mimetype"] = utils.string.escape(blob.content_type)
-            if any([x in blob.name for x in "$<>'\""]):  # Prevent these Characters from being used in a fileName
+
+            if any([ch in blob.name for ch in "$<>'\""]):  # Prevent these Characters from being used in a fileName
```
sveneberth, line 907: Shouldn't this be the new method?

```suggestion
            if not self.is_valid_filename(blob.name):
```

sveneberth (changes requested):

src/viur/core/modules/file.py
phorward: I wonder why this condition is here, as the filename of the blob checked when the upload URL is being provided. I think we can remove this entire thing? There's also no helpful comment about *why* this is being done here.

sveneberth: with `getUploadURL` you only request a url for upload and say that the file has certain metadata. Which data you finally send to this upload URL is then independent. Therefore, everything (size, filename, mimetype) must be checked again at this point.

sveneberth: But the check cloud also be
```py
if blob.name != skel["name"].removesuffix(PENDING_POSTFIX):
    raise ValueError("this is not the file you asked for in getUploadUrl")

phorward: Hmm okay... then let's do it!

sveneberth (approved):

PR #1052, opened 2024-01-31
fix: `RelationalBone` locking bug

phorward:
This fixes the bug that comes up when there are two RelationalBones with `RelationalConsistency.PreventDeletion` being set to the same target kind. It raises an error 500 on a customer project, as the assertion and helpful "PROGRAMMING ERROR DETECTED" message are raised, which causes that this bug arose.

### DIFF
--- a/src/viur/core/bones/relational.py
+++ b/src/viur/core/bones/relational.py
@@ -506,7 +506,7 @@ class RelationalBone(BaseBone):
         for newLock in newRelationalLocks - oldRelationalLocks:
             # Lock new Entry
             if referencedObj := db.Get(newLock):
-                referencedOb.setdefault("viur_incomming_relational_locks", [])
+                referencedObj.setdefault("viur_incomming_relational_locks", [])
 
                 if skel["key"] not in referencedObj["viur_incomming_relational_locks"]:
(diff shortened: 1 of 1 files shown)


src/viur/core/bones/relational.py
```diff
@@ -486,35 +486,40 @@ def serialize(self, skel: 'SkeletonInstance', name: str, parentIndexed: bool) ->
...
-            referencedObj = db.Get(newLock)
-            assert referencedObj, "Programming error detected?"
-            if not referencedObj.get("viur_incomming_relational_locks"):
-                referencedObj["viur_incomming_relational_locks"] = []
-            assert skel["key"] not in referencedObj["viur_incomming_relational_locks"]
-            referencedObj["viur_incomming_relational_locks"].append(skel["key"])
-            db.Put(referencedObj)
+            if referencedObj := db.Get(newLock):
+                if not referencedObj.get("viur_incomming_relational_locks"):
+                    referencedObj["viur_incomming_relational_locks"] = []
```
ArneGudermann, line 510:
```suggestion
                referencedOb.setdefault("viur_incomming_relational_locks", [])
```

ArneGudermann (changes requested):

ArneGudermann (approved):

PR #1053, opened 2024-01-31
feat: Allow `None` in `skel setBoneValue`

### LINKED ISSUE #1047: `skel.setBoneValue` should accept `None` as value
I've situations where my method get values from outside and it should be okay to provide a (real) value or `None` to clear this value. But currently I've to do
```py
discount_key: db.Key | None
if discount_key is None:
    skel["discount"] = None
else:
    skel.setBoneValue("discount", discount_key)
```
I want to write only
```py
discount_key: db.Key | None
skel.setBoneValue("discount", discount_key)
```

ArneGudermann:
Resolves #1047.
@sveneberth  I hope I have understood your issue correctly.  
This PR sets the value of the bones to `None`.

### DIFF
--- a/src/viur/core/skeleton.py
+++ b/src/viur/core/skeleton.py
@@ -391,30 +391,49 @@ class BaseSkeleton(object, metaclass=MetaBaseSkel):
                 bone.setSystemInitialized()
 
     @classmethod
-    def setBoneValue(cls, skelValues: t.Any, boneName: str, value: t.Any,
-                     append: bool = False, language: t.Optional[str] = None) -> bool:
+    def setBoneValue(
+        cls,
+        skelValues: SkeletonInstance,
+        boneName: str,
+        value: t.Any,
+        append: bool = False,
+        language: t.Optional[str] = None
+    ) -> bool:
         """
-            Allow setting a bones value without calling fromClient or assigning to valuesCache directly.
-            Santy-Checks are performed; if the value is invalid, that bone flips back to its original
+            Allows for setting a bones value without calling fromClient or assigning a value directly.
+            Sanity-Checks are performed; if the value is invalid, that bone flips back to its original
             (default) value and false is returned.
 
-            :param boneName: The Bone which should be modified
+            :param boneName: The name of the bone to be modified
             :param value: The value that should be assigned. It's type depends on the type of that bone
-            :param append: If true, the given value is appended to the values of that bone instead of
+            :param append: If True, the given value is appended to the values of that bone instead of
                 replacing it. Only supported on bones with multiple=True
-            :param language: Set/append which language
+            :param language: Language to set
+
             :return: Wherever that operation succeeded or not.
         """
         bone = getattr(skelValues, boneName, None)
+
         if not isinstance(bone, BaseBone):
-            raise ValueError(f"{boneName} is no valid bone on this skeleton ({skelValues})")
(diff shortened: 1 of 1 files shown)


sveneberth (dismissed):
Simple solution, thanks

src/viur/core/skeleton.py
```diff
@@ -409,6 +409,9 @@ def setBoneValue(cls, skelValues: t.Any, boneName: str, value: t.Any,
         if not isinstance(bone, BaseBone):
             raise ValueError(f"{boneName} is no valid bone on this skeleton ({skelValues})")
         skelValues[boneName]  # FIXME, ensure this bone is unserialized first
+        if value is None:
+            skelValues[boneName] = None
+            return True
```
phorward, line 414: Doesn't deserialization make sense when not being in the `None`-case? 
```suggestion
        if value is None:
            skelValues[boneName] = None
            return True

        _ = skelValues[boneName]  # ensure the bone is being unserialized first
```

phorward (changes requested):

phorward (changes requested):
I'm sorry, but there are two additional issues regarding this PR:

1. What about the `append`-parameter? IMHO the None-logic should only apply when append is False.
2. What about the `language`-parameter? Can you add additional logic to this, too?

phorward:
@ArneGudermann can you please re-check the two points stated?

ArneGudermann:
In #1047 @sveneberth says he want's to clear the bone and I think we should ignore the language and `append`.

phorward:
> In #1047 @sveneberth says he want's to clear the bone and I think we should ignore the language and `append`.

yeah, in his case calling `skel.setBoneValue("bla", None)`, this should be the case. But in `skel.setBoneValue("bla", None, append=True)` removing previously added data is not right, same as `skel.setBoneValue("bla", None, language="en")`, the caller assumes to clear the "en"-part of the bone only.

Please add these cases appropriately.

sveneberth:
I think it's good to only clear the value of a language with `lang`. But append `None` I think is basically nonsense and should raise an exception.

phorward (changes requested):
Hello @ArneGudermann, I'm still a little unhappy with this.

Here's an improved proposal on your PR: https://github.com/ArneGudermann/viur-core/pull/2

This entire function with its ugly name is a mess. `bone.setBoneValue` could also be the place to do it. Nevertheless, I think with my proposal (untested!) this might work better is is more integrative.

In VIUR4, this Gelumpe should be generally revised.

src/viur/core/skeleton.py
```diff
@@ -408,7 +408,16 @@ def setBoneValue(cls, skelValues: t.Any, boneName: str, value: t.Any,
...
         if not isinstance(bone, BaseBone):
             raise ValueError(f"{boneName} is no valid bone on this skeleton ({skelValues})")
-        skelValues[boneName]  # FIXME, ensure this bone is unserialized first
+        if value is None:
+            if append:
+                raise ValueError("Can not append None")
+            if language:
+                skelValues[boneName][language] = None
+            else:
+                skelValues[boneName] = None
```
sveneberth, line 417: Shouldn't this better be a list on multiple bones?
```suggestion
                skelValues[boneName][language] = [] if bone.multiple else None
            else:
                skelValues[boneName] = [] if bone.multiple else None
```

sveneberth (changes requested):

phorward (approved):

sveneberth (approved):

PR #1060, opened 2024-02-05
fix: patch_user_skel (fix for #983)

ArneGudermann:
Fix for #983


phorward (approved):
Hoppla, thanks for fixing this!

PR #1067, opened 2024-02-12
fix: `current.user` unset in deferred task calls

phorward:
Already released as v3.5.14 and tested in a live environment.

### DIFF
--- a/src/viur/core/modules/user.py
+++ b/src/viur/core/modules/user.py
@@ -1217,11 +1217,9 @@ class User(List):
         return getattr(self, "f2_%s" % cls.__name__.lower())
 
     def getCurrentUser(self):
-        # May be a deferred task
-        if not (session := current.session.get()):
-            return None
+        session = current.session.get()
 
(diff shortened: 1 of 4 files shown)


src/viur/core/modules/user.py
```diff
@@ -1217,11 +1217,9 @@ def secondFactorProviderByClass(self, cls) -> UserSecondFactorAuthentication:
         return getattr(self, "f2_%s" % cls.__name__.lower())
 
     def getCurrentUser(self):
-        # May be a deferred task
-        if not (session := current.session.get()):
-            return None
+        session = current.session.get()
 
-        if user := session.get("user"):
+        if session and (user := session.get("user")):
```
sveneberth, line 1222: Am I blind? But the way I see it, there's no logical change here, just a "design change", right?

sveneberth (approved):

src/viur/core/modules/user.py
phorward: right!

PR #1070, opened 2024-02-15
Changed inconsistent return value for delete function in JSON renderer.

theVAX:
We should provide a consistent result for all functions.


phorward (changes requested):
1. Respect PEP-8!
2. A standardization of the JSON/XML/MsgPack protocol is under considerations and depends on several preliminaries required for modular resolving. Therefore this pull request is only a drop in the ocean, and must be carefully harmonized with the other ViUR clients currently working out there.

phorward:
> We should provide a consistent result for all functions.

Please turn your pull request into an issue first, so that ideas can be collected. This is only one little part of a huger problem, where ideas are out there.

PR #1000, opened 2023-12-18
feat+refactor: `skel.dump()` for simplified value serialization (replacement for JSON value rendering)

ArneGudermann:
This PR moves the JSON rendering to the bones.
Resolves #978.

PR #1072, opened 2024-02-15
feat: `utils.json` module / improving `JsonBone`

### LINKED ISSUE #1064: Use custom decoder and encoder in `JsonBone`
I want to store a `datetime` object in a `JsonBone`. But this isn't serializable.
Can we use the `preprocessJsonObject` and `jsonDecodeObjectHook` from the `tasks` module here too?

phorward:
Proposal to fix #1064.

I'm still a little unhappy with this, as it bites with #1000, but after all it's a different use case. Maybe we can bring both together in some way.

### DIFF
--- a/src/viur/core/utils/json.py
+++ b/src/viur/core/utils/json.py
@@ -6,36 +6,37 @@ import typing as t
 from viur.core import db
 
 
-def _preprocess_json_object(obj: t.Any) -> t.Any:
+class ViURJsonEncoder(json.JSONEncoder):
     """
     Adds support for db.Key, db.Entity, datetime, bytes and and converts the provided obj
     into a special dict with JSON-serializable values.
     """
-    if isinstance(obj, db.Key):
-        return {".__key__": db.encodeKey(obj)}
-    elif isinstance(obj, datetime.datetime):
-        return {".__datetime__": obj.astimezone(pytz.UTC).isoformat()}
-    elif isinstance(obj, bytes):
-        return {".__bytes__": base64.b64encode(obj).decode("ASCII")}
-    elif isinstance(obj, db.Entity):
-        # TODO: Support Skeleton instances as well?
-        return {
-            ".__entity__": _preprocess_json_object(dict(obj)),
-            ".__ekey__": db.encodeKey(obj.key) if obj.key else None
-        }
-    elif isinstance(obj, dict):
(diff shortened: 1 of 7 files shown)


ArneGudermann (dismissed):

src/viur/core/utils/json.py
```diff
@@ -0,0 +1,66 @@
...
+        return [_preprocess_json_object(x) for x in obj]
+
+    return obj
+
+
+def dumps(obj: t.Any) -> str:
+    """
+    Wrapper for json.dumps() which converts additional ViUR datatypes.
+    """
+    return json.dumps(_preprocess_json_object(obj))
```
sveneberth, line 38:
```suggestion
def dumps(obj: t.Any, *args, **kwargs) -> str:
    """
    Wrapper for json.dumps() which converts additional ViUR datatypes.
    """
    return json.dumps(_preprocess_json_object(obj), *args, **kwargs)
```

sveneberth (changes requested):

ArneGudermann (approved):

src/viur/core/utils/json.py
```diff
@@ -0,0 +1,70 @@
...
+            return {".__bytes__": base64.b64encode(obj).decode("ASCII")}
+        elif isinstance(obj, db.Entity):
+            # TODO: Handle SkeletonInstance as well?
+            return {
+                ".__entity__": self.default(dict(obj)),
+                ".__key__": db.encodeKey(obj.key) if obj.key else None
+            }
+        elif isinstance(obj, dict):
+            return {self.default(key): self.default(value) for key, value in obj.items()}
+        elif isinstance(obj, (list, tuple, set)):
```
sveneberth, line 29: Maybe we should memorize the type. That's helpful for task chunks which want to use `set`s.

sveneberth (changes requested):

ArneGudermann:
Can we fix this https://github.com/viur-framework/viur-core/blob/471bee591af7c1620357bf1a7c24f03abb1bfffc/src/viur/core/bones/relational.py#L17-L21 with this PR too.

src/viur/core/utils/json.py
phorward: see 72203f7149146e55065ef032685f35c6f21759b7

sveneberth (approved):

ArneGudermann (approved):

PR #1073, opened 2024-02-19
fix: Several improvements on `ModuleConf`

phorward:
- Refactoring
- `listFilter` that falls back to any modules the current user is allowed to see
- `canEdit` which evaluates the `{module}-manage` access right
- Moved `read_all_modules` as staticmethod into ModuleConf


theVAX:
thanks a lot !

PR #1074, opened 2024-02-21
fix: `read_all_modules` after  #1073 and #1037 and merge 2a2b76ec16

sveneberth:
fix `read_all_modules` after  #1073 and #1037 and merge 2a2b76ec16


phorward (approved):
Thanks a lot!

PR #1075, opened 2024-02-21
fix: Make ViUR runable again after the extremely security-relevant PR #1037

sveneberth:
Make ViUR, especially the user system runable again after the _extremely_ security-relevant and _non-breaking_ PR #1037


phorward (approved):
Perfect!

PR #1076, opened 2024-02-22
feat: Provide `default_order` for `List` and `Tree`

### LINKED ISSUE #1011: Pre-defined ordering for `List` and `Tree` prototypes
Regardless of the setting in `adminInfo`, it might be wanted that a `List` or `Tree` module should always order (or maybe also filter) queries exactly in the same way, except when another order is given as `orderby` and `orderdir` parameter, or pre-defined in a `db.Query` before.

Currently, this behavior can be achieved by setting a `listFilter` this way:
```python
    def listFilter(self, query):
        query = super().listFilter(query)

        # enforce order by serial when no other filters, orders or search was specified
        if (query
                # and not query.queries.filters  # in case there's no other filter given
                and not query.queries.orders
                and not current.request.get().kwargs.get("search")
        ):
            query.order("serial")

        return query
```
This enforces all queries to order by "serial", in case no other order or a "search"-parameter is given within the request. The commented line might also add a rule to set this order only when no other filter is given.

This issue was now made as this might be a built-in feature by `List` and `Tree` prototypes, so one can define or overwrite the default order (in case of `Tree`, order by "sortindex" is default!), and all the above cases are (either configurable) checked before, and modified on each request.

phorward:
To avoid the creation of unsecure and error-prone `listFilter`s, this approach tries to fix #1011 and improves the ViUR system generally:

1. Trees default to sort by "sortindex"
2. The `File`-module defaults sorts by "name"
3. The `User`-module defaults to sort by "name" as well

### DIFF
--- a/src/viur/core/prototypes/list.py
+++ b/src/viur/core/prototypes/list.py
@@ -14,10 +14,17 @@ class List(SkelModule):
         The list module prototype handles datasets in a flat list. It can be extended to filters and views to provide
         various use-cases.
 
-        Definitely, it is the mostly-used prototype in any ViUR project.
+        It is undoubtedly the most frequently used prototype in any ViUR project.
     """
     handler = "list"
     accessRights = ("add", "edit", "view", "delete", "manage")
+    default_order: str | tuple[str, db.SortOrder] | None = None
+    """
+    Allows to specify a default order for this module, which is applied when no other order is specified.
+
+    Setting a default_order might result in the requirement of additional indexes, which are being raised
+    and must be specified.
+    """
 
     def viewSkel(self, *args, **kwargs) -> SkeletonInstance:
         """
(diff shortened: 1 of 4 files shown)


ArneGudermann (approved):

sveneberth (changes requested):
- [x] make `default_list` Callable too
- [x] Order a `List` module always, since we have always the `name` bone. 
  - [x] But If there's a `sortindex` bone, this bone should be used.

src/viur/core/prototypes/skelmodule.py
```diff
@@ -1,8 +1,14 @@
...
+from viur.core import Module, db
 from viur.core.skeleton import skeletonByKind, Skeleton, SkeletonInstance
 import typing as t
 
 
+ORDER_TYPE = str | tuple[str, db.SortOrder] | None
+"""
+Type for sort order definitions.
+"""
+
```
sveneberth, line 10:
```suggestion
DEFAULT_ORDER_TYPE = ORDER_TYPE | t.Callable[[db.Query], ORDER_TYPE]
```

src/viur/core/prototypes/tree.py
```diff
@@ -47,6 +47,14 @@ class Tree(SkelModule):
     nodeSkelCls = None
     leafSkelCls = None
 
+    default_order: ORDER_TYPE | t.Callable[[db.Query], ORDER_TYPE] = "sortindex"
```
sveneberth, line 50:
```suggestion
    default_order: DEFAULT_ORDER_TYPE = "sortindex"
```

src/viur/core/prototypes/list.py
```diff
@@ -14,11 +14,20 @@ class List(SkelModule):
         The list module prototype handles datasets in a flat list. It can be extended to filters and views to provide
         various use-cases.
 
-        Definitely, it is the mostly-used prototype in any ViUR project.
+        It is undoubtedly the most frequently used prototype in any ViUR project.
     """
     handler = "list"
     accessRights = ("add", "edit", "view", "delete", "manage")
 
+    default_order: ORDER_TYPE | t.Callable[[db.Query], ORDER_TYPE] = \
```
sveneberth, line 22:
```suggestion
    default_order: DEFAULT_ORDER_TYPE = \
```

sveneberth (changes requested):

sveneberth (approved):

ArneGudermann (approved):

PR #1086, opened 2024-02-29
feat: Implement `utils.parse.timedelta`

sveneberth:


src/viur/core/utils/parse.py
```diff
@@ -2,6 +2,7 @@
 ViUR utility functions regarding parsing.
 """
 import typing as t
+from datetime import timedelta as td
```
ArneGudermann, line 5:
```suggestion
from datetime import timedelta as td
```
Please use `import datetime`.

phorward (approved):

PR #1090, opened 2024-03-05
Feature/on action

sveneberth:
Usage
```py
class MyModule(List):
    @on_edited
    @on_added
    def entry_observer(self, skel):
        print(f"Entry changed: {skel = }")
```

What do u think of this concept?


phorward:
> What do u think of this concept?

Quite interesting. This looks like a more dynamic replacement for #972.

PR #1095, opened 2024-03-11
fix: `tr_key` maybe empty in existing database

phorward:
There is a VIUR project which already uses translations, so there are existing entities, but the tr_key does not exist. This fix handles this specific case.


sveneberth:
Did the deferred task not work for you?

https://github.com/viur-framework/viur-core/blob/develop/src/viur/core/i18n.py#L439

phorward:
The problem was not reproducible, therefore we're closing it for now.

PR #1102, opened 2024-03-14
fix: User roles: Automatic "view" right when "edit" or "delete" is provided

phorward:
The `File`-module has default setting "add", "edit" for editor-role, which didn't work. One correction would be to simply add "view" there, but this method automatically adds "view" when "edit" or "delete" is given.

### DIFF
--- a/src/viur/core/modules/user.py
+++ b/src/viur/core/modules/user.py
@@ -210,6 +210,10 @@ class UserSkel(skeleton.Skeleton):
                             if right in module.accessRights:
                                 access.add(f"{name}-{right}")
 
+                                # special case: "edit" and "delete" actions require "view" as well!
+                                if right in ("edit", "delete") and "view" in module.accessRights:
+                                    access.add(f"{name}-view")
+
             skel["access"] = list(access)
 
         return super().toDB(skel, *args, **kwargs)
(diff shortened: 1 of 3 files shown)


ArneGudermann (approved):

src/viur/core/modules/user.py
```diff
@@ -210,6 +210,10 @@ def toDB(cls, skel, *args, **kwargs):
                             if right in module.accessRights:
                                 access.add(f"{name}-{right}")
 
+                                # special case: "edit" and "delete" actions require "view" as well!
+                                if right in ("edit", "delete") and "view" in module.accessRights:
```
sveneberth, line 214: Shouldn't this be `not in`?

```suggestion
                                if right in ("edit", "delete") and "view" not in module.accessRights:
```
or an `or`?

```suggestion
                                if right in ("edit", "delete") or "view" in module.accessRights:
```

sveneberth (changes requested):

src/viur/core/modules/user.py
phorward: Nope, because "view" should only be set when it's a valid access right for the module. The condition is correct, and tested. The `access` variable in this context is a set already.

PR #1103, opened 2024-03-15
feat: add `skel.update` and `skel.__ior__`

### LINKED ISSUE #1098: Implement `.update(dict_values)` on `Skeleton`/`SkeletonInstance`
provide a `dict` or kwargs to merge multiple values into skel (like [`dict.update()`](https://docs.python.org/3/library/stdtypes.html#dict.update))

ArneGudermann:
resolves #1098


sveneberth (changes requested):

src/viur/core/skeleton.py
```diff
@@ -302,12 +302,29 @@ def __str__(self) -> str:
...
+            # We're not overriding the key
+            for key, value in other.items():
+                new_entity[key] = value
+            self.setEntity(new_entity)
+        elif isinstance(other, SkeletonInstance):
+            new_entity = self.dbEntity or db.Entity()
+            # We're not overriding the key
+            for key, value in other.dbEntity.items():
+                new_entity[key] = value
+            self.setEntity(new_entity)
```
sveneberth, line 320: In this case I would merge the `accessedValues` values from other into self. If other has been changed, but not written this might be different to the dbEntity.

sveneberth (changes requested):

phorward:
Same as #1147, `Skeleton` could be inherited from `dict` or `db.Enitity` to provide all `dict` features. What do you think? @ArneGudermann @sveneberth

src/viur/core/skeleton.py
```diff
@@ -319,11 +319,8 @@ def __ior__(self, other: dict | SkeletonInstance | db.Entity) -> SkeletonInstanc
                 new_entity[key] = value
             self.setEntity(new_entity)
         elif isinstance(other, SkeletonInstance):
-            new_entity = self.dbEntity or db.Entity()
-            # We're not overriding the key
-            for key, value in other.dbEntity.items():
-                new_entity[key] = value
-            self.setEntity(new_entity)
+            for key, value in other.accessedValues.items():
```
sveneberth, line 322: But, why not all values and only accessed value?

ArneGudermann: You mean the values form the `dbEntity` and the accessedValues ?

sveneberth (approved):

phorward (approved):

PR #1104, opened 2024-03-15
Add readonly for Skeleton

### LINKED ISSUE #1100: Provide a `Skeleton.readonly()` function
Shortcut, to immediately obtain a readonly skeleton.

Avoid things like these...
```py
    def baseSkel(self):
        # Make all bones readOnly!
        # FIXME: There should be a skel.readonly() function soon...
        skel = super().baseSkel().clone()

        for bone in skel.values():
            bone.readOnly = True

        return skel
```

ArneGudermann:
resolves #1100

### DIFF
--- a/src/viur/core/skeleton.py
+++ b/src/viur/core/skeleton.py
@@ -242,7 +242,7 @@ class SkeletonInstance:
         # Load a @classmethod from the Skeleton class and bound this SkeletonInstance
         elif item in {"fromDB", "toDB", "all", "unserialize", "serialize", "fromClient", "getCurrentSEOKeys",
                       "preProcessSerializedData", "preProcessBlobLocks", "postSavedHandler", "setBoneValue",
-                      "delete", "postDeletedHandler", "refresh"}:
+                      "delete", "postDeletedHandler", "refresh", "readonly"}:
             return partial(getattr(self.skeletonCls, item), self)
         # Load a @property from the Skeleton class
         try:
@@ -524,6 +524,15 @@ class BaseSkeleton(object, metaclass=MetaBaseSkel):
 
             _ = skel[key]  # Ensure value gets loaded
             bone.refresh(skel, key)
+    @classmethod
+    def readonly(cls, skel: SkeletonInstance):
+        """
+            Set all bones to readonly in the Skeleton.
+        """
+        for bone in skel.values():
+            if not isinstance(bone, BaseBone):
+                continue
+            bone.readOnly = True
 
     def __new__(cls, *args, **kwargs) -> SkeletonInstance:
         return SkeletonInstance(cls, *args, **kwargs)


sveneberth:
I rather imagine a readOnly state at skeleton level here.

So on the one hand an attribute that can be set at skeleton level.
Then the value is inherited when rendering (but not set in the bone), so that you can deactivate the readOnly state again and know which bones were originally readonly.
And the readonly should then also prevent a toDB. And possibly this should also be mentioned when rendering in the renderer. (maybe this could be interesting later for cases where an edit is not possible but a view is).

phorward (changes requested):
Hello @ArneGudermann, thanks for facing this issue.

1. I would prefer a `skel = skel.readonly()` assignment, which is similar to `skel.ensure_is_cloned()`
2. The pull request has conflicts, which must be merged
3. I would like to move this to viur-core 3.8.

src/viur/core/skeleton.py
```diff
@@ -430,6 +434,11 @@ def ensure_is_cloned(self):
 
         return self
 
+    def readonly(self):
+        skel = self.ensure_is_cloned()
```
sveneberth, line 438: Why does this skeleton to be cloned? It does not change the bone map or any other attribute. IMO we can set the attribute on any `SkeletonInstance`.

src/viur/core/skeleton.py
```diff
@@ -1220,6 +1229,7 @@ def write(
                 return cls.toDB(skel, update_relations=update_relations)
 
         assert skel.renderPreparation is None, "Cannot modify values while rendering"
+        assert not skel.is_readonly, "Cannot write a readonly Skeleton"
```
sveneberth, line 1232: Please do not use `assert` for validations.

sveneberth (changes requested):
Additionally I think we need to consider the `is_readonly` here in `fromClient`
https://github.com/viur-framework/viur-core/blob/d3556b05147bd43b5c8821fa5ca2f7a35293235f/src/viur/core/skeleton.py#L653
to sth. like 
```python
            if (ignore is None and (bone.readOnly or self.is_readonly)) or key in (ignore or ()):

```

src/viur/core/skeleton.py
```diff
@@ -1220,6 +1228,8 @@ def write(
                 return cls.toDB(skel, update_relations=update_relations)
 
         assert skel.renderPreparation is None, "Cannot modify values while rendering"
+        if skel.is_readonly:
+            raise ValueError("Cannot modify wirte data when the skeleton is readonly")
```
phorward, line 1232: A readonly-flagged skeleton can still be modified and written by program code.
The idea of a readonly-skeleton is just that it cannot be read from client, as all bones are readonly.
```suggestion
```

src/viur/core/skeleton.py
```diff
@@ -155,6 +156,7 @@ def __init__(
         clone: bool = False,
         # FIXME: BELOW IS DEPRECATED!
         clonedBoneMap: t.Optional[t.Dict[str, BaseBone]] = None,
+        readonly: bool = False,
```
phorward, line 159: Parameter description in docstring below is missing.

phorward (changes requested):
Hello @ArneGudermann, I find your proposal an interesting and memory-saving alternative to my first idea. But in such a case, the bones should all be flagged (or pseudo-flagged) as readonly, too.

The idea is: Give the user a read-only version of a skeleton, just to review it. 

...

Hmm well after some more consieration... the idea would also be, to get a readonly skel and flag only one bone as writable. That was also my initial proposal. I really would like to replace this pull request by a version that modifies the bones... maybe your initial version was the right solution already.

ArneGudermann:
@phorward You tagged it already. Lets discussed this in the viur-meeting.

PR #1106, opened 2024-03-15
fix: Improvement `conf["viur.paramFilterFunction"]`

skoegl:
Remove the static filter that removes any parameters passed, starting with "_", to a configurable filter function which provides more flexibility.

### DIFF
--- a/src/viur/core/config.py
+++ b/src/viur/core/config.py
@@ -196,7 +196,7 @@ conf = Conf({
     "viur.maxPostParamsCount": 250,
 
     # Function which decides if a request param should be used or filtered out
-    "viur.paramFilterFunction": lambda key, value: True,
+    "viur.paramFilterFunction": lambda key, value: key.startswith("_"),
 
     # Describing the internal ModuleConfig-module
     "viur.moduleconf.admin_info": {
--- a/src/viur/core/config.py
+++ b/src/viur/core/config.py
@@ -195,6 +195,9 @@ conf = Conf({
     # Upper limit of the amount of parameters we accept per request. Prevents Hash-Collision-Attacks
     "viur.maxPostParamsCount": 250,
 
+    # Function which decides if a request param should be used or filtered out
+    "viur.paramFilterFunction": lambda key, value: True,
+
     # Describing the internal ModuleConfig-module
     "viur.moduleconf.admin_info": {
         "icon": "icon-settings",
--- a/src/viur/core/request.py
+++ b/src/viur/core/request.py
@@ -437,7 +437,7 @@ class Router:
 
         param_filter = conf["viur.paramFilterFunction"]
         if param_filter and not callable(param_filter):
-           raise ValueError(f"""{param_filter=} is not callable""")
+            raise ValueError(f"""{param_filter=} is not callable""")
 
         for key, value in self.request.params.items():
             try:
--- a/src/viur/core/request.py
+++ b/src/viur/core/request.py
@@ -435,6 +435,7 @@ class Router:
                 f" of {conf['viur.maxPostParamsCount']} allowed arguments per request"
             )
 
+        param_filter = conf["viur.paramFilterFunction"]
         for key, value in self.request.params.items():
             try:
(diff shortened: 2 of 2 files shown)


src/viur/core/config.py
```diff
@@ -195,6 +195,9 @@ def __setitem__(self, key, value):
     # Upper limit of the amount of parameters we accept per request. Prevents Hash-Collision-Attacks
     "viur.maxPostParamsCount": 250,
 
+    # Function which decides if a request param should be used or filtered out
+    "viur.paramFilterFunction": lambda key, value: True,
```
ArneGudermann, line 199:
```suggestion
    "viur.paramFilterFunction": None,
```

src/viur/core/request.py
```diff
@@ -444,7 +445,7 @@ def _route(self, path: str) -> None:
                 # someone tries to exploit unicode normalisation bugs)
                 raise errors.BadRequest()
 
-            if key.startswith("_"):  # Ignore keys starting with _ (like VI's _unused_time_stamp)
+            if not param_filter(key, value):
```
ArneGudermann, line 448:
```suggestion
            if param_filter and param_filter(key, value):
```
So we not call the lamda all the times

src/viur/core/request.py
```diff
@@ -435,6 +435,7 @@ def _route(self, path: str) -> None:
                 f" of {conf['viur.maxPostParamsCount']} allowed arguments per request"
             )
 
+        param_filter = conf["viur.paramFilterFunction"]
```
ArneGudermann, line 438:
```suggestion
        param_filter = conf["viur.paramFilterFunction"]
        if param_filter and not callable(param_filter):
           raise ValueError(f"""{param_filter=} is not callable""")
```

ArneGudermann (changes requested):

src/viur/core/config.py
sveneberth, line 199: We should keep the old behavior.
```suggestion
    "viur.paramFilterFunction": lambda key, value: key.startswith("_"),
```

sveneberth (changes requested):

src/viur/core/request.py
```diff
@@ -435,6 +435,9 @@ def _route(self, path: str) -> None:
...
+        if param_filter and not callable(param_filter):
+           raise ValueError(f"""{param_filter=} is not callable""")
```
phorward, line 440:
```suggestion
           raise ValueError(f"""{param_filter=} is not callable""")

```

phorward (approved):

src/viur/core/request.py
```diff
@@ -435,6 +435,10 @@ def _route(self, path: str) -> None:
...
```
phorward, line 440:
```suggestion
            raise ValueError(f"""{param_filter=} is not callable""")
```

src/viur/core/request.py
```diff
@@ -435,6 +435,10 @@ def _route(self, path: str) -> None:
                 f" of {conf['viur.maxPostParamsCount']} allowed arguments per request"
             )
 
+        param_filter = conf["viur.paramFilterFunction"]
+        if param_filter and not callable(param_filter):
+            raise ValueError(f"""{param_filter=} is not callable""")
```
sveneberth, line 440: I think this check could be moved to a better place where it is not called on every request. But okay for `main`. On `develop`, we can validate it in the `config` object.

sveneberth (approved):

src/viur/core/request.py
phorward: >  On `develop`, we can validate it in the `config` object.

Can you either tell me how, or make the merge?

sveneberth: Maybe with an `@property` like https://github.com/viur-framework/viur-core/blob/80b93191df5d6d8c426ac5181f8d6696c666c168/src/viur/core/config.py#L717-L747

In the longer term, we could also look to validate the values in setattr against the type hints.

PR #1107, opened 2024-03-15
feat: store compute value on unserialize

### LINKED ISSUE #709: Store computed values on unserialize
However, the value is still recalculated for each view/list.
IMO, as soon as the value is recalculated in `unserialize` (by the compute function), it should also be written to the database. In general, you have more read than write OPS. For example, if you look at a folder and have a compute bone for the amount of children, it would recalculate the value on every view. Only when I rename the folder would the value be cached. In my opinion, this is inefficient and does not make sense for a value cached by lifetime.

_Originally posted by @sveneberth in https://github.com/viur-framework/viur-core/pull/639#pullrequestreview-1358111635_
            

additionally add the possibility to flush existing calculations.

### Addition:
The [ComputeMethod.Lifetime](https://github.com/viur-framework/viur-core/blob/486d500c0e226feac67c28bdd3d2a6d0f27c4fb9/src/viur/core/bones/base.py#L143) or a new method like `LifetimeCache` should work as a cache, regardless of whether it's currently being read or written. But it should also be possible to force the recalculation like `myskel.mybone.compute_now()`

ArneGudermann:
resolves #709


src/viur/core/bones/base.py
```diff
@@ -743,7 +743,17 @@ def unserialize(self, skel: 'viur.core.skeleton.SkeletonInstance', name: str) ->
...
-                        skel.accessedValues[name] = self._compute(skel, name)
+                        value = self._compute(skel, name)
+
+                        def transact():
+                            db_obj = db.Get(skel["key"])
+                            db_obj[f"_viur_compute_{name}_"] = now
+                            db_obj[name] = value
+                            db.Put(db_obj)
+
+                        db.RunInTransaction(transact)
```
sveneberth, line 754:
```suggestion
                        if db.IsInTransaction():
                            transact()
                        else:
                            db.RunInTransaction(transact)
```

sveneberth (changes requested):

sveneberth:
But, this does not work for a computed `RelationalBone`, because this is not calling `.serialize()`

sveneberth (dismissed):
So far this seems to work, I observe this for bone types and then we can do a follow-up fix.

---

@phorward  Can we move this to [3.6](https://github.com/viur-framework/viur-core/milestone/1)?

phorward:
> @phorward Can we move this to [3.6](https://github.com/viur-framework/viur-core/milestone/1)?

Is this still valid? IMHO we can do it.

sveneberth (approved):

phorward (approved):

PR #1109, opened 2024-03-15
feat: `default_order` should support multiple orders as well

sveneberth:
e.g.:
```py
    default_order: DEFAULT_ORDER_TYPE = (
        ("firstname", db.SortOrder.Ascending),
        ("lastname", db.SortOrder.Ascending),
    )
```

### DIFF
--- a/src/viur/core/prototypes/list.py
+++ b/src/viur/core/prototypes/list.py
@@ -187,6 +187,7 @@ class List(SkelModule):
         if query := self.listFilter(self.viewSkel().all().mergeExternalFilter(kwargs)):
             # Apply default order when specified
             if self.default_order and not query.queries.orders and not current.request.get().kwargs.get("search"):
+                # TODO: refactor: Duplicate code in prototypes.Tree
                 if callable(default_order := self.default_order):
                     default_order = default_order(query)
 
--- a/src/viur/core/prototypes/list.py
+++ b/src/viur/core/prototypes/list.py
@@ -190,7 +190,7 @@ class List(SkelModule):
                 if callable(default_order := self.default_order):
                     default_order = default_order(query)
 
-                if isinstance(default_order, tuple) and all(isinstance(pair, tuple) for pair in default_order):
+                if isinstance(default_order, tuple | list) and all(isinstance(pair, tuple) and len(pair) == 2 for pair in default_order):
                     query.order(*default_order)
                 elif default_order:
                     query.order(default_order)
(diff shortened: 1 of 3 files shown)


src/viur/core/prototypes/list.py
```diff
@@ -190,7 +190,9 @@ def list(self, *args, **kwargs) -> t.Any:
                 if callable(default_order := self.default_order):
                     default_order = default_order(query)
 
-                if default_order:
+                if isinstance(default_order, tuple) and all(isinstance(pair, tuple) for pair in default_order):
```
phorward, line 193:
```suggestion
                if isinstance(default_order, tuple | list) and all(isinstance(pair, tuple) and len(pair) == 2 for pair in default_order):
```

phorward (changes requested):
Hello @sveneberth,

this is a nice extension, anyway please fix the definition of `DEFAULT_ORDER_TYPE` in `skelmodule.py` accordingly.

I think when the Query class gets back to viur-core again, we should  re-define the order-tuple to a namedtuple or dataclass, so that it has a clear definition.

phorward (approved):

src/viur/core/prototypes/list.py
```diff
@@ -190,7 +190,10 @@ def list(self, *args, **kwargs) -> t.Any:
                 if callable(default_order := self.default_order):
                     default_order = default_order(query)
 
-                if default_order:
+                if isinstance(default_order, tuple | list) and all(isinstance(pair, tuple) and len(pair) == 2
+                                                                   for pair in default_order):
+                    query.order(*default_order)
+                elif default_order:
                     query.order(default_order)
```
phorward, line 197: Wait a moment... I think this is wrong, and your initial attempt is better. This is all ugly. A type should be better, as already mentioned here:
> I think when the Query class gets back to viur-core again, we should re-define the order-tuple to a namedtuple or dataclass, so that it has a clear definition.

Here's my alternative suggestion:
```suggestion
                if default_order:
                    if (
                        isinstance(default_order, str)
                        or (
                            isinstance(default_order, tuple)
                            and len(default_order) == 2
                            and isinstance(default_order[0], str)
                            and isinstance(default_order[1], db.SortOrder)
                        )
                    ):
                        query.order(default_order)
                    else:
                        query.order(*default_order)
```

phorward (changes requested):
@sveneberth please apply the default_order check like this also to the `Tree`-prototype. Please test your changes, also with the Tree-module (e.g. File-Module).

I've tried to shorten it, but `isinstance(default_order, SINGLE_ORDER_TYPE)` cannot be executed at runtime, therefore this is the "best" choice.

phorward (approved):

PR #1118, opened 2024-03-19
feat: Provide `LoginSkel` on `UserPassword.login`

phorward:
- Implements action skel `LoginSkel` correctly with the right action
- Modifies PasswordBone to not render tests when test_threshold is set to 0, which is a security improvement, as it is not rendered to non-authenticated users

### DIFF
--- a/src/viur/core/bones/password.py
+++ b/src/viur/core/bones/password.py
@@ -3,10 +3,9 @@ The PasswordBone class is a specialized version of the StringBone class designed
 data. It hashes the password data before saving it to the database and prevents it from being read
 directly. The class also includes various tests to determine the strength of the entered password.
 """
-
 import hashlib
 import re
-
+import typing as t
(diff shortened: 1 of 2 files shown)


phorward:
Hi, I've tested it, works fine.

ArneGudermann (approved):

src/viur/core/bones/password.py
```diff
@@ -58,22 +57,25 @@ class PasswordBone(StringBone):
         (r"^.{8,}$", translate("core.bones.password.too_short",
                                defaultText="The password is too short. It requires for at least 8 characters."), True),
     )
-    """Provides tests based on regular expressions to test the password stength."""
+    """Provides tests based on regular expressions to test the password stength.
```
phorward, line 60:
```suggestion
    """Provides tests based on regular expressions to test the password strength.
```

PR #1121, opened 2024-03-22
added functionality to set a subset of countries, updated documentation

cheeaza:
There was no possibility to limit the set of available countries in a selectcountryBone.
I added the parameter countries to limit it.  I also updated the docstring.

### DIFF
--- a/src/viur/core/bones/selectcountry.py
+++ b/src/viur/core/bones/selectcountry.py
@@ -748,14 +748,15 @@ ISO2TOISO3 = {v: k for k, v in ISO3TOISO2.items()}  # Build the invert map
 
 
 class SelectCountryBone(SelectBone):
-    """
+     """
     A bone representing a country selection input field in a web application or form.
 
     The SelectCountryBone is designed to provide a user-friendly way to select a country from a predefined list of
     countries. It inherits from the BaseBone class and extends it to support country-specific functionalities,
     such as displaying country names and handling country codes (e.g., ISO 3166-1 alpha-2 or alpha-3).
 
-    :params List[str] countries: A list of countries supported by the bone, typically represented by their codes.
+    :params str codes: Parameter to switch between ISO 3166-1 alpha-2 and alpha-3. Represented as 2 or 3. (default 2)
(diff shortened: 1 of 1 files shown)


src/viur/core/bones/selectcountry.py
```diff
@@ -755,7 +755,9 @@ class SelectCountryBone(SelectBone):
     countries. It inherits from the BaseBone class and extends it to support country-specific functionalities,
     such as displaying country names and handling country codes (e.g., ISO 3166-1 alpha-2 or alpha-3).
 
-    :params List[str] countries: A list of countries supported by the bone, typically represented by their codes.
+    :params str codes: Parameter to switch between ISO 3166-1 alpha-2 and alpha-3. Represented as 2 or 3. (default 2)
+    :params dict countries: Dictionary with subset of countries from ISO 3166-1 list.
+    Represented as {'de':'Germany', 'en':'English'}
```
phorward, line 760: This is not consistent with the signature of the function below!

phorward (changes requested):
Hello @cheeaza,

thanks for creating this pull request. Your implementation might work, but I think the `SelectCountryBone` is implemented very bad and old fashioned, and should generally be refactored so that it gets back closer to the `SelectBone`.

1. The current globals `ISO2CODES` and `ISO3CODES` should become part of the SelectCountryBone class, so they can be accessed via `SelectCountryBone.ISO2` and `SelectCountryBone.ISO3`.
2. The signature of the `__init__` function should become
   ```py
   def __init__(
       self,
       *,
       descr="Country",  # default description
       values=SelectCountryBone.ISO2,  # default code set
       **kwargs
    ):
   ```
3. Afterwards, the behavior defaults to the same as before. In case one wants to filter only specific countries, this can be done with a simple comprehension, like
   ```py
   deliver_to = SelectCountryBone(
       values={
           k: v for k, v in SelectCountryBone.ISO2 if k in ("de", "en")
       }
   )
   ```
4. The current `codes`-parameter could still be supported with a deprecation warning.

I hope you're with me in this case. Would you like to propose an alternative to this pull request?

phorward:
Replaced by #1342 and due lack of activity.

PR #1125, opened 2024-03-26
fix: Improving several `User` auth method handling

phorward:
- Provides `GoogleLogin` only when `conf.user.google_client_id` is set
- Allows for `None`-providers, which are simply ignored
- Temporarily removes deprecation warning for `getAuthMethods`, which produces *not* the same result as `login()`, but should be replaced by `login()` soon.
- Docstrings

### DIFF
--- a/src/viur/core/modules/user.py
+++ b/src/viur/core/modules/user.py
@@ -1016,7 +1016,6 @@ class AuthenticatorOTP(UserSecondFactorAuthentication):
         """
         return bool(skel.dbEntity.get("otp_app_secret", ""))
 
-
     @classmethod
     def patch_user_skel(cls, skel_cls):
         """


ArneGudermann (approved):

src/viur/core/modules/user.py
```diff
@@ -1138,22 +1140,41 @@ class User(List):
     verifyEmailAddressMail = "user_verify_address"
     passwordRecoveryMail = "user_password_recovery"
 
-    authenticationProviders: list[UserAuthentication] = [
+    authenticationProviders: t.Iterable[UserPrimaryAuthentication] = (
         UserPassword,
-        GoogleAccount
-    ]
+        conf.user.google_client_id and GoogleAccount,
+    )
```
sveneberth, line 1146: I find it inappropriate to allow None here. There should only be real classes here, as required by the type hint. Checking this every time at runtime in the methods with an if is on the one hand unnecessary duplicate code but also avoidable performance losses.

Simply filter out the values:

```py
    authenticationProviders: t.Iterable[UserPrimaryAuthentication] = tuple(filter(None(
        UserPassword,
        conf.user.google_client_id and GoogleAccount,
    ))
```

src/viur/core/modules/user.py
```diff
@@ -126,12 +126,14 @@ def __new__(cls):
         authentication methods.
         """
         for provider in conf.main_app.vi.user.authenticationProviders:
-            assert issubclass(provider, UserPrimaryAuthentication)
-            provider.patch_user_skel(cls)
+            if provider:
```
sveneberth, line 129: Please remove these checks

phorward:
@sveneberth I've applied your suggestions.

sveneberth (approved):

PR #1126, opened 2024-03-27
    feat: Implement `translation_key_prefix` for `SelectBone` values

sveneberth:
Currently only the label is used as `translation` key, but also not in
every case. This PR changes allows to set a custom prefix or use a
included method to generator a prefix from the skeleton and bone name.

Furthermore context is added via translation hint.

This is especially interesting for enums, where you cannot set a custom label.

### DIFF
--- a/src/viur/core/bones/base.py
+++ b/src/viur/core/bones/base.py
@@ -187,6 +187,12 @@ class BaseBone(object):
     type = "hidden"
     isClonedInstance = False
 
+    _owner = None
+    """Skeleton class to which this bone instance belongs"""
+
+    _name = None
+    """Name of this bone (attribute name in the skeletons containing this bone)"""
+
     def __init__(
         self,
         *,
@@ -293,6 +299,10 @@ class BaseBone(object):
             self._prevent_compute = False
 
         self.compute = compute
+
+    def __set_name__(self, owner: t.Any, name: str) -> None:
+        self._owner = owner
+        self._name = name
 
     def setSystemInitialized(self):
         """
--- a/src/viur/core/bones/base.py
+++ b/src/viur/core/bones/base.py
@@ -16,6 +16,9 @@ import typing as t
 from viur.core import db, utils
 from viur.core.config import conf
 
+if t.TYPE_CHECKING:
+    from ..skeleton import Skeleton
+
 __system_initialized = False
 """
 Initializes the global variable __system_initialized
@@ -187,10 +190,10 @@ class BaseBone(object):
     type = "hidden"
     isClonedInstance = False
 
-    _owner = None
+    skel_cls = None
     """Skeleton class to which this bone instance belongs"""
 
-    _name = None
+    name = None
     """Name of this bone (attribute name in the skeletons containing this bone)"""
 
     def __init__(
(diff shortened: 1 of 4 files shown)


src/viur/core/bones/base.py
```diff
@@ -187,6 +187,12 @@ class BaseBone(object):
     type = "hidden"
     isClonedInstance = False
 
+    _owner = None
```
phorward, line 190:
```suggestion
    _skel_cls = None
```
Can you rename this into `skel_cls`?

src/viur/core/bones/base.py
```diff
@@ -294,6 +300,10 @@ def __init__(
 
         self.compute = compute
 
+    def __set_name__(self, owner: t.Any, name: str) -> None:
```
phorward, line 303:
```suggestion
    def __set_name__(self, owner: `Skeleton`, name: str) -> None:
```

src/viur/core/bones/select.py
```diff
@@ -20,17 +20,23 @@
...
+
+
+def translation_key_prefix_skeleton_bonename(bones_instance: BaseBone) -> str:
+    """Generate a translation key prefix based on the skeleton name"""
+    return f'skeleton.{bones_instance._owner.__name__.lower().removesuffix("skel")}.{bones_instance._name}.'
+
+
+def translation_key_prefix_bonename(bones_instance: BaseBone) -> str:
+    """Generate a translation key prefix based on the skeleton and bone name"""
+    return f'skeleton.{bones_instance._owner.__name__.lower().removesuffix("skel")}.{bones_instance._name}.'
```
phorward, line 33: Why isn't this defined in `BaseBone`?

src/viur/core/bones/select.py
```diff
@@ -44,14 +50,26 @@ def __init__(
...
-
+        """
+        Initializes a new SelectBone.
+
+        :param defaultValue: key(s) of the values which will be checked by default.
+        :param values: dict of key->value pairs from which the user can choose from
+            -- or a callable that returns a dict.
+        :param translation_key_prefix: A prefix for the key of the translation object.
+            It is empty by default, so that only the label (dict value) from the values is used.
+            A static string or dynamic method can be used (like `translation_key_prefix_bonename`).
```
phorward, line 64: Doesn't this make sense in BaseBone as well?

phorward (changes requested):

src/viur/core/bones/select.py
sveneberth: But other bones doesn't have `values`.

src/viur/core/bones/select.py
sveneberth: Because only the `SelectBone` has predefined `values`

src/viur/core/bones/base.py
sveneberth: Okay, that obviously depends on your point of view. I'm taking the general case first: A bone can be packed into any python object. In the ViUR context, of course, this will usually only be a skeleton.

src/viur/core/bones/base.py
sveneberth: Done in afca4d1b.

I've also removed the underscore, as in the methods it's ok to access it from outside and there's no name collision.

phorward (approved):

ArneGudermann (approved):

PR #1128, opened 2024-04-02
fix: Add param `_call_deferred`

sveneberth:
Calls the `@CallDeferred` decorated method directly. This is for example necessary, to call a super method which is decorated with `@CallDeferred`.

### DIFF
--- a/src/viur/core/tasks.py
+++ b/src/viur/core/tasks.py
@@ -470,6 +470,20 @@ def CallDeferred(func: t.Callable) -> t.Callable:
     _call_deferred: Calls the @CallDeferred decorated method directly.
         This is for example necessary, to call a super method which is decorated with @CallDeferred.
 
+    ..  code-block:: python
+
+        # Example for use of the _call_deferred-parameter
+        class A(Module):
+            @CallDeferred
+            def task(self):
+                ...
+
+        class B(A):
+            @CallDeferred
+            def task(self):
+                super().task(_call_deferred=False)  # avoid secondary deferred call
+                ...
+
     See also:
         https://cloud.google.com/python/docs/reference/cloudtasks/latest/google.cloud.tasks_v2.types.Task
         https://cloud.google.com/python/docs/reference/cloudtasks/latest/google.cloud.tasks_v2.types.CreateTaskRequest
--- a/src/viur/core/tasks.py
+++ b/src/viur/core/tasks.py
@@ -1,22 +1,23 @@
 import abc
 import datetime
 import functools
-import grpc
 import json
 import logging
 import os
-import requests
 import sys
 import time
 import traceback
 import typing as t
+
+import grpc
+import requests
 from google import protobuf
 from google.cloud import tasks_v2
+
 from viur.core import current, db, errors, utils
 from viur.core.config import conf
 from viur.core.decorators import exposed, skey
 from viur.core.module import Module
-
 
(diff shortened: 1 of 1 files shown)


src/viur/core/tasks.py
```diff
@@ -466,6 +467,9 @@ def CallDeferred(func: t.Callable) -> t.Callable:
     _target_version: Specify a version on which to run this task.
         By default, a task will be run on the same version where the wrapped method has been called.
 
+    _call_deferred: Calls the @CallDeferred decorated method directly.
+        This is for example necessary, to call a super method which is decorated with @CallDeferred.
+
```
phorward, line 472: Can you please provide a piece of example code in the docstring?

src/viur/core/tasks.py
```diff
@@ -478,11 +482,12 @@ def CallDeferred(func: t.Callable) -> t.Callable:
     def make_deferred(func, self=__undefinedFlag_, *args, **kwargs):
         # Extract possibly provided task flags from kwargs
         queue = kwargs.pop("_queue", "default")
+        call_deferred = kwargs.pop("_call_deferred", True)
```
phorward, line 485: Can't these be directly put as arguments of `make_deferred`?

phorward (changes requested):

src/viur/core/tasks.py
sveneberth: I'm not sure what kind of code you image. It's just an argument for the method call

src/viur/core/tasks.py
sveneberth: I agree, that would be cleaner. But it's not done that way with the other arguments either. They should all be in the same style, otherwise it's a confusion. I would do this in a later refactoring.

src/viur/core/tasks.py
phorward: I had this case nowhere yet, therefore it should be helpful to others to show such a use-case and how the parameter is used.

src/viur/core/tasks.py
phorward: Therefore marked lines 484 to 485

phorward: OK let's do this in another PR.

src/viur/core/tasks.py
sveneberth: Something like this?
```py
        class A(Module):
            @CallDeferred
            def task(self):
                ...
        class B(A):
            @CallDeferred
            def task(self):
                super().task(_call_deffered=False)
                ...
            
            @exposed
            def run(self):
                self.task()
```

src/viur/core/tasks.py
sveneberth: Okay, I've opened #1130

src/viur/core/tasks.py
phorward: yes, this is helpful to understand how the parameter shall be used.

phorward:
```suggestion
    _call_deferred: Calls the @CallDeferred decorated method directly.
        This is for example necessary, to call a super method which is decorated with @CallDeferred.

    ..  code-block:: python

        # Example for use of the _call_deferred-parameter
        class A(Module):
            @CallDeferred
            def task(self):
                ...

        class B(A):
            @CallDeferred
            def task(self):
                super().task(_call_deferred=False)  # avoid secondary deferred call
                ...
    ..  

```

phorward (approved):

PR #1129, opened 2024-04-02
fix: Pass arguments in `JsonBone.__init__()` to `super()` call

sveneberth:
Otherwise the bone is indexed and crashes at the 1500 bytes index limit in case of large data

### DIFF
--- a/src/viur/core/bones/json.py
+++ b/src/viur/core/bones/json.py
@@ -30,7 +30,7 @@ class JsonBone(RawBone):
         schema: t.Mapping = {},
         *args, **kwargs
     ):
-        super().__init__(*args, **kwargs)
+        super().__init__(*args, indexed=indexed, multiple=multiple, languages=languages, **kwargs)
         assert not multiple
         assert not languages
         assert not indexed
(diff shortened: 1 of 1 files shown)


ArneGudermann (approved):

src/viur/core/bones/json.py
```diff
@@ -30,7 +30,7 @@ def __init__(
         schema: t.Mapping = {},
         *args, **kwargs
     ):
-        super().__init__(*args, **kwargs)
+        super().__init__(*args, indexed=indexed, multiple=multiple, languages=languages, **kwargs)
         assert not multiple
         assert not languages
         assert not indexed
```
phorward:
```suggestion
        # JsonBone is bound to certain limits
        assert not multiple
        assert not languages
        assert not indexed

        super().__init__(*args, indexed=indexed, multiple=multiple, languages=languages, **kwargs)
```

phorward (approved):
@sveneberth I've updated the `JsonBone`'s entire `__init__` into a more convenient style including your change within your branch. This looks fine to me.

Issue #1117, opened 2024-03-19
Implement `UidBone`


sveneberth:
@ArneGudermann we talked about this already in a viur-shop meeting. ([Usecase](https://github.com/viur-framework/viur-shop/blob/9aeb5536a9e9fbd348c359a7f4f3520b909d5f0a/src/viur/shop/skeletons/order.py#L56-L61))

This bone should generate a unique, new identifier according to a specifc scheme

phorward:
Fixed by #1131

PR #1131, opened 2024-04-02
feat: Add UidBone

ArneGudermann:
Proposal for #1117

### DIFF
--- a/src/viur/core/bones/uid.py
+++ b/src/viur/core/bones/uid.py
@@ -1,0 +1,90 @@
+import secrets
+import string
+import warnings
+
+import logging
+import typing as t
+
+from viur.core import current, db, utils
+from viur.core.bones.base import BaseBone, ReadFromClientError, ReadFromClientErrorSeverity, UniqueValue, \
+    UniqueLockMethod
+
+
+class UidBone(BaseBone):
+    """
+    The "StringBone" represents a data field that contains text values.
+    """
+    type = "str"
+
+    def __init__(
+        self,
+        *,
+        length: int | None = 13,
+        pattern: str | None = "*",
+        generate_fn: t.Callable | None = None,
+        unique=UniqueValue(UniqueLockMethod.SameValue, False,"Unique Value already in use"),
+        **kwargs
+    ):
+        """
+        Initializes a new UidBone.
+
+
+        :param length: The length allowed for values of this bone.
+        :param kwargs: Inherited arguments from the BaseBone.
+        """
+        # fixme: Remove in viur-core >= 4
+
+        super().__init__(unique=unique, **kwargs)
+
+        if self.multiple or self.languages:
+            raise ValueError("UidBone cannot be multiple or translated")
+        if not self.readOnly:
+            self.readOnly = True
+            # raise ValueError("UidBone must be readOnly")
+
+        self.length = length
+        self.pattern = pattern
+        self.generate_fn = generate_fn
+
+    def serialize(self, skel: 'SkeletonInstance', name: str, parentIndexed: bool) -> bool:
+        logging.error("seri")
+        logging.error(skel.accessedValues[name])
+
+        if super().serialize(skel, name, parentIndexed):
+            return True
+        else:
+            logging.error("generate new id")
+            skel.dbEntity[name] = self.generate_uid()
+            return True
+
+    def unserialize(self, skel: 'viur.core.skeleton.SkeletonInstance', name: str) -> bool:
+        logging.error("unseri")
+        logging.error(skel.dbEntity[name])
+        if name in skel.dbEntity:
+            skel.accessedValues[name] = skel.dbEntity[name]
+        else:
+            skel.accessedValues[name] = self.generate_uid()
+
+        return True
+
+    def generate_uid(self):
+        if "*" in self.pattern and "?" not in self.pattern and "#" not in self.pattern:
+            return self.pattern.replace("*", utils.string.random(self.length))
+        res = self.pattern
+        remaining_chars = self.length
+        if "?" in self.pattern:
+            while "?" in res:
+                remaining_chars -= 1
+                res = res.replace("?", utils.string.random(1), 1)
+        if "#" in self.pattern:
+            while "#" in res:
+                remaining_chars -= 1
+                res = res.replace("#", secrets.choice(string.digits), 1)
+        return res.replace("*", utils.string.random(remaining_chars))
+
+    def structure(self) -> dict:
+        ret = super().structure() | {
+            "length": self.length,
+            "pattern": self.pattern
+        }
+        return ret
--- a/src/viur/core/bones/__init__.py
+++ b/src/viur/core/bones/__init__.py
@@ -32,6 +32,7 @@ from .string import StringBone
 from .text import TextBone
 from .treeleaf import TreeLeafBone
 from .treenode import TreeNodeBone
+from .uid import UidBone
 from .user import UserBone
 
 # Expose only specific names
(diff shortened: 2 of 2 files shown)


ArneGudermann:
Some Points for the viur meeting

1. should the value for an add be rendered out?
2. readonly ?
3. lang/multiple allowed?

src/viur/core/bones/uid.py
```diff
@@ -0,0 +1,92 @@
...
+    """
+    The "UidBone" represents a data field that contains text values.
+    """
+    type = "uid"
+
+    def __init__(
+        self,
+        *,
+        generate_fn: t.Callable = generate_uid,
+        fill_chars="",
```
sveneberth, line 52: [`str.rjust`](https://docs.python.org/3/library/stdtypes.html#str.rjust) takes only one char. So please rename this to singular `fill_char` — and maybe we should follow the python naming and name it `fillchar` too?

```suggestion
        fillchar:str="",
```

src/viur/core/bones/uid.py
```diff
@@ -0,0 +1,92 @@
...
+            unique=UniqueValue(UniqueLockMethod.SameValue, False, "Unique Value already in use"),
+            **kwargs
+        )
+
+        if self.multiple or self.languages:
+            raise ValueError("UidBone cannot be multiple or translated")
+        if not self.readOnly:
+            self.readOnly = True
+
+        self.fill_chars = str(fill_chars)
```
sveneberth, line 78: Like the parameter the instance attribute as well. See above.
```suggestion
        self.fillchar = str(fillchar)
```

src/viur/core/bones/uid.py
```diff
@@ -0,0 +1,92 @@
...
+        self.length = length
+        if isinstance(pattern, t.Callable):
+            pattern = pattern()
+        self.pattern = str(pattern)
+        if self.pattern.count("*") != 1:
+            raise ValueError("Only one Wildcard (*) is allowed in the pattern")
+
+    def structure(self) -> dict:
+        ret = super().structure() | {
+            "fill_chars": self.fill_chars,
```
sveneberth, line 88:
```suggestion
            "fillchar": self.fillchar,
```

sveneberth (changes requested):

ArneGudermann:
mhh i have found an interesting bug. So if I call up a skel view in which this bone exists but it has not yet been written, a uid is automatically assigned but not written in the skel. This leads to the number not being used.

ArneGudermann:
> mhh i have found an interesting bug. So if I call up a skel view in which this bone exists but it has not yet been written, a uid is automatically assigned but not written in the skel. This leads to the number not being used.

Fixed by: https://github.com/viur-framework/viur-core/pull/1131/commits/041da3b979822982825205ec6e612873fe02976a

sveneberth (dismissed):
Worked in my test, but I didn't any pentest. But tests in production are usually the best test anyways :laughing:

src/viur/core/bones/uid.py
```diff
@@ -0,0 +1,97 @@
...
+        if not self.readOnly:
+            self.readOnly = True
+
+        self.fillchar = str(fillchar)
+        self.length = length
+        if isinstance(pattern, t.Callable):
+            pattern = pattern()
+        self.pattern = str(pattern)
+        if self.pattern.count("*") != 1:
+            raise ValueError("Only one Wildcard (*) is allowed in the pattern")
```
phorward, line 87:
```suggestion
            raise ValueError("Only one wildcard (*) is allowed and required in the pattern")
```

src/viur/core/bones/uid.py
```diff
@@ -0,0 +1,97 @@
...
+        if len(self.fillchar) != 1:
+            raise ValueError("Only one char is allowed as fillchar")
```
phorward, line 89: The ValueError is also raised on an empty fillchar, which is the default. Is this wanted?

src/viur/core/bones/uid.py
```diff
@@ -0,0 +1,97 @@
...
+        super().__init__(
+            compute=Compute(fn=generate_fn, interval=ComputeInterval(ComputeMethod.Once)),
+            unique=UniqueValue(UniqueLockMethod.SameValue, False, "Unique Value already in use"),
+            **kwargs
+        )
+
+        if self.multiple or self.languages:
+            raise ValueError("UidBone cannot be multiple or translated")
+        if not self.readOnly:
+            self.readOnly = True
```
phorward, line 79: Can you solve this differently? Default `readOnly` flag shall be True, and changing it should raise a ValueError. The user should know that this bone can't be set `readOnly=False`.

src/viur/core/bones/uid.py
```diff
@@ -0,0 +1,97 @@
...
+        generate_fn: t.Callable = generate_uid,
+        fillchar: str = "",
+        length: int = 13,
+        pattern: str | t.Callable | None = "*",
+        **kwargs
+    ):
+        """
+        Initializes a new UidBone.
+
+
```
phorward, line 63:
```suggestion
```

phorward (changes requested):

phorward (approved):
Looks ok to me. What about @sveneberth ?

sveneberth (dismissed):
> Looks ok to me. What about @sveneberth ?

still fine

src/viur/core/bones/uid.py
```diff
@@ -0,0 +1,99 @@
...
+            compute=Compute(fn=generate_fn, interval=ComputeInterval(ComputeMethod.Once)),
+            unique=UniqueValue(UniqueLockMethod.SameValue, False, "Unique Value already in use"),
+            readOnly=readOnly,
+            **kwargs
+        )
+        if self.multiple or self.languages:
+            raise ValueError("UidBone cannot be multiple or translated")
+
+        if self.readOnly:
+            raise ValueError("UidBone must be read-only")
```
phorward, line 81: I'm still a little bit confused by the code: Is the `UidBone()` now always readOnly or is it not allowed to set it readOnly? Why is this not allowed?
```suggestion
        if self.readOnly:
            raise ValueError("UidBone must not be read-only")
```

My wish suggestion would be to remove the entire readOnly-stuff, as its on the developer if a bone is readonly or not.

phorward (changes requested):
Sorry, we have to clarify this.

src/viur/core/bones/uid.py
sveneberth:
```suggestion
        if not self.readOnly:
            raise ValueError("UidBone must be read-only")
```

src/viur/core/bones/uid.py
```diff
@@ -0,0 +1,99 @@
...
+        :param fillchar The char that are filed in when the uid has not the length.
+        :param length: The length allowed for values of this bone.
+        :param pattern: The pattern for this Bone. "*" will be replaced with the uid value.
+        :param kwargs: Inherited arguments from the BaseBone.
+        """
+
+        super().__init__(
+            compute=Compute(fn=generate_fn, interval=ComputeInterval(ComputeMethod.Once)),
+            unique=UniqueValue(UniqueLockMethod.SameValue, False, "Unique Value already in use"),
+            readOnly=readOnly,
```
sveneberth, line 74: isn't this covered by the kwargs?
```suggestion
        **kwargs
    ):
        """
        Initializes a new UidBone.

        :param generate_fn: The compute function to calculate the unique value,
        :param fillchar The char that are filed in when the uid has not the length.
        :param length: The length allowed for values of this bone.
        :param pattern: The pattern for this Bone. "*" will be replaced with the uid value.
        :param kwargs: Inherited arguments from the BaseBone.
        """

        super().__init__(
            compute=Compute(fn=generate_fn, interval=ComputeInterval(ComputeMethod.Once)),
            unique=UniqueValue(UniqueLockMethod.SameValue, False, "Unique Value already in use"),
```

phorward (approved):

PR #1132, opened 2024-04-03
fix: Improve `List.default_order` to respect languages config

phorward:
The current default_order caused, that entries didn't show up when there was a name-bone which had a languages-setting, so it needed to be manually corrected to e.g. `name.de`. This PR fixes the behavior to either use the current or the first language. Its still possible to redefine default_order in any style like befor.

### DIFF
--- a/src/viur/core/prototypes/list.py
+++ b/src/viur/core/prototypes/list.py
@@ -4,6 +4,7 @@ from viur.core import current, db, errors, utils
 from viur.core.decorators import *
 from viur.core.cache import flushCache
 from viur.core.skeleton import SkeletonInstance
+from viur.core.bones import BaseBone
 from .skelmodule import SkelModule, DEFAULT_ORDER_TYPE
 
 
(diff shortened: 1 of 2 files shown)


ArneGudermann (approved):

src/viur/core/prototypes/list.py
```diff
@@ -192,6 +207,8 @@ def list(self, *args, **kwargs) -> t.Any:
                     default_order = default_order(query)
 
                 if default_order:
+                    logging.debug(f"Applying {default_order=}")
```
sveneberth, line 210: Are these debugs indented or forgotten to be removed after development?

src/viur/core/prototypes/list.py
```diff
@@ -19,14 +20,28 @@ class List(SkelModule):
     handler = "list"
     accessRights = ("add", "edit", "view", "delete", "manage")
 
-    default_order: DEFAULT_ORDER_TYPE = \
-        lambda _self, query: next((prop for prop in ("sortindex", "name") if prop in query.srcSkel), None)
-    """
-    Allows to specify a default order for this module, which is applied when no other order is specified.
+    def default_order(self, query: db.Query, bone_order: t.Iterable[str] = ("sortindex", "name")) -> DEFAULT_ORDER_TYPE:
```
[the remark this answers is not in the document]
phorward: Well, this is for a special-purpose I had in my testing environment: I have a skeleton that contains both a sortindex and a name-bone with languages. The sortindex is used in a different context, but the module itself should sort by name. To allow for sorting by the current language setting by name, I made it this way to allow for a super-call:
```py
    def default_order(self, query):
        return super().default_order(query, ("name",))
```

src/viur/core/prototypes/list.py
phorward: I would like to keep these, as they might help to find unwanted behavior faster, especially when porting projects to viur-core 3.6.

src/viur/core/prototypes/list.py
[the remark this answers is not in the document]
sveneberth: okay

src/viur/core/prototypes/list.py
sveneberth: okay

sveneberth (approved):

PR #1137, opened 2024-04-05
fix: Replace old dict `conf` access with attribute access

sveneberth:


phorward (approved):

phorward (approved):
Huppalla... I think this came in through a merge.

PR #1140, opened 2024-04-11
feat Add session `setdefault`

ArneGudermann:
Resolves https://github.com/viur-framework/viur-shop/blob/f6e84e772be01d7cf2beb74d674a3aeda6ac303b/src/viur/shop/modules/abstract.py#L45

### DIFF
--- a/src/viur/core/session.py
+++ b/src/viur/core/session.py
@@ -211,8 +211,7 @@ class Session:
         """
             Set a default value in the current session
         """
-        self.session.setdefault(key, value)
-        return value
+        return self.session.setdefault(key, value)
 
 
 @tasks.CallDeferred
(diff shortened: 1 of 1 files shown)


src/viur/core/session.py
```diff
@@ -207,6 +207,13 @@ def items(self) -> 'dict_items':
         """
         return self.session.items()
 
+    def setdefault(self, key: str, value: t.Any) -> t.Any:
+        """
+            Set a default value in the current session
+        """
+        self.session.setdefault(key, value)
+        return value
```
sveneberth, line 215: Cloud be more simple, and this returns the correct value if this key was already in the session.
```suggestion
        return self.session.setdefault(key, value)
```

sveneberth (changes requested):

sveneberth (approved):

phorward (approved):
It's okay for me. We generally should think about subclassing `Session` from `db.Entity`, which is a subclass of dict, so that the session can be used like an ordinary dict in all situations.

PR #1141, opened 2024-04-11
feat Add `charset` for `StringBone`

ArneGudermann:
Resolves https://github.com/viur-framework/viur-shop/blob/f6e84e772be01d7cf2beb74d674a3aeda6ac303b/src/viur/shop/skeletons/discount_condition.py#L115

### DIFF
--- a/src/viur/core/bones/string.py
+++ b/src/viur/core/bones/string.py
@@ -17,6 +17,7 @@ class StringBone(BaseBone):
         self,
         *,
         caseSensitive: bool = True,
+        charset: str = "",
         max_length: int | None = 254,
         min_length: int | None = None,
         natural_sorting: bool | t.Callable = False,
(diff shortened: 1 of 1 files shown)


src/viur/core/bones/string.py
```diff
@@ -123,6 +126,10 @@ def isInvalid(self, value):
             return "Maximum length exceeded"
         if self.min_length is not None and len(value) < self.min_length:
             return "Minimum length not reached"
+
+        if self.charset:
+            if not all([char in self.charset for char in value]):
```
sveneberth, line 131: This might by a little bit faster, because this will fail fast as generator with `any`.
```suggestion
            if any(char not in self.charset for char in value):
```

sveneberth (changes requested):

sveneberth (approved):

phorward (changes requested):
Hi there... I'm a little unhappy with this extension, because it can also be simply resolved using a `vfunc`-lambda, like this:
```python
blah = StringBone(
    required=True,
    descr="Blah",
    vfunc=lambda value:
        "Not all letters are available in the charset"
        if any(char not in "abcdefg" for char in value)
        else None,
)
```
We should discuss about this.

phorward:
This should be solved differently: Provide some helper functions in StringBone which can be inserted as vfuncs.

phorward:
Substituted by #1183

PR #1146, opened 2024-04-26
fix: Enfore serialized values are always strings in the datastore

sveneberth:


src/viur/core/bones/string.py
```diff
@@ -60,41 +66,70 @@ def __init__(
...
 
         :param value: The serialized value to unserialize.
         :return: The unserialized value.
         """
         if isinstance(value, dict) and "val" in value:
-            return value["val"]
-        elif value:
+            value = value["val"]  # Process with the raw value
+        if value:
             return str(value)
```
sveneberth, line 128: We could discuss about this one. What happens if there's bullsh*t in the datastore, e.g. after someone changed bone params from multi to single or it was a RelationalBone, ...

phorward: Leave it as it is now.

phorward (approved):

ArneGudermann (approved):

PR #1153, opened 2024-05-06
refactor: Sub-class `Session` from `db.Entity` to behave `dict`-compliant

### LINKED ISSUE #1147: Subclass `Session` from `db.Entity` to inherit all dict-features
> We generally should think about subclassing `Session` from `db.Entity`, which is a subclass of dict, so that the session can be used like an ordinary dict in all situations.

_Originally posted by @phorward in https://github.com/viur-framework/viur-core/pull/1140#pullrequestreview-2027977727_

KadirBalku:
Resolves #1147, replaces #1140.


phorward (changes requested):
Hello @KadirBalku, thanks for your first pull request!

I've added some improvement suggestions.
Can you test your branch with other projects than "viur-trial" please, just for confidence.

phorward (changes requested):
Hello @KadirBalku,

have you tested it with other projects?

And can you please take a look at the functions I've commented. Maybe they must be kept, or the internally use `__setitem__`. In latter case, all is fine, and the `changed`-flag is being set properly. 

Take a look into the [Python docs](https://docs.python.org/3/library/stdtypes.html#typesmapping) to get an overview about all dict-related functions.

phorward (changes requested):
@KadirBalku looks good to me!
Please fix the unused import issue, then we can merge.

src/viur/core/session.py
```diff
@@ -1,5 +1,5 @@
 import datetime
-import hmac
+import hmac  # is this necessary ?
```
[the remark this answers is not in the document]
KadirBalku: the comment is added by me, so i can remove it without hesitation ?

src/viur/core/session.py
```diff
@@ -73,7 +72,11 @@ def load(self, req):
                     return False
 
                 self.cookie_key = cookie_key
-                self.session = data["data"]
+
+                # Here we have to use the pure db.Entity function to avoid changing the 'changed' flag
+                db.Entity.clear(self)
+                db.Entity.update(self, data["data"])
```
sveneberth, line 78: I think a `super()` call would be more pythonic?
```suggestion
                super().clear()
                super().update(data["data"])
```

src/viur/core/session.py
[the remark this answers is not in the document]
sveneberth: yes, this is unnecessary and can be removed.

sveneberth (changes requested):
What about `Session.pop()` and where is `Session.setdefault()` from https://github.com/viur-framework/viur-core/pull/1140?

src/viur/core/session.py
```diff
@@ -164,6 +166,29 @@ def reset(self) -> None:
...
+        super().__ior__(other)
+        self.changed = True
+        return self
+
+    def update(self, other: dict) -> None:
+        """
+        Merges the contents of a dict into the session.
+        """
+        self |= other
+
```
ArneGudermann, line 190:
```suggestion
        self |= other
        self.changed = True
```
We change the session data here to.

ArneGudermann (changes requested):

src/viur/core/session.py
phorward: No, because `|=`-operator triggers __ior__, which sets the changed flag.

phorward:
> What about `Session.pop()` and where is `Session.setdefault()` from #1140?

@KadirBalku `Session.pop()` must be implemented.
@sveneberth `Session.setdefault()` is provided by sub-classing `db.Entity`, which sub-classes `dict`.

src/viur/core/session.py
KadirBalku: applied and tested, worked.

src/viur/core/session.py
```diff
@@ -199,20 +161,37 @@ def reset(self) -> None:
...
+    def update(self, other: dict) -> None:
         """
-            Returns all items in the current session.
+        Merges the contents of a dict into the session.
         """
-        return self.session.items()
+        self |= other
 
-    def setdefault(self, key: str, value: t.Any) -> t.Any:
+    def pop(self, key: str) -> None:
```
sveneberth, line 189: This is not how `pop()` of a container usually works. Take a look at [`dict.pop()` in python docs](https://docs.python.org/3/library/stdtypes.html#dict.pop). It must return the pop'ed element or the default value (if provided)

sveneberth (changes requested):

src/viur/core/session.py
KadirBalku: I have updated #1153

src/viur/core/session.py
```diff
@@ -199,20 +161,48 @@ def reset(self) -> None:
...
+        default = kwargs.get("default", None)
+
+        if default is None:
+            ret = super().pop(key)
+        else:
+            ret = super().pop(key, default)
+
+        self.changed = True
+
+        return ret
```
sveneberth, line 205: In your implementation `default` cannot be provided as positional argument and you allow more argument than necessary.
I would work in this case with an sentinel.


```py
_SENTINEL : t.Final[object] = object()
```


```suggestion
    def pop(self, key: str, default=_SENTINEL) -> t.Any:
        """
        Delete a specified key from the session.
        
        If key is in the session, remove it and return its value, else return default. 
        If default is not given and key is not in the session, a KeyError is raised.
        """
        if key in self or default is _SENTINEL:
            value = super().pop(key)
            self.changed = True
            return value
        return default
```

The `KeyError` will be thrown through the super call

sveneberth (changes requested):

phorward:
@KadirBalku we should discuss about dropping Python 3.10 support with viur-core>=3.7

src/viur/core/session.py
```diff
@@ -199,20 +163,46 @@ def reset(self) -> None:
...
+        self |= other
+
+    def pop(self, key: str, default=_SENTINEL) -> t.Any:
+        """
+        Delete a specified key from the session.
+
+        If key is in the session, remove it and return its value, else return default.
+        If default is not given and key is not in the session, a KeyError is raised.
+        """
+        print(f"{_SENTINEL=}")
```
sveneberth, line 198:
```suggestion
```

phorward (approved):

sveneberth (approved):
LGTM, But we should discuss the python 3.10 issue first, before we merge this. So let's wait until Friday.

phorward (approved):

sveneberth (approved):

ArneGudermann (approved):

PR #1156, opened 2024-05-08
feat: View script by path

phorward:
When providing permalinks to Scriptor scripts across multiple projects, it's better to either use the path of the script rather then the key. This is also useful for the script runner and for scriptor actions integrated somewhere.

Additionally allows to view a script by its path, e.g. `http://localhost:8080/json/script/view/leaf/export.py` or `http://localhost:8080/json/script/view/leaf?key=folder/sub/result.py`

### DIFF
--- a/src/viur/core/modules/script.py
+++ b/src/viur/core/modules/script.py
@@ -1,6 +1,7 @@
 import re
+import typing as t
 from viur.core.bones import *
-from viur.core.prototypes.tree import Tree, TreeSkel
+from viur.core.prototypes.tree import Tree, TreeSkel, SkelType
 from viur.core import db, conf, current, skeleton, tasks, errors
 from viur.core.decorators import exposed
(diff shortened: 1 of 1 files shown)


src/viur/core/modules/script.py
```diff
@@ -98,6 +97,17 @@ def getAvailableRootNodes(self):
 
         return [{"name": "Scripts", "key": self.ensureOwnModuleRootNode().key}]
 
+    @exposed
+    def view(self, skelType, key, *args, **kwargs):
```
sveneberth, line 101:
```suggestion
    def view(self, skelType: SkelType, key: db.Key | int | str, *args, **kwargs) -> t.Any:
```

sveneberth (changes requested):

sveneberth (approved):

Issue #1063, opened 2024-02-09
A module with `_` can be routed under different names


sveneberth:
I have just found out that I can access the `shipping_config` module with `_` and also with `-`.
So
- `/vi/shop/shipping-config`
- `/vi/shop/shipping_config`

I don't know if this is intentional, it seems strange to me. But it doesn't cause any problems for now.

PR #1157, opened 2024-05-08
fix: Python identifier mapping in request `Router`

phorward:
The `sitemap.xml` routing in several ViUR projects is broken. This re-enables support of parts containing dots within module/method routing.

### DIFF
--- a/src/viur/core/request.py
+++ b/src/viur/core/request.py
@@ -493,11 +493,11 @@ class Router:
                 raise errors.Unauthorized()
 
             idx += 1
-            part = part.replace("-", "_")
+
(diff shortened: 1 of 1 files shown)


sveneberth (changes requested):
I removed this for reasons in https://github.com/viur-framework/viur-core/pull/865 ...

phorward:
> I removed this for reasons in #865 ...

Then please talk to @icubes and tell him how to provide a `sitemap.xml` generated from a template, which worked in viur-core prior 3.6.

phorward:
OK the routing will be done manually in this case, so the code won't be changed.

Furthermore, there will be a follow-up on this PR to entirely remove any replacements, to fix #1063.

PR #1159, opened 2024-05-10
fix: Support `/deploy/admin` folder as well

sveneberth:
With viur-cli v2 the vi-admin moved from `/deploy/vi/` to `/deploy/admin/`. Therefore the redirect from `HOST/vi` to `HOST/vi/s/main.html` does no longer work. Now we have to check both path.
(We could make this configurable in 3.7 ...)


phorward (approved):
Thanks for fixing this!

PR #1162, opened 2024-05-14
feat: Add `EmailTransportMailjet` to `email.py`

bb-mausbrand:


src/viur/core/email.py
```diff
@@ -449,3 +452,57 @@ def check_sib_quota() -> None:
...
+            email['headers'] = headers
+        if attachments:
+            atts = []
+            for att in attachments:
+                next_att = {
+                    "filename": att['filename'],
+                    "base64content": b64encode(att['content']).decode('ASCII')
+                }
+                if 'contenttype' in att:
+                    next_att['contenttype'] = att['contenttype']
```
sveneberth, line 498: Please use `mimetype` here as described in https://github.com/viur-framework/viur-core/blob/5416176fa03a3492a169517ea3861c8f2b7d3708/src/viur/core/email.py#L178
```suggestion
                if 'mimetype' in att:
                    next_att['contenttype'] = att['mimetype']
```
Btw: I you like you can change the type hint of attachment from `dict` to a [`TypedDict`](https://docs.python.org/3/library/typing.html#typing.TypedDict), this would be more precise.

src/viur/core/email.py
```diff
@@ -449,3 +452,57 @@ def check_sib_quota() -> None:
...
+            :return: split dict
+        """
+        posLt = address.rfind("<")
+        posGt = address.rfind(">")
+        if -1 < posLt < posGt:
+            email = address[posLt + 1:posGt]
+            sname = address.replace(f"<{email}>", "", 1).strip()
+            return {"name": sname, "email": email}
+        else:
+            return {"email": address}
```
sveneberth, line 473: This method, apart from the docstring, looks the same as `EmailTransportSendInBlue.splitAddress`. To avoid duplicate code I think it would be better to move the method to module level or into utils where both classes can use the same method.

sveneberth (changes requested):
Hey @bb-mausbrand, 
thanks for your contributing.

Since you are new, a few words about the syntax:
- We use a linter that runs via GitHub Action, locally you can start it with `pipenv run pep8check` — after you have installed your pipenv (you seem to have done that already). Please correct what he complains about :)
- We generally use double quotes (`"`), not single (`'`) for strings.

<br>

And a few comments on the content, in addition to the other comments:
- Please regenerate the `requirements.txt` once after you have changed packages. Otherwise they will not be in the python wheel later.
  ```sh
  pipenv install --dev
  pipenv requirements --hash > src/viur/core/requirements.txt
  ```
  @phorward We could think about making the `mailjet-rest` package an extra. That way it won't be installed if you don't need it.
- Please add `conf.email.mailjet_api_key` and `conf.email.mailjet_api_secret` to the config object.

Pipfile
```diff
@@ -32,11 +32,13 @@ urllib3 = "==1.26.18"  # for appengine-python-standard used by some projects (ht
 viur-datastore = "~=1.0"
 webob = "~=1.0"
 appengine-python-standard = "~=1.0"
+file-magic = "~=0.4.1"
+mailjet-rest = "~=1.3.4"
```
phorward, line 36: use at least minor version, patch level shall be updated automatically.
```suggestion
file-magic = "~=0.4"
mailjet-rest = "~=1.3"
```

phorward (approved):

phorward (changes requested):
>  @phorward We could think about making the `mailjet-rest` package an extra. That way it won't be installed if you don't need it.

@bb-mausbrand can you please find out how to define extras as dependencies, and maybe tie it together with the pipenv-based tooling (see [CONTRIBUTING.md](https://github.com/viur-framework/viur-core/blob/main/CONTRIBUTING.md#dependency-management) for a short explanation).

src/viur/core/email.py
```diff
@@ -14,6 +17,11 @@
 from viur.core.config import conf
 from viur.core.tasks import CallDeferred, DeleteEntitiesIter, PeriodicTask
 
+try:
+    import mailjet_rest
+except ModuleNotFoundError:
+    pass
```
phorward, line 23:
```suggestion
    mailjet_rest = None
```

src/viur/core/email.py
```diff
@@ -449,3 +458,41 @@ def check_sib_quota() -> None:
             entity["latest_warning_for"] = None
 
         db.Put(entity)
+
+
+if 'mailjet_rest' in sys.modules:
```
phorward, line 463:
```suggestion
if mailjet_rest:
```

src/viur/core/email.py
```diff
@@ -449,3 +458,41 @@ def check_sib_quota() -> None:
...
+            if cc:
+                email["cc"] = [EmailTransportMailjet.splitAddress(c) for c in cc]
+            if headers:
+                email["headers"] = headers
+            if attachments:
+                atts = []
+                for att in attachments:
+                    next_att = {
+                        "filename": att["filename"],
+                        "base64content": b64encode(att["content"]).decode("ASCII")
```
phorward, line 485: `base64` was already imported above, therefore its not needed to extra-import `b64encode`.
```suggestion
                        "base64content": base64.b64encode(att["content"]).decode("ASCII"),
```

src/viur/core/email.py
```diff
@@ -2,9 +2,12 @@
 import json
 import logging
 import os
+import sys
 from abc import ABC, abstractmethod
 import typing as t
 from urllib import request
+from base64 import b64encode
```
phorward, line 9: `base64` was already imported above, therefore its not needed to extra-import `b64encode`. See correction of b64encode-call below as well.
```suggestion
```

src/viur/core/email.py
```diff
@@ -449,3 +458,41 @@ def check_sib_quota() -> None:
...
+                atts = []
+                for att in attachments:
+                    next_att = {
+                        "filename": att["filename"],
+                        "base64content": b64encode(att["content"]).decode("ASCII")
+                    }
+                    if "mimetype" in att:
+                        next_att["contenttype"] = att["mimetype"]
+                    else:
+                        next_att["contenttype"] = magic.detect_from_content(att["content"]).mime_type
```
phorward, line 490: You can shorten this:
```suggestion
                        "contenttype": att.get("mimetype") or magic.detect_from_content(att["content"]).mime_type,
                    }
```

Pipfile
```diff
@@ -32,11 +32,13 @@ urllib3 = "==1.26.18"  # for appengine-python-standard used by some projects (ht
 viur-datastore = "~=1.0"
 webob = "~=1.0"
 appengine-python-standard = "~=1.0"
+file-magic = "~=0.4"
```
phorward, line 35: Isn't `file-magic` also a dependency required by the `mailjet`-feature only?

phorward (changes requested):
Hello @bb-mausbrand,

I've found some issues regarding the final pull request. Can you check them out?

phorward (dismissed):

sveneberth (changes requested):

src/viur/core/requirements.txt
```diff
@@ -24,9 +25,10 @@ grpcio==1.63.0 --hash=sha256:01799e8649f9e94ba7db1aeb3452188048b0019dc37696b0f5c
...
 idna==3.7; python_version >= '3.5' --hash=sha256:028ff3aadf0609c1fd278d8ea3089299412a7a8b9bd005dd08b9f8285bcb5cfc --hash=sha256:82fee1fc78add43492d3a1898bfa6d8a904cc97d8427f683ed8e798d07761aa0
-jinja2==3.1.3; python_version >= '3.7' --hash=sha256:7d6d50dd97d52cbc355597bd845fabfbac3f551e1f99619e39a35ce8c370b5fa --hash=sha256:ac8bd6544d4bb2c9792bf3a159e80bba8fda7f07e81bc3aed565432d5925ba90
+jinja2==3.1.4; python_version >= '3.7' --hash=sha256:4a3aee7acbbe7303aede8e9648d13b8bf88a429282aa6122a993f0ac800cb369 --hash=sha256:bc5dd2abb727a5319567b7a813e6a2e7318c39f4f487cfe6c89c6f9c7d25197d
 jsonschema==4.22.0; python_version >= '3.8' --hash=sha256:5b22d434a45935119af990552c862e5d6d564e8f6601206b305a61fdf661a2b7 --hash=sha256:ff4cfd6b1367a40e7bc6411caec72effadd3db0bbe5017de188f2d6108335802
 jsonschema-specifications==2023.12.1; python_version >= '3.8' --hash=sha256:48a76787b3e70f5ed53f1160d2b81f586e4ca6d1548c5de7085d1682674764cc --hash=sha256:87e4fdf3a94858b8a2ba2778d9ba57d8a9cafca7c7489c46ba0d30a8bc6a9c3c
+mailjet-rest==1.3.4 --hash=sha256:635d53ac3fd61020f309c24ee977ae3458654ab39f9c36fc4b50c74e5d8ad410 --hash=sha256:e02663fa0369543bcd48c37a146e8143bb12b9f3512af2d5ba6dfbcc99e64a2d
```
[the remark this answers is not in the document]
phorward: @sveneberth will it work now?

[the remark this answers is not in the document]
sveneberth: Yes, installation with (`pipenv install -e viur-core[mailjet]`) and without extra worked as editable.
I think directly adding this to `setup.cfg` was the simplest solution.

sveneberth (approved):

PR #1165, opened 2024-05-17
fix: Correct `default_order` for not case-sensitive `StringBone`s

sveneberth:
Without the `.idx` suffix an empty result is returned


phorward (changes requested):
Hm, indeed this is a possible fix, but on the other side the order setting should be generally handled by the bones, and not in the prototypes, like in `BaseBone.buildDBSort`. 

There's already pull request #1077 on this, which can be useful. Maybe it would be better to allow for a `dict` here as well, which is being fed to `Query.mergeExternalFilter()`, which then handles the ordering correctly?

sveneberth:
> Hm, indeed this is a possible fix, but on the other side the order setting should be generally handled by the bones, and not in the prototypes, like in `BaseBone.buildDBSort`.

Yeah, absolutely. After I found this bug I thought the same thing. But after looking at `buildDBSort`, I saw this only works on queries, we don't have here. Therefore, it was too much work for a hotfix. I'd appreciate to use the help of the bones, but for now let's go with this fix to get things working.

phorward:
> But after looking at `buildDBSort`, I saw this only works on queries, we don't have here.

I've made an alternative proposal for this PR in #1169. It's not perfect, but uses the provided facilities.

Furthermore, #1077 was updated and tested now, and I would like to merge it into develop.

sveneberth:
Replaced by https://github.com/viur-framework/viur-core/pull/1169

PR #1167, opened 2024-05-21
cicd: Enable tests for python 3.12

sveneberth:


sveneberth:
> @sveneberth I've updated some more parts of your branch, to enforce Python 3.12 also for codecov and for linting.

okay for me

PR #1169, opened 2024-05-23
fix: Allow dict-filters for `default_order`

phorward:
- Also allows for dict-style filter definitions as default_order
- Alternative fix for #1165


sveneberth (approved):
It's easier than I thought.

PR #1177, opened 2024-05-27
fix: Add `google.cloud.logging_v2.handlers.transports.background_thread`  to the `EXCLUDED_LOGGER_DEFAULTS`

ArneGudermann:
This fix the Problem that the logging explorer is filled with trash logs.
Like 
image


phorward (approved):
Uh perfect, thanks for fixing!

Issue #1178, opened 2024-05-29
`BooleanBone`'s `defaultValue` cannot be a `callable`


sveneberth:
The validation allows only bool, if you provide a `callable` (that returns a boolean) a `TypeError` is raised

https://github.com/viur-framework/viur-core/blob/33b85bea33d529a3caef58bb8c6edfc2b8a8256c/src/viur/core/bones/boolean.py#L26C80-L26C111

phorward:
Fixed by #1274

PR #1179, opened 2024-06-07
fix: Add check for `callable` in `BooleanBone`

ArneGudermann:
Fix for #1178

### DIFF
--- a/src/viur/core/bones/boolean.py
+++ b/src/viur/core/bones/boolean.py
@@ -18,11 +18,13 @@ class BooleanBone(BaseBone):
     def __init__(
         self,
         *,
-        defaultValue: bool | list[bool] | dict[str, list[bool] | bool] = None,
+        defaultValue: bool | list[bool] | dict[str, list[bool] | bool] | t.Callable = None,
         **kwargs
     ):
         if defaultValue is not None:
             # We have given an explicit defaultValue and maybe a complex structure
+            if callable(defaultValue):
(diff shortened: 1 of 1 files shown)


sveneberth (changes requested):
That's not really useful. A callable `defaultValue` should be called at runtime: https://github.com/viur-framework/viur-core/blob/40d4e8228e72110c69d86ed80e680a55efc64656/src/viur/core/bones/base.py#L358

otherwise I could `BooleanBone(defaultValue=my_func())` directly ...

ArneGudermann:
Yes, you are right. The whole thing with the default value needs to be revised... . I don't know if we should wait for #1154. Or is it enough to allow callables for now?

phorward (dismissed):

sveneberth:
> Yes, you are right. The whole thing with the default value needs to be revised... . I don't know if we should wait for #1154. Or is it enough to allow callables for now?

Is okay

```py
my_bone = BooleanBone(...)
my_bone.defaultValue = lambda skel, bone: ...
```
does the job for now too

phorward:
Ok,what's with this PR now? It's still flagged for ViUR-core 3.6. Can we close it?

sveneberth:
> Ok,what's with this PR now? It's still flagged for ViUR-core 3.6. Can we close it?

As it's not really solving the problem and it's possible to workaround with a monkey patch I would recommend to close this too and wait for 3.7.

PR #1180, opened 2024-06-07
feat: Add checksums for files

ArneGudermann:
This PR add the CRC32C and MD5 hash to the File skel. 

Resolves: #1176

### DIFF
--- a/src/viur/core/modules/file.py
+++ b/src/viur/core/modules/file.py
@@ -352,13 +352,13 @@ class FileLeafSkel(TreeSkel):
         readOnly=True,
         visible=False,
     )
-    crc32c_hash = StringBone(
-        descr="CRC32C hash for this File",
-        readOnly=True
+    crc32c_checksum = StringBone(
+        descr="CRC32C checksum",
+        readOnly=True,
     )
-    md5_hash = StringBone(
-        descr="MD5 hash for this File",
-        readOnly=True
+    md5_checksum = StringBone(
+        descr="MD5 checksum",
+        readOnly=True,
     )
 
     def preProcessBlobLocks(self, locks):
@@ -629,8 +629,8 @@ class File(Tree):
         skel["weak"] = True
         skel["width"] = width
         skel["height"] = height
-        skel["crc32c_hash"] = base64.b64decode(blob.crc32c).hex()
-        skel["md5_hash"] = base64.b64decode(blob.md5_hash).hex()
+        skel["crc32c_checksum"] = base64.b64decode(blob.crc32c).hex()
+        skel["md5_checksum"] = base64.b64decode(blob.md5_hash).hex()
 
         return skel.toDB()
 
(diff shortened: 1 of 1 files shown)


sveneberth (changes requested):
Nice, can you add this in [`write`](https://github.com/viur-framework/viur-core/blob/40d4e8228e72110c69d86ed80e680a55efc64656/src/viur/core/modules/file.py#L603-L633) too?

src/viur/core/modules/file.py
```diff
@@ -352,6 +352,14 @@ class FileLeafSkel(TreeSkel):
...
         visible=False,
     )
+    crc32c_hash = StringBone(
+        descr="CRC32C hash for this File",
+        readOnly=True
+    )
+    md5_hash = StringBone(
+        descr="MD5 hash for this File",
+        readOnly=True
+    )
```
phorward, line 362: Can we rename it to checksum to clarify that this feature is used to identify same files?
```suggestion
    crc32c_checksum = StringBone(
        descr="CRC32C checksum",
        readOnly=True,
    )
    md5_checksum = StringBone(
        descr="MD5 checksum",
        readOnly=True,
    )
```

src/viur/core/modules/file.py
```diff
@@ -621,6 +629,8 @@ def write(self, filename: str, content: t.Any, mimetype: str = "text/plain", wid
         skel["weak"] = True
         skel["width"] = width
         skel["height"] = height
+        skel["crc32c_hash"] = base64.b64decode(blob.crc32c).hex()
+        skel["md5_hash"] = base64.b64decode(blob.md5_hash).hex()
```
phorward, line 633:
```suggestion
        skel["crc32c_checksum"] = base64.b64decode(blob.crc32c).hex()
        skel["md5_checksum"] = base64.b64decode(blob.md5_hash).hex()
```

src/viur/core/modules/file.py
```diff
@@ -907,6 +917,8 @@ def add(self, skelType: SkelType, node: db.Key | int | str | None = None, *args,
             skel["size"] = blob.size
             skel["parentrepo"] = rootNode["key"] if rootNode else None
             skel["weak"] = rootNode is None
+            skel["crc32c_hash"] = base64.b64decode(blob.crc32c).hex()
+            skel["md5_hash"] = base64.b64decode(blob.md5_hash).hex()
```
phorward, line 921:
```suggestion
            skel["crc32c_checksum"] = base64.b64decode(blob.crc32c).hex()
            skel["md5_checksum"] = base64.b64decode(blob.md5_hash).hex()
```

phorward (changes requested):
Hello @ArneGudermann, thanks for achieving this issue.

First of all, I would like to refer the hashed value as "checksum" as it should be an indicator for same files, regardless of the filename and the size. 

Then, I'm wondering about the chosen hash algorithms. Why did you chose CRC32c and MD5?

After a little research, I think that SHA-256 would be the best choice for collision resistance. MD5, SHA-1 and CRC32c are not designed for collision resistance. Therefore, I would prefer to only provide a sha256 checksum for all files, and avoid use of two hash algorithms (what's the advantage?). 

I'm open for discussions.

ArneGudermann:
@phorward i used [md5](https://cloud.google.com/python/docs/reference/storage/latest/google.cloud.storage.blob.Blob#google_cloud_storage_blob_Blob_md5_hash) and  [crc32c](https://cloud.google.com/python/docs/reference/storage/latest/google.cloud.storage.blob.Blob#google_cloud_storage_blob_Blob_crc32c) because they are the only two that are provided by google storage. otherwise we would have to download the whole blob and hash it ourselves.

phorward (approved):
IMHO, we can use this for now.

sveneberth (approved):

Issue #803, opened 2023-07-31
SEO-Features not working properly with umlauts


phorward:
There are several problems with SEO-URLs managed by VIUR3.
It doesn't seem to work with umlauts and spaces properly.

E.g. it's unclear what the correct URL for "Speicherbräu Bier"  will be. It is saved as something like "speicherbräu bier", but the browser will request "speicherbr%C3%A4%40bier" which doesn't match the database entry properly. It first seemed to be a case-sensitivity problem as fixed by #801 but this is only a part of the problem.

There is further investigation required on this issue.

phorward:
I would like to resolve this together with PR#800, therefore the bug will be fixed with viur-core==3.5

PR #1181, opened 2024-06-07
feat: Add Redirect map

### LINKED ISSUE #1168: a Simple Redirect Handler for mapping urls
A simple configureable dict which maps from one url to another. So that we can e.g. put the admin on different URLs.
```
{"/vi-admin": "/admin/index.html",
"/vi": "/vi/main.html"}
```

Since the url "vi-admin" is no longer in the vi renderer, you could possibly remove it from the renderer.

ArneGudermann:
Resolves #1168 

This PR adds an Redirect handler with two modes. 
In the normal mode its only matches the key in the dict.

In the advanced mode you can provides key with wildcards like:

```python
conf.redirect_map["/test/add*"]="/test/edit/"
```


sveneberth (changes requested):

src/viur/core/config.py
```diff
@@ -825,6 +825,12 @@ class Conf(ConfigType):
     }
     """Describing the Script module"""
 
+    redirect_map = {}
```
sveneberth, line 828:
```suggestion
    redirect_map: dict[str, str] = {}
```

src/viur/core/config.py
```diff
@@ -825,6 +825,12 @@ class Conf(ConfigType):
...
+    """Map for redirect Urls"""
+
+    redirect_map_advanced_mode = False
```
sveneberth, line 831:
```suggestion
    redirect_map_advanced_mode: bool = False
```

sveneberth (dismissed):

sveneberth (approved):

phorward (changes requested):
Hi @ArneGudermann, I would like to combine this pull request with #1218 (#803) to get rid of the SEO-stuff. Your redirect-map is part of it, but should come from a SkelModule.

If this PR is not necessary for current customer projects, I would like to shift it to v3.8.

PR #1183, opened 2024-06-07
feat: Add charset  `v_func` for `StringBone` 

ArneGudermann:
Replacement for #1141

### DIFF
--- a/src/viur/core/bones/string.py
+++ b/src/viur/core/bones/string.py
@@ -383,9 +383,9 @@ class StringBone(BaseBone):
     @classmethod
     def v_func_valid_chars(cls, valid_chars: t.Iterable = string.printable) -> t.Callable:
         """
-        Returns a function that takes a string and check whether it contains valid characters.
-        If all characters are valid, it returns None.
-        If not all characters are valid, it returns an error Message.
+        Returns a function that takes a string and checks whether it contains valid characters.
+        If all characters of the string are valid, it returns None, and succeeds.
+        If invalid characters are present, it returns an appropriate error message.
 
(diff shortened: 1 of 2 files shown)


sveneberth (approved):

phorward (changes requested):
Hello @ArneGudermann, this looks good, can you please provide a docstring with an example as well?

src/viur/core/bones/string.py
```diff
@@ -381,7 +381,16 @@ def structure(self) -> dict:
         return ret
 
     @classmethod
-    def v_func_valid_chars(cls, valid_chars=string.printable) -> t.Callable:
+    def v_func_valid_chars(cls, valid_chars: t.Iterable = string.printable) -> t.Callable:
+        """
+        Returns a function that takes a string and check whether it contains valid characters.
+        If all characters are valid, it returns None.
+        If not all characters are valid, it returns an error Message.
```
phorward, line 388:
```suggestion
        Returns a function that takes a string and checks whether it contains valid characters.
        If all characters of the string are valid, it returns None, and succeeds.
        If invalid characters are present, it returns an appropriate error message.
```

phorward (changes requested):
@ArneGudermann can you provide an example definition for a `StringBone` that uses the `v_func_valid_chars`-feature? Then, developers can directly abstract the code for their specific use-case and the feature will be used with ease.

sveneberth (approved):

phorward (changes requested):
Still missing an example in the docstring!

ArneGudermann:
@phorward now ok ?

PR #1184, opened 2024-06-07
fix: Keep HTML-entities in `HtmlSerializer`

### LINKED ISSUE #1182: Core cleans non breaking spaces from Text Bone Input
It looks like we should be respecting these entity refs, but the method that seems responsible for handling them has a FIXME with no further comment. Is this an easy fix? It clobbers user input in a project where I need this to work.

https://github.com/viur-framework/viur-core/blob/33b85bea33d529a3caef58bb8c6edfc2b8a8256c/src/viur/core/bones/text.py#L125

sveneberth:
With the default option `convert_charrefs=True` the serializer replace HTML entities to Unicode (like `&nbsp;` to `\u00a0`). This PR disables this option.

Furthermore it adds some tests (which are adapted from `StringBone`).

fixes #1182

### DIFF
--- a/src/viur/core/bones/text.py
+++ b/src/viur/core/bones/text.py
@@ -91,7 +91,6 @@ class HtmlSerializer(HTMLParser):  # html.parser.HTMLParser
          "\n": "",
          "\0": ""})
 
-
     def __init__(self, validHtml=None, srcSet=None, convert_charrefs: bool = True):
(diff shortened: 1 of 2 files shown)


phorward (approved):
I'm well with this.

This also refers to #631

phorward (approved):

PR #1186, opened 2024-06-10
feat: Add `sorted` feature to `MultipleConstraints`

### LINKED ISSUE #1185: Extend `MultipleConstraints` to accept for a sort method
It sometimes might be useful to automatically sort multiple bones alphabetically, or by some kind of algorithm.

Therefore, this could be possible to define:
```py
sorted_names = StringBone(
    descr="Names",
    multiple=MultipleConstraints(sorted=True),
)
```
but also
```py
sorted_names = StringBone(
    descr="Names",
    multiple=MultipleConstraints(sorted=lambda k: k[-1]),
)
```

ArneGudermann:
Resolves #1185

### DIFF
--- a/src/viur/core/bones/base.py
+++ b/src/viur/core/bones/base.py
@@ -138,6 +138,8 @@ class MultipleConstraints:  # Used to define constraints on multiple bones
     """An integer representing the upper bound of how many entries can be submitted (default: 0)."""
     preventDuplicates: bool = False  # Prevent the same value of being used twice
     """A boolean value indicating if the same value can be used twice (default: False)."""
+    sorted: bool | t.Callable = False
+    """A boolean value or a method indicating if the value must be sorted (default: False)."""
 
 
 class ComputeMethod(Enum):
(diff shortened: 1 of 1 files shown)


phorward (changes requested):
Thanks for this first quick draft on #1185.

Can you please provide a way to allow for reversed sorting as well? I think this should be another member of `MultipleConstraints`. See [here](https://docs.python.org/3/library/functions.html#sorted) for details.

src/viur/core/bones/base.py
```diff
@@ -140,6 +140,11 @@ class MultipleConstraints:  # Used to define constraints on multiple bones
     """A boolean value indicating if the same value can be used twice (default: False)."""
     sorted: bool | t.Callable = False
     """A boolean value or a method indicating if the value must be sorted (default: False)."""
+    sorted_reversed: bool = False
```
phorward, line 143:
```suggestion
    reverse: bool = False
```

phorward (approved):

sveneberth (approved):

PR #1189, opened 2024-06-13
fix: render structure on first list (improvement of #774)

phorward:
This improves #774, and renders the structure on calls without cursor, so legacy systems can work better with it and still use the feature.

### DIFF
--- a/src/viur/core/config.py
+++ b/src/viur/core/config.py
@@ -646,7 +646,7 @@ class Conf(ConfigType):
     compatibility: Multiple[str] = [
         "json.bone.structure.camelcasenames",  # use camelCase attribute names (see #637 for details)
         "json.bone.structure.keytuples",  # use classic structure notation: `"structure = [["key", {...}] ...]` (#649)
-        "json.bone.structure.inlists",  # dump skeleton structure with every JSON list response (#774 for details)
+        "json.bone.structure.inlists",  # dump skeleton structure with every JSON list response (details: #774 & #1189)
     ]
     """Backward compatibility flags; Remove to enforce new layout."""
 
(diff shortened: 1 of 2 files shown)


sveneberth (changes requested):
huh? That's exactly the point of the compatibility mode (yes, I've seen that it's only about the first batch).
But now you are reactivating exactly what we have abolished: the structure is rendered with every first list request. Regardless of whether I make several search requests one after the other or the structure doesn't interest me at all.
I am against this change. 

<sub>(I do not request a change, I want to leave it as it is)</sub>

---

**Edit:** We could make this controllable via an input parameter. e.g. via context variable `@include_structure=True` or via a header `X-Include-Structure: True`.

phorward:
> **Edit:** We could make this controllable via an input parameter. e.g. via context variable `@include_structure=True` or via a header `X-Include-Structure: True`.

What do you think about a general feature to provide X-ViUR-Structure-header to obtain the structure in any call? This also relates to view/edit/add/clone and other actions.

sveneberth:
> > **Edit:** We could make this controllable via an input parameter. e.g. via context variable `@include_structure=True` or via a header `X-Include-Structure: True`.
> 
> What do you think about a general feature to provide X-ViUR-Structure-header to obtain the structure in any call? This also relates to view/edit/add/clone and other actions.

Would be nice.

But to make this working on vue.js development server (sending headers in a browser to another server than the current host), we need to support preflight CORS requests. I'd already some issues in other use-cases, so I've started #1215.

phorward:
We should switch to separation of structure and data fetching, but we should modify cache header so that structure requests are being cached for a specific amount of time. We can think about a decorator for setting a cache-time.

phorward:
This relates to #1273.

phorward:
Closing this for now due to irrelevancy.

Issue #1071, opened 2024-02-15
Consistent json/xml/msgpack rendering protocol


phorward:
This issue is a follow-up on #1070.

> We should provide a consistent result for all functions.

Yes. A well defined, generic and sophisticated protocol should be created to fulfill any needs on a modern REST-like system.

In the current protocol, only the upper 7 states are designed well. The rest is stupid junk which needs to be redefined and refactored.

This issue should define a new, standardized concept and 

## Current protocol (JSON)

- Well defined
  - view
  - preview
  - list
  - add
  - addSuccess
  - edit
  - editSuccess
- Bad defined
  - deleteSuccess => "OKAY"
    - delete ... okay this is not defined, lol (would be with Action Skels)
  - listRootNodes => ??? sense??? Yes, it is used and has a sense, but why?
  - login => renders edit-skel??
  - loginChoices => list of methods
  - loginSucceeded => "OKAY"
  - logoutSuccess => "OKAY"
  - verifySuccess => "OKAY"
  - verifyFailed => "FAILED"
  - passwdRecoverInfo => Returns either an edit-skel or a msg?
  - passwdRecover => returns edit
  - second_factor_add => Returns `{otp_uri: "xyz"}`
  - second_factor_add_success => "OKAY"
  - second_factor_choice => list of second factors?

PR #1190, opened 2024-06-14
feat: Implement abstract renderer

### LINKED ISSUE #1163: Add abstract / base class for all renderers
I can currently not check if the provided argument is any renderer, because their only common base is `object`.

sveneberth:
Resolves #1163


phorward (approved):
IMHO we can drop the XML renderer, as it can be re-implemented easily with planned changes coming soon regarding the json-renderer. (see #1071, part of ActionSkel-paradigm)

ArneGudermann (approved):

Issue #1191, opened 2024-06-14
Fix xml renderer


sveneberth:
The xml renderer was always left behind in the past and sometimes has errors (such as [this undefined symbol `render`](https://github.com/viur-framework/viur-core/blob/main/src/viur/core/render/xml/user.py#L12C22-L12C28))

Maybe in the future the html renderer can also inherit from the xml renderer and some parts can be moved.

phorward:
> Maybe in the future the html renderer can also inherit from the xml renderer and some parts can be moved.

I see this as part of #1071 and would generally advise to drop the renderer entirely for now, and re-implement it later on demand using a master renderer constructing JSON-serializable objects which simply is dumped to the specific target format, like JSON/MsgPack or even XML.

PR #1192, opened 2024-06-17
refactor: Remove `xml` renderer

phorward:
This relates to #1071 and #1191.
XML is not gone, just the old implementation should be removed.
We can discuss about it.


phorward:
In the viur-meeting on 2024-06-21, we decided to remove the current XML renderer implementation now, and revive it later in a new attempt related to #1071.

PR #1193, opened 2024-06-21
feat: Add `read` method in `RefSkel`

### LINKED ISSUE #1101: Provide a `RefSkel.read()` shortcut to read entire target skeleton
For several use-cases, this
```py
studyworkflow_skel = conf.main_app.studyworkflow.viewSkel()

if not studyworkflow_skel.fromDB(skel["studyworkflow"]["dest"]["key"]):
    logging.error(f"""The studyworkflow {skel["studyworkflow"]["dest"]=} was meanwhile deleted?""")
    return
```
could be substituted by 
```py
studyworkflow_skel = skel["studyworkflow"]["dest"].read()
```
with nice exceptions being thrown when the skeleton is gone!

ArneGudermann:
Resolves #1101

### DIFF
--- a/src/viur/core/skeleton.py
+++ b/src/viur/core/skeleton.py
@@ -248,7 +248,7 @@ class SkeletonInstance:
         # Load a @classmethod from the Skeleton class and bound this SkeletonInstance
         elif item in {"fromDB", "toDB", "all", "unserialize", "serialize", "fromClient", "getCurrentSEOKeys",
                       "preProcessSerializedData", "preProcessBlobLocks", "postSavedHandler", "setBoneValue",
-                      "delete", "postDeletedHandler", "refresh"}:
+                      "delete", "postDeletedHandler", "refresh", "read"}:
             return partial(getattr(self.skeletonCls, item), self)
         # Load a @property from the Skeleton class
         try:
@@ -343,15 +343,6 @@ class SkeletonInstance:
         memodict[id(self)] = res
         return res
 
-    def read(self):
-        """
-        Read the full skel form the Database.
-        Can be used in the RelationalBone for reading the full Skeleton from a RefSkel
-        :raise AssertionError:  If the entry is no longer in the database.
-        """
-        skel = skeletonByKind(self["key"].kind)()
-        assert skel.fromDB(self["key"])
-        return skel
 
 
 class BaseSkeleton(object, metaclass=MetaBaseSkel):
@@ -1377,6 +1368,16 @@ class RefSkel(RelSkel):
             bone_map |= {k: fromSkel.__boneMap__[k] for k in fnmatch.filter(fromSkel.__boneMap__.keys(), arg)}
         newClass.__boneMap__ = bone_map
         return newClass
+
+    def read(self):
+        """
+        Read the full skel form the Database.
+        Can be used in the RelationalBone for reading the full Skeleton from a RefSkel
+        :raise AssertionError:  If the entry is no longer in the database.
+        """
+        skel = skeletonByKind(self["key"].kind)()
+        assert skel.fromDB(self["key"])
+        return skel
 
 
 class SkelList(list):
--- a/src/viur/core/skeleton.py
+++ b/src/viur/core/skeleton.py
@@ -343,6 +343,16 @@ class SkeletonInstance:
         memodict[id(self)] = res
         return res
 
+    def read(self):
+        """
+        Read the full skel form the Database.
+        Can be used in the RelationalBone for reading the full Skeleton from a RefSkel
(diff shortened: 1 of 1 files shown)


src/viur/core/skeleton.py
```diff
@@ -1368,6 +1368,16 @@ def fromSkel(cls, kindName: str, *args: list[str]) -> t.Type[RefSkel]:
         newClass.__boneMap__ = bone_map
         return newClass
 
+    def read(self):
+        """
+        Read the full skel form the Database.
```
phorward, line 1373:
```suggestion
        Read full skeleton instance of the RefSkel from the database.
```

src/viur/core/skeleton.py
```diff
@@ -1368,6 +1368,16 @@ def fromSkel(cls, kindName: str, *args: list[str]) -> t.Type[RefSkel]:
         newClass.__boneMap__ = bone_map
         return newClass
 
+    def read(self):
```
phorward, line 1371:
```suggestion
    def read(self) -> SkeletonInstance:
```

src/viur/core/skeleton.py
```diff
@@ -1368,6 +1368,16 @@ def fromSkel(cls, kindName: str, *args: list[str]) -> t.Type[RefSkel]:
...
+        """
+        Read the full skel form the Database.
+        Can be used in the RelationalBone for reading the full Skeleton from a RefSkel
+        :raise AssertionError:  If the entry is no longer in the database.
+        """
+        skel = skeletonByKind(self["key"].kind)()
+        assert skel.fromDB(self["key"])
```
phorward, line 1378:
```suggestion
        if not skel.fromDB(self["key"]):
            raise ValueError(f"""The key {self["key"]!r} seems to be gone""") 
```

phorward (changes requested):
Hello @ArneGudermann, Thanks for facing this issue.
Can you first fix the provided suggestions?

Furthermore, I think it might be good to add a `read`-method generally to `Skeleton`, so that it can be used as a replacement for `fromDB()` - Why? Because issue #630 already proposes a  `Skeleton.read()` method.

It should have the signature `def read(self, key=None)`, and in the case that no key is provided, `skel["key"]` will be used (as it's the case here). If you want, I can make a follow-up pull request when this is merged.

src/viur/core/skeleton.py
```diff
@@ -1368,6 +1368,17 @@ def fromSkel(cls, kindName: str, *args: list[str]) -> t.Type[RefSkel]:
         newClass.__boneMap__ = bone_map
         return newClass
 
+    def read(self) -> SkeletonInstance:
+        """
+        Read full skeleton instance of the RefSkel from the database.
+        Can be used in the RelationalBone for reading the full Skeleton from a RefSkel
+        :raise AssertionError:  If the entry is no longer in the database.
```
sveneberth, line 1375:
```suggestion
        :raise ValueError: If the entry is no longer in the database.
```

sveneberth (changes requested):

sveneberth (approved):

src/viur/core/skeleton.py
```diff
@@ -1368,6 +1368,17 @@ def fromSkel(cls, kindName: str, *args: list[str]) -> t.Type[RefSkel]:
...
+        """
+        Read full skeleton instance of the RefSkel from the database.
+        Can be used in the RelationalBone for reading the full Skeleton from a RefSkel
+        :raise ValueError: If the entry is no longer in the database.
+        """
+        skel = skeletonByKind(self["key"].kind)()
+        if not skel.fromDB(self["key"]):
+            raise ValueError(f"""The key {self["key"]!r} seems to be gone""")
+        return skel
+
```
phorward, line 1381: Hello, because there was no change [regarding to my proposal](https://github.com/viur-framework/viur-core/pull/1193#pullrequestreview-2135517669), I would like to suggest this alternative version of `RefSkel.read()`. It also doesn't use the kind name from the key, but the `kindName` from the class, which is safer.

```suggestion
    def read(self, key: t.Optional[db.Key | str | int] = None) -> SkeletonInstance:
        """
        Read full skeleton instance referenced by the RefSkel from the database.

        Can be used for reading the full Skeleton from a RefSkel.
        The `key` parameter also allows to read another, given key from the related kind.

        :raise ValueError: If the entry is no longer in the database.
        """
        skel = skeletonByKind(self.kindName)()

        if not skel.fromDB(key or self["key"]):
            raise ValueError(f"""The key {self["key"]!r} seems to be gone""")

        return skel

```

phorward (changes requested):

phorward (approved):

sveneberth (approved):

PR #1194, opened 2024-06-21
feat: Add `check_for_changes` and `serialize`

ArneGudermann:
This PR introduces the methods `serialize` and `check_for_changes`.
The `serialize`  method is responsible for serializing all bones in a skeleton. Method `check_for_changes` can be used to check whether data that is written to the datastore has changed. This is used to avoid unnecessary write operations

### DIFF
--- a/src/viur/core/prototypes/list.py
+++ b/src/viur/core/prototypes/list.py
@@ -1,3 +1,4 @@
+import copy
 import logging
 import typing as t
 from viur.core import current, db, errors, utils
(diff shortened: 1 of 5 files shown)


phorward (changes requested):
Hello @ArneGudermann, thanks for making this proposal.

I think that we should not integrate this proposal, as it is sometimes necessary to make an edit without changing any data, for example to update the `changedate` or rebuild a search index for a specific entry.

We can discuss it in the viur-meeting, maybe we can find an alternative use case or integrate it later in a special "lazy" edit-action.

PR #1198, opened 2024-06-27
refactor: `DatabaseAdapter` with simplified triggers

phorward:
Hello!

This pull request was made to provide the requirement for an history logging module needed by some ViUR projects.

It does the following:
- Renamed and simplified the signatures of the hook-functions
- Allow `Skeleton.database_adapters` to be set to an iterable, to allow running multiple adapter hooks
- Refactor some parts of Skeleton generally for improved maintenance


src/viur/core/skeleton.py
```diff
@@ -1220,8 +1264,11 @@ def fixDotNames(entity):
                 updateRelations(key, time() + 1, None)
 
         # Inform the custom DB Adapter of the changes made to the entry
-        if skel.customDatabaseAdapter:
-            skel.customDatabaseAdapter.updateEntry(db_obj, skel, change_list, is_add)
+        for adapter in utils.ensure_iterable(skel.customDatabaseAdapter):
+            db_obj = adapter.updateEntry(db_obj, skel, change_list, is_add)  # FIXME: DEPRECATED.
```
sveneberth, line 1268: It's just a hook.


https://github.com/viur-framework/viur-core/blob/41820476eb03252a604bc741ca3a815731940456/src/viur/core/skeleton.py#L651

```suggestion
            adapter.updateEntry(db_obj, skel, change_list, is_add)  # FIXME: DEPRECATED.
```

src/viur/core/skeleton.py
```diff
@@ -1132,8 +1173,11 @@ def txnUpdate(write_skel):
             db_obj = skel.preProcessSerializedData(db_obj)
 
             # Allow the custom DB Adapter to apply last minute changes to the object
-            if skel.customDatabaseAdapter:
-                db_obj = skel.customDatabaseAdapter.preprocessEntry(db_obj, skel, change_list, is_add)
+            for adapter in utils.ensure_iterable(skel.customDatabaseAdapter):
+                db_obj = adapter.preprocessEntry(db_obj, skel, change_list, is_add)  # FIXME: DEPRECATED.
```
sveneberth, line 1177: Why should the `#FIXME: DEPRECATED` mean?

src/viur/core/skeleton.py
```diff
@@ -1132,8 +1173,11 @@ def txnUpdate(write_skel):
             db_obj = skel.preProcessSerializedData(db_obj)
 
             # Allow the custom DB Adapter to apply last minute changes to the object
-            if skel.customDatabaseAdapter:
-                db_obj = skel.customDatabaseAdapter.preprocessEntry(db_obj, skel, change_list, is_add)
+            for adapter in utils.ensure_iterable(skel.customDatabaseAdapter):
```
sveneberth, line 1176: Wouldn't it be more efficient to create / normalize an iterable once in the meta class or similar instead of doing this every time at runtime when saving?!

sveneberth (changes requested):

src/viur/core/skeleton.py
phorward: Because all these events with ugly names like `preprocessEntry` or `updateEntry` are being superseded by the new `trigger` handler.

src/viur/core/utils/__init__.py
```diff
@@ -101,6 +102,25 @@ def normalizeKey(key: t.Union[None, 'db.KeyClass']) -> t.Union[None, 'db.KeyClas
...
 
 
+def ensure_iterable(obj: t.Any, test: t.Optional[t.Callable[[t.Any], bool]] = None) -> t.Iterable[t.Any]:
+    """
+    Ensures an object to be iterable.
+    An additional test can be provided to check additionally.
+
+    If the object is not considered to be iterable, a tuple with the object is returned.
+    """
+    if isinstance(obj, Iterable):
```
phorward, line 112:
```suggestion
    if isinstance(obj, t.Iterable):
```

src/viur/core/utils/__init__.py
```diff
@@ -101,6 +102,25 @@ def normalizeKey(key: t.Union[None, 'db.KeyClass']) -> t.Union[None, 'db.KeyClas
...
 
 
+def ensure_iterable(obj: t.Any, test: t.Optional[t.Callable[[t.Any], bool]] = None) -> t.Iterable[t.Any]:
+    """
+    Ensures an object to be iterable.
+    An additional test can be provided to check additionally.
+
+    If the object is not considered to be iterable, a tuple with the object is returned.
+    """
+    if isinstance(obj, t.Iterable):
```
phorward, line 112:
```suggestion
    if isinstance(obj, Iterable):  # uses collections.abc.Iterable
```

src/viur/core/skeleton.py
phorward: @sveneberth I've achieved your suggestion in commit 0071940f7fe0435c3df2091994f99e946805cc48

src/viur/core/skeleton.py
sveneberth: But i find it clearer to directly subclass suitable methods for corresponding events than to build cases in a method.

So instead of 
```python
class MyAdapter(CustomDatabaseAdapter):
    def updateEntry(self, entity: db.Entity, skel: SkeletonInstance,
                    changeList: list[str], isAdd: bool) -> None:
        # my update logic

    def deleteEntry(self, entity: db.Entity, skel: SkeletonInstance) -> None:
        # my delete logic
``` 

I should do in future

```python
class MyAdapter(CustomDatabaseAdapter):
    def trigger(
        self,
        action: str,
        old_skel: db.Entity | SkeletonInstance = None,
        new_skel: SkeletonInstance = None,
        change_list: t.Iterable[str] = (),
        descr: t.Optional[str] = None,
        user: t.Optional[SkeletonInstance] = None,
        tags: t.Iterable[str] = (),
        db_write: bool = False
    ) -> str | None:
        match action:
            case "edit":
                # my update logic
        
            case "delete":
                # my delete logic
``` 

?

phorward: Well, yes this is how its currently intended to be used.

The problem is similar to the ActionSkel event handling we discussed, and where I also don't have the perfect solution right now.

I would like to abandon ugly functions with ugly signatures defined for every operation, when it can be done  with just one function and conditional code on demand.

src/viur/core/skeleton.py
```diff
@@ -1132,8 +1182,12 @@ def txnUpdate(write_skel):
             db_obj = skel.preProcessSerializedData(db_obj)
 
             # Allow the custom DB Adapter to apply last minute changes to the object
-            if skel.customDatabaseAdapter:
-                db_obj = skel.customDatabaseAdapter.preprocessEntry(db_obj, skel, change_list, is_add)
+            for adapter in skel.customDatabaseAdapter:
+                # FIXME: DEPRECATED: Superseeded by adapter.trigger.
```
sveneberth, line 1186: We decided in viur-meeting to keep both.
```suggestion
```

src/viur/core/skeleton.py
```diff
@@ -1220,8 +1274,11 @@ def fixDotNames(entity):
                 updateRelations(key, time() + 1, None)
 
         # Inform the custom DB Adapter of the changes made to the entry
-        if skel.customDatabaseAdapter:
-            skel.customDatabaseAdapter.updateEntry(db_obj, skel, change_list, is_add)
+        for adapter in skel.customDatabaseAdapter:
+            adapter.updateEntry(db_obj, skel, change_list, is_add)  # FIXME: DEPRECATED: Superseeded by adapter.trigger.
```
sveneberth, line 1278: We decided in viur-meeting to keep both.
```suggestion
            adapter.updateEntry(db_obj, skel, change_list, is_add)
```

src/viur/core/skeleton.py
```diff
@@ -1330,9 +1387,11 @@ def txnDelete(skel: SkeletonInstance):
         for boneName, _bone in skel.items():
             _bone.postDeletedHandler(skel, boneName, key)
         skel.postDeletedHandler(key)
+
         # Inform the custom DB Adapter
-        if skel.customDatabaseAdapter:
-            skel.customDatabaseAdapter.deleteEntry(dbObj, skel)
+        for adapter in skel.customDatabaseAdapter:
+            adapter.deleteEntry(dbObj, skel)  # FIXME: DEPRECATED: Superseeded by adapter.trigger.
```
sveneberth, line 1393: We decided in viur-meeting to keep both.
```suggestion
            adapter.deleteEntry(dbObj, skel)
```

src/viur/core/skeleton.py
```diff
@@ -1442,6 +1501,35 @@ def __init__(self, baseSkel=None):
...
 
 
+# Module functions
+
+
+def skeletonByKind(kindName: str) -> t.Type[Skeleton]:
+    """
+        Returns the Skeleton-Class for the given kindName. That skeleton must exist, otherwise an exception is raised.
+        :param kindName: The kindname to retreive the skeleton for
+        :return: The skeleton-class for that kind
```
sveneberth, line 1511: Please unindent these docstrings according to our [Coding Convention](https://github.com/viur-framework/viur-core/blob/main/CONTRIBUTING.md#coding-convention), where we follow [PEP 257](https://peps.python.org/pep-0257). :)

```suggestion
    Returns the Skeleton-Class for the given kindName. That skeleton must exist, otherwise an exception is raised.
    :param kindName: The kindname to retreive the skeleton for
    :return: The skeleton-class for that kind
```

sveneberth (changes requested):

phorward:
Hello @sveneberth, I've created an alternative approach within this pull request, see [8f7475c](https://github.com/viur-framework/viur-core/pull/1198/commits/8f7475c68ed6a11bdd4396164aaf8afe248e1ea4) for the significant change.

This is not backward-compatible, but there are only few usages of this feature.

ArneGudermann (dismissed):

src/viur/core/skeleton.py
```diff
@@ -1202,7 +1241,7 @@ def fixDotNames(entity):
 
         # END of __txn_update subfunction
 
-        # Run our SaveTxn
+        # Run transactional function
```
sveneberth, line 1244: IMO comments like this are unnecessary if you write self-explaining code and can be removed completely.
A more helpful comment would be: _"Run the function in a transaction or inline if we're already in a transaction"_. But even that is self-explaining and uncessary, imo.




This is the same like:
```py
# Set variable foo to 3
foo = 3
```

sveneberth (changes requested):

src/viur/core/skeleton.py
```diff
@@ -609,56 +624,68 @@ def __init__(cls, name, bases, dct):
...
 
 
-class CustomDatabaseAdapter:
-    # Set to True if we can run a fulltext search using this database
+class DatabaseAdapter:
+    """
+    Adapter class used to bind or use other databases and hook operations when working with a Skeleton.
+    """
+
     providesFulltextSearch: bool = False
```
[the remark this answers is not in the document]
phorward: I can't, because this member is used by viur-datastore, as a reverse dependency!

src/viur/core/skeleton.py
```diff
@@ -1141,9 +1180,9 @@ def __txn_update(write_skel):
 
             db_obj = skel.preProcessSerializedData(db_obj)
 
-            # Allow the custom DB Adapter to apply last minute changes to the object
-            if skel.customDatabaseAdapter:
-                db_obj = skel.customDatabaseAdapter.preprocessEntry(db_obj, skel, change_list, is_add)
+            # Allow the database adapter to apply last minute changes to the object
+            for adapter in skel.database_adapters:
+                adapter.prewrite(skel, is_add, change_list)
```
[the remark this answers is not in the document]
phorward: **First;** The object can be changed in all situations.
**Second;** No, the *db_obj* is contained in *skel* and can be accessed by *skel.dbEntity*. It can freely be changed.
**Third;** No, its bad naming convention and should be eliminated. I want to have some kind of hooks named pre-<action>, <action>, post-<action>. This is a concept. And a pre-hook generally is a hook and nothing which returns any object, which *might* be changed. It still can be. Therefore, I want to keep the concept now as it is, especially because it is more concept with conciseness than the messed-up ugly stuff below, left untended and unloved for a decade.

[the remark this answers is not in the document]
phorward: Additionally, to underline the reasons for this naming convention, is that #630 already defines a similar concept, although much is missing. The current situation is a mess, and not a standard.

src/viur/core/skeleton.py
phorward: the comment was there before and related to a function which doesn't exist since 12 years or so. Your comment on the comment is just as useless.

sveneberth: Yes, so just remove it instead of `verschlimmbessern` it.
```suggestion
```

src/viur/core/skeleton.py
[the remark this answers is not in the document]
sveneberth: > The object can be changed in all situations

Yes, but it should obviously be recognisable that this method changes an object. And you make this recognisable via a return value and not just via side effects.



> No, the _db_obj_ is contained in _skel_ and can be accessed by _skel.dbEntity_. It can freely be changed

If my custom `preProcessSerializedData` creates a new instance, you would completely ignore this object in this method, but this new object would be written later with `db.Put`.

```
def preProcessSerializedData(self, db_obj):
    new_obj = db.Entity()
    # [...] Some edge case copy logic or whatever
    return new_obj
```

Therefore, you need to continue with the `db_obj` and not with `skel.dbEntity`, they must no longer be the identical one.





> No, its bad naming convention and should be eliminated. I want to have some kind of hooks named _pre-action_, _action_, and _post-action_, although _action_ is the post-action. This is a concept

Well, it's just a name. Call it what you like. I still think it fits better under a known name, considering the existing code. And then later in ViUR4 it's better to rename everything consistently than to call it _preprocess_ in one class and _prewrite_ in another.




> And a pre-hook generally is a hook and nothing which returns any object, which _might_ be changed.

I don't know what makes you think that a hook doesn't return anything, but a hook can very well change objects, return them and thus change the behaviour of software. That's exactly the point of hooks, to create the ability to intervene in code. There are several examples on [Wikipedia](https://en.wikipedia.org/wiki/Hooking) and in the linked literature.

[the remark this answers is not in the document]
phorward: @sveneberth I've now updated the pull request again.

> Therefore, you need to continue with the `db_obj` and not with `skel.dbEntity`, they must no longer be the identical one.

Well, there should be no further distinction between the Skeleton instance and the database object that is written. This is a datastore-specific thing, and should not be part of any hooks, the hook should work with the skeleton instance object and its members, and not with the skeleton instance object and a copy of the content as db.Entity.

It is nonsense to return and store and entirely different database object than the `skel.dbEntity`-one. To get rid of the `db_obj` variable handled inside the transactional function, I've now replaced it so that `skel.dbEntity` is used throughout the entire process.

You can still replace it by everything you want, on your demand and on your responsibility. The name `prewrite` is more clerly than "preprocess"... preprocesses what? C source code, like a C-pre-processor #ifdef #ifndef, or what? (scnr).

This is my last comment on this issue now. I need the feature for an important and critical project and also a cleaner code base, which is not the result of several years of omitted maintenance.

akelch (approved):

ArneGudermann (approved):

PR #1204, opened 2024-07-01
fix: Clean-up `KeyBone` and added unserialization

phorward:
Unable to test because of viur-framework/vi-vue-utils#13


src/viur/core/bones/key.py
```diff
@@ -76,81 +75,38 @@ def singleValueFromClient(self, value, skel, bone_name, client_data):
...
-            and isinstance(skel.dbEntity, db.Entity)
-            and skel.dbEntity.key
-                and not skel.dbEntity.key.is_partial):
+    def singleValueUnserialize(self, val):
+        if isinstance(val, db.Key):
+            rval = db.normalizeKey(val)
+        else:
+            rval, err = self.singleValueFromClient(val)
+            if err:
+                print(val)
```
sveneberth, line 84: I think this was only there for debugging?
```suggestion
```

src/viur/core/bones/key.py
```diff
@@ -76,81 +75,38 @@ def singleValueFromClient(self, value, skel, bone_name, client_data):
...
-
-        if (name == "key"
-            and isinstance(skel.dbEntity, db.Entity)
-            and skel.dbEntity.key
-                and not skel.dbEntity.key.is_partial):
+    def singleValueUnserialize(self, val):
+        if isinstance(val, db.Key):
+            rval = db.normalizeKey(val)
+        else:
+            rval, err = self.singleValueFromClient(val)
```
sveneberth, line 82: I don't think it's a good idea to use `singleValueFromClient` here. There are the `allowed_kind`, existance `check` and `isInvalid` (`vFunc`) checks. I on of these check logic has changed or something that causes the saved key to be okay before, but not anymore the unserialize will now crash.

sveneberth (changes requested):

src/viur/core/bones/key.py
phorward: This guarantees that bones that previously where e.g. StringBone or RawBone are being correctly converted to KeyBones. Can you give an example for your considerations?

sveneberth: Image you have a `TreeSkel`, where keybone was initially simply 

```
parententry = KeyBone(
    descr="Parent",
)
```

but than you decided nesting of entries should be limited. So you implement an `isInvalid` method (not tested):

```
def is_depth_okay(value: db.Key) -> str | None:
    """parent is okay, if it's depth is max 3"""
    depth = 1
    while True:
        skel = self.viewSkel("node")
        skel.fromDB(value)
        if skel["parententry"] == skel["parentrepo"]:  # We reached the top level
            break
        depth += 1
        value = skel["parententry"]
    if depth > 3:
        return f"depth of {depth} is too much"
    return None
```


and use it for your bone:
```
parententry = KeyBone(
    descr="Parent",
    vfunc=is_depth_okay,
)
```


I there are now nested structures, which had already more than the 3 allowed parents (bc before this was not forbidden), the key would be illegal to use.
If u use now `singleValueFromClient` (where the `isInvalid` check is performed) in `singleValueUnserialize`, this would refuse to load the value. 

It is therefore no longer even possible to load the entity into a skeleton in order to change it with the new rules. in unserialise, only the format of the entry should be changed from database to skeleton format, but no ‘old’ data should be validated as invalid.

phorward: Hallo @sveneberth,

I've modified the call to `singleValueFromClient` to avoid calling the `isInvalid()` function. Does this resolve your considerations?

src/viur/core/bones/key.py
```diff
@@ -60,97 +59,54 @@ def singleValueFromClient(self, value, skel, bone_name, client_data):
...
-                return self.getEmptyValue(), [
-                    ReadFromClientError(
-                        ReadFromClientErrorSeverity.Invalid,
-                        "The provided key does not exist"
-                    )
-                ]
+        if not parse_only:
+            # Check custom validity
+            err = self.isInvalid(key)
+            if err:
```
ArneGudermann, line 65:
```suggestion
            
            if err := self.isInvalid(key):
```
Maybe we can simplify this

ArneGudermann (changes requested):

src/viur/core/bones/key.py
phorward: @sveneberth still awaiting your response.

sveneberth: > @sveneberth still awaiting your response.

@sveneberth is still waiting for https://github.com/viur-framework/vi-vue-utils/issues/13 to test this. But so far this looks okay, and there's no failing unittest, so it must be okay ^^

sveneberth (approved):

sveneberth (approved):

ArneGudermann (approved):

PR #1205, opened 2024-07-02
feat: Add `PhoneBone`

### LINKED ISSUE #1155: `PhoneBone` required
I need a `PhoneBone`, which accepts only valid phone numbers, including international ones.

Requirements:
- subclass of `StringBone`, type `"str.phone"` (likewise `"str.email"`)
- Accept phone numbers of the format "0123 45678", "+49 123 45678", "+49-123-45678", "0049 123 45678" (maybe there's standard regex for this)
- Automatically correct phone-numbers without country prefix, so "0123" becomes "+49 123" (configurable)
- Can also be used for fax numbers (in Germany, this is currently a very important feature)

KadirBalku:
Resolves #1155 

Default regex from PhoneBone seems working in Javascript as well.

Screenshot 2024-07-02 220000

Applied more variants in the console.log() expression and the regex is copied from the structure of my configured PhoneBone in my test Environment. All worked.


src/viur/core/bones/phone.py
```diff
@@ -0,0 +1,119 @@
...
+    The PhoneBone class is designed to store validated phone/fax numbers in configurable formats.
+    This class provides a number validation method, ensuring that the given phone/fax number conforms to the
+    required/configured format and structure.
+    """
+    type: str = "str.phone"
+    """
+    A string representing the type of the bone, in this case "str.phone".
+    """
+    def __init__(
+        self,
```
phorward, line 19: Parameters for bones are always kwarg-only, therefor you need to specify it. See https://peps.python.org/pep-3102/ for details.

```suggestion
        self,
        *,
```

src/viur/core/bones/phone.py
```diff
@@ -0,0 +1,119 @@
...
+        """
+        Initializes the PhoneBone with an optional custom regex for phone number validation, a default country code,
+        and a flag to apply the default country code if none is provided.
+
+        Args:
+            custom_regex (Optional[Pattern[str]]): An optional custom regex pattern for phone number validation.
+            max_length (int): The maximum length of the phone number.
+            default_country_code (str): The default country code to apply if none is provided.
+            apply_default_country_code (bool): Whether to apply the default country code if none is provided.
+            **kwargs (Any): Additional keyword arguments.
```
phorward, line 35: These are passed to the `StringBone`, you should mention this here shortly.

src/viur/core/bones/phone.py
```diff
@@ -0,0 +1,119 @@
...
+            return "No value entered"
+
+        if not self.phone_regex.match(value):
+            return "Invalid phone number entered"
+
+        is_invalid: Optional[str] = super().isInvalid(value)
+
+        if is_invalid:
+            return is_invalid
+        return None
```
phorward, line 75: You should add a super-call here to StringBones isValid-function. This allows further customization.

src/viur/core/bones/phone.py
```diff
@@ -0,0 +1,119 @@
...
+        Returns:
+            Tuple[Optional[str], Optional[List[ReadFromClientError]]]:
+            A tuple containing the processed phone number and an optional list of errors.
+        """
+        # Replace country code starting with 00 with +
+        if value.startswith('00'):
+            value = '+' + value[2:]
+
+        # Apply default country code if none is provided and apply_default_country_code is True
+        if self.apply_default_country_code and not value.startswith(('+', '00')):
```
phorward, line 99: When `value` previously started with "00", it is replaced by "+". The test for "00" again here makes no sense.
```suggestion
        if self.default_country_code and value[0] != "+":
```

src/viur/core/bones/phone.py
```diff
@@ -0,0 +1,119 @@
...
+            value (str): The phone number provided by the client.
+            skel (Any): Skeleton data (not used in this method).
+            bone_name (str): The name of the bone (not used in this method).
+            client_data (Any): Additional client data (not used in this method).
+
+        Returns:
+            Tuple[Optional[str], Optional[List[ReadFromClientError]]]:
+            A tuple containing the processed phone number and an optional list of errors.
+        """
+        # Replace country code starting with 00 with +
```
phorward, line 94: There should be a 
```py
        value = value.strip()
```
before

src/viur/core/bones/phone.py
```diff
@@ -0,0 +1,119 @@
...
+        if value.startswith('00'):
+            value = '+' + value[2:]
+
+        # Apply default country code if none is provided and apply_default_country_code is True
+        if self.apply_default_country_code and not value.startswith(('+', '00')):
+            if value.startswith('0'):
+                value = value[1:]  # Remove leading 0 from city code
+            value = self.default_country_code + value
```
phorward, line 102: I would suggest to add a space between country code and value?
```suggestion
            value = f"{self.default_country_code} {value}"
```

src/viur/core/bones/phone.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def structure(self) -> Dict[str, Any]:
+        """
+        Returns the structure of the PhoneBone, including the phone regex pattern.
+
+        Returns:
+            Dict[str, Any]: A dictionary representing the structure of the PhoneBone.
+        """
+        return super().structure() | {
+            "phone_regex": self.phone_regex.pattern if self.phone_regex else ""
```
phorward, line 118: Modify it according to function signature comments.

src/viur/core/bones/phone.py
```diff
@@ -0,0 +1,119 @@
...
+    required/configured format and structure.
+    """
+    type: str = "str.phone"
+    """
+    A string representing the type of the bone, in this case "str.phone".
+    """
+    def __init__(
+        self,
+        custom_regex: Optional[Pattern[str]] = None,
+        max_length: int = 15,
```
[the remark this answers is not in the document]
KadirBalku: default was set to 15 for test reasons, wanted to discuss about a realistic default rate with you but default of super class (255) was too high for my tests

src/viur/core/bones/phone.py
KadirBalku: i cant find a isValid function in StringBone neither in BaseBone, where is the definition of isValid ?
an ive added a super-call  to StringBone.isInvalid already

src/viur/core/bones/phone.py
```diff
@@ -0,0 +1,127 @@
...
+
+        Args:
+            value (str): The phone number provided by the client.
+            skel (t.Any): Skeleton data (not used in this method).
+            bone_name (str): The name of the bone (not used in this method).
+            client_data (t.Any): Additional client data (not used in this method).
+
+        Returns:
+            t.Tuple[t.Optional[str], t.Optional[t.List[ReadFromClientError]]]:
+            A tuple containing the processed phone number and an optional list of errors.
```
sveneberth, line 100: Please remove the types here, the type hints in the signature are enough.
And please use the [Sphinx docstring format](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html#the-sphinx-docstring-format).

sveneberth (changes requested):

src/viur/core/bones/phone.py
phorward: I've meant `isInvalid` - sorry. One of ViUR's many ugly names...

phorward:
@KadirBalku We're awaiting this pull request to become completed with viur-core 3.7, as there are also tests needed. Can you please apply the given suggestions, so that the server-related part of the code can be integrated?

src/viur/core/bones/phone.py
```diff
@@ -0,0 +1,129 @@
...
+
+from viur.core.bones.string import StringBone
+from viur.core.bones.base import ReadFromClientError, ReadFromClientErrorSeverity
+from viur.core import conf
+
+from pathlib import Path
+
+DEFAULT_REGEX = r"^\+?(\d{1,3})[-\s]?(\d{1,4})[-\s]?(\d{1,4})[-\s]?(\d{1,9})$"
+
+COUNTRYFILE = Path(__file__).parent / "../languages/country_information.json"
```
sveneberth, line 13: Can you add this file to the [`package-data`](https://github.com/viur-framework/viur-core/blob/dd0431bc14c2a2b897f2f54de092cac45b0d5e71/pyproject.toml#L70)? Otherwise, I think it wouldn't be part of the wheel later, or am I wrong @phorward? 

See [Data Files Support](https://setuptools.pypa.io/en/latest/userguide/datafiles.html)

src/viur/core/bones/phone.py
```diff
@@ -0,0 +1,129 @@
...
+        Initializes the PhoneBone with an optional custom regex for phone number validation, a default country code,
+        and a flag to apply the default country code if none is provided.
+
+        :param test: An optional custom regex pattern for phone number validation.
+        :param max_length: The maximum length of the phone number. Passed to 'StringBone'.
+        :param default_country_code: The default country code to apply if none is provided.
+        :param kwargs: Additional keyword arguments. Passed to 'StringBone'.
+        """
+        self.test: t.Pattern[str] = re.compile(test) if isinstance(test, str) else test
+        self.default_country_code: str = default_country_code
```
sveneberth, line 49: The parameter can be `None` or a `str`, but the instance variable is only `str`. How?

src/viur/core/bones/phone.py
```diff
@@ -0,0 +1,129 @@
...
+        default_country_code: t.Optional[str] = None,
+        **kwargs: t.Any,
+    ) -> None:
+        """
+        Initializes the PhoneBone with an optional custom regex for phone number validation, a default country code,
+        and a flag to apply the default country code if none is provided.
+
+        :param test: An optional custom regex pattern for phone number validation.
+        :param max_length: The maximum length of the phone number. Passed to 'StringBone'.
+        :param default_country_code: The default country code to apply if none is provided.
```
sveneberth, line 45: I find this none here unfavourable. Does it mean &laquo;_if `None` is provided_&raquo;   or  &laquo;_if no country code is provided in the client data_&raquo;?

sveneberth (dismissed):

src/viur/core/bones/phone.py
KadirBalku: This is for the bone definition, if u apply it in the definition the default country code will be prefixed automatically. if no default_country_code is given in the definition of the Bone, there will be no automatic appliance of any country code. But if a client adds 0049 for example the 00 will be replaced by a +. It was wanted like this

src/viur/core/bones/phone.py
phorward, line 49:
```suggestion
        self.default_country_code: t.Optional[str] = default_country_code
```

src/viur/core/bones/phone.py
phorward: @KadirBalku yes, please reformat the JSON to be pretty-printed, and add it to the package-data section.

phorward (changes requested):

src/viur/core/bones/phone.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def structure(self) -> t.Dict[str, t.Any]:
+        """
+        Returns the structure of the PhoneBone, including the test regex pattern.
+
+        :return: A dictionary representing the structure of the PhoneBone.
+        """
+        return super().structure() | {
+            "test": self.test.pattern if self.test else "",
+            "countries": self.countries,
```
phorward, line 127: Hello @KadirBalku, thanks for all improvements on this issue. I'm unhappy with this part of the pull request.

1. The JSON-file `country_information.json` is read into memory from the PhoneBone
2. The content is then assigned to `self.countries` of the PhoneBone
3. `self.countries`, the entire JSON-content from `country_information.json` is then passed to the client. 

This reveals to me that `country_information.json` does only contain information which is necessary to the client, and does nothing in the logic of the PhoneBone. Therefore, I would like to propose to move the content of `country_information.json` to the PhoneBone client source code.

phorward (changes requested):
@KadirBalku can you please move the client-only information to the client part of the PhoneBone, and keep the PhoneBone as small as possible?

The rest looks fine to me. I'm missing a way to specify which country codes are allowed, so that e.g. only phone numbers from Germany, Austria and Switzerland are allowed, and Germany is the default. But this might also be an extension apart from this PR.

src/viur/core/bones/phone.py
```diff
@@ -0,0 +1,119 @@
...
+        default_country_code: t.Optional[str] = None,
+        **kwargs: t.Any,
+    ) -> None:
+        """
+        Initializes the PhoneBone with an optional custom regex for phone number validation, a default country code,
+        and a flag to apply the default country code if none is provided.
+
+        :param test: An optional custom regex pattern for phone number validation.
+        :param max_length: The maximum length of the phone number. Passed to 'StringBone'.
+        :param default_country_code: The default country code to apply.
```
phorward, line 36: Can you clarify, what to apply? Is it the number prefix, "+49" or "0049", or just "49"? Please give an example.

src/viur/core/bones/phone.py
```diff
@@ -0,0 +1,119 @@
...
+            return "No value entered"
+
+        if not self.test.match(value):
+            return "Invalid phone number entered"
+
+        # make sure max_length is not exceeded.
+        is_invalid: t.Optional[str] = super().isInvalid(self._extract_digits(value))
+
+        if is_invalid:
+            return is_invalid
```
phorward, line 76: Simplify...
```suggestion
        if is_invalid := super().isInvalid(self._extract_digits(value)):
            return is_invalid

```

src/viur/core/bones/phone.py
```diff
@@ -0,0 +1,119 @@
...
+
+        The method checks if the provided phone number is valid according to the following criteria:
+        1. The phone number must not be empty.
+        2. The phone number must match the provided or default phone number format.
+        3. The phone number cannot exceed 15 digits, or the specified maximum length if provided (digits only).
+        """
+        if not value:
+            return "No value entered"
+
+        if not self.test.match(value):
```
phorward, line 69: What if self.test is None?
```suggestion
        if self.test and not self.test.match(value):
```

src/viur/core/bones/phone.py
```diff
@@ -0,0 +1,119 @@
...
+            value = "+" + value[2:]
+
+        # Apply default country code if none is provided and default_country_code is set
+        if self.default_country_code and value[0] != "+":
+            if value.startswith("0"):
+                value = value[1:]  # Remove leading 0 from city code
+            value = f"{self.default_country_code} {value}"
+
+        err: t.Optional[str] = self.isInvalid(value)
+        if err:
```
phorward, line 105: Simplify...
```suggestion
        if err := self.isInvalid(value):
```

phorward (changes requested):
@KadirBalku we're almost done, can you please review my comments?

phorward (approved):

phorward (approved):

phorward (approved):

PR #1209, opened 2024-07-10
feat: Add `SpamBone`

phorward:
Adds a simple math question as spam protection, that works without any third-party products.
We got a fine experience with this yet in several projects.

Requires #1208 to be fixed.

### DIFF
--- a/src/viur/core/bones/spam.py
+++ b/src/viur/core/bones/spam.py
@@ -57,6 +57,14 @@ class SpamBone(NumericBone):
 
         return i18n.translate(self.descr_template).translate(a=self.values[a - 1], b=self.values[b - 1])
 
+    @descr.setter
+    def descr(self, value: t.Any) -> None:
+        """
+        This setter is a null operation. It makes it possible to make a
+        super call in init, although the `descr` attribute is set there.
+        """
+        pass
+
     def isInvalid(self, value):
         session = current.session.get()
 
--- a/src/viur/core/bones/spam.py
+++ b/src/viur/core/bones/spam.py
@@ -1,39 +1,66 @@
-import random, logging
-from viur.core import utils, i18n, current
+import logging
+import random
+import typing as t
+from viur.core import i18n, current
 from viur.core.bones import NumericBone
 
 
 class SpamBone(NumericBone):
+    type = "numeric.spam"
 
-    def __init__(self, required=True, precision=0, *args, **kwargs):
-        super().__init__(required=required, precision=0, *args, **kwargs)
-        self.defaultNumbers = ["eins", "zwei", "drei", "vier", "fÃ¼nf", "sechs", "sieben", "acht", "neun"]
+    def __init__(
+        self,
+        descr: str = i18n.translate(
+            "core.bones.spam.question",
+            "What is the result of the addition of {{a}} and {{b}}?"
+        ),
+        values: t.Iterable[str] = (
+            i18n.translate(f"core.bones.spam.value.{digit}", digit)
+            for digit in ("one", "two", "three", "four", "five", "six", "seven", "eight", "nine")
+        ),
+        required: bool = True,
+        precision: int = 0,
+        msg_invalid: str = i18n.translate(
+            "core.bones.spam.invalid",
+            "Your answer was wrong! Please try again."
+        ),
+        **kwargs
+    ):
+        if precision != 0:
+            raise ValueError(f"Cannot use {self.__class__.__name__!r} with a precision")
 
-    def _getRandomNumber(self):
+        super().__init__(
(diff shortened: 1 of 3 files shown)


src/viur/core/bones/spam.py
```diff
@@ -0,0 +1,86 @@
...
+        self.descr_template = descr
+        self.values = tuple(values)
+        self.msg_invalid = msg_invalid
+
+    def _dice(self):
+        num = 0
+        while num == 0:
+            num = int(random.random() * len(self.values) + 1)
+
+        return num
```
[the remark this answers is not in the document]
phorward: Nice! :+1:

src/viur/core/bones/spam.py
```diff
@@ -0,0 +1,78 @@
...
+            "Your answer was wrong! Please try again."
+        ),
+        **kwargs
+    ):
+        if precision != 0:
+            raise ValueError(f"Cannot use {self.__class__.__name__!r} with a precision")
+
+        super().__init__(
+            required=required,
+            precision=precision,
```
sveneberth, line 34:
```suggestion
        msg_invalid: str = i18n.translate(
            "core.bones.spam.invalid",
            "Your answer was wrong! Please try again."
        ),
        **kwargs
    ):
        super().__init__(
            required=required,
            precision=0,
```

sveneberth (changes requested):

src/viur/core/bones/spam.py
phorward: Can you shortly explain what the change does?

sveneberth: It removes the precision argument from signature since it can only be 0 anyway

src/viur/core/bones/spam.py
```diff
@@ -0,0 +1,86 @@
...
+
+        if a is None or b is None:
+            a = session["spambone.value.a"] = self._dice()
+            b = session["spambone.value.b"] = self._dice()
+
+        return i18n.translate(self.descr_template).translate(a=self.values[a - 1], b=self.values[b - 1])
+
+    @descr.setter
+    def descr(self, value):
+        pass
```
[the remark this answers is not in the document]
sveneberth: Ohh, this was necessary. 

I'm getting now an error bc BaseBone's init tries to set it:

```py
  File "/project/deploy/skeletons/my.py", line 6, in <module>
    class MySkel(Skeleton):
  File "/project/deploy/skeletons/my.py", line 69, in MySkel
    spam = SpamBone()
           ^^^^^^^^^^
  File "/project/viur-core/src/viur/core/bones/spam.py", line 32, in __init__
    super().__init__(
  File "/project/viur-core/src/viur/core/bones/numeric.py", line 44, in __init__
    super().__init__(**kwargs)
  File "/project/viur-core/src/viur/core/bones/base.py", line 231, in __init__
    self.descr = descr
    ^^^^^^^^^^
  File "/project/viur-core/src/viur/core/bones/numeric.py", line 79, in __setattr__
    return super().__setattr__(key, value)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/project/viur-core/src/viur/core/bones/base.py", line 397, in __setattr__
    super().__setattr__(key, value)
AttributeError: property 'descr' of 'SpamBone' object has no setter
```

We have do re-add it, mut write a comment why it has to be a noop.

src/viur/core/bones/spam.py
phorward: yes, but with your suggestion its possible to provide `precision` to the kwargs, and this will raise a TypeError because precision is given twice in this case. My initial version fixes this problem.

sveneberth: Yes, but why is this a problem? Providing `precision=0` is redundant, bc it's already the default value and any other value would raise your `ValueError`.

phorward: So you think, when assigning a `precision` to the SpamBone() an error `TypeError: super().__init__() got multiple values for keyword argument 'precision'` will be more helpful, than `ValueError: Cannot use 'SpamBone' with precision`?

sveneberth: Not really, but the suggestion of my IDE to set the precision argument is not helpful too image

What do u think about this?

```suggestion
        msg_invalid: str = i18n.translate(
            "core.bones.spam.invalid",
            "Your answer was wrong! Please try again."
        ),
        **kwargs
    ):
        if "precision" in kwargs:
            raise ValueError(f"Cannot use {self.__class__.__name__!r} with a precision")

        super().__init__(
            required=required,
            precision=0,
```

sveneberth (approved):

PR #1210, opened 2024-07-11
feat: Add `EmailTransportAppengine` as default email transport class

phorward:
Brings back the default Google AppEngine Mail API as an abstracted `EmailTransport`-class.

PR #1212, opened 2024-07-11
refactor: `email`-module/`EmailTransportMailjet` fixes

phorward:
This refactors the entire `email`-module to the following:

- Clean-up code
- Use snake-case identifiers
- Use constants
- Check attachments correctly (gcsfile was never used, mimetype is required for Mailjet and GoogleAppEngineMailAPI)

This fixes the following problems in `EmailTransportMailjet`:

- Only the last attachment was added
- The "mimetype"-parameter to each attachment was invalid; it must be called "contenttype"
- Remove puremagic dependency again (it is not required, as the mimetype should be passed to the attachment)

Should come together in v3.6.12 with #1210.


src/viur/core/email.py
```diff
@@ -209,76 +218,93 @@ def sendEMail(*,
...
         raise ValueError("You have to set the params 'tpl' xor a 'stringTemplate'.")
-    if attachments:
+
+    if attachments := normalize_to_list(attachments):
         # Ensure each attachment has the filename key and rewrite each dict to db.Entity so we can exclude
         # it from being indexed
         for _ in range(0, len(attachments)):
             attachment = attachments.pop(0)
-            assert "filename" in attachment
+            assert all(att in attachment for att in ("filename", "content"))
```
sveneberth, line 242: Please don't enforce the content. I've have a `TransportClass` implemention where u can provide just the file key.

```suggestion
            assert "filename" in attachment
```

sveneberth (changes requested):

src/viur/core/email.py
phorward: Ok, but then, this must generally be reworked, I think.

sveneberth: Maybe we need a `transport_class.validate_attachment`... But I think this wasn't necessary before, so it shouldnt be a problem now either.

phorward: Yes, please LGTM!

sveneberth (approved):

PR #528, opened 2022-10-07
Rewrite `getReferencedBlobs` and `getSearchTags`

sveneberth:
* Introduce helper method `iter_bone_value`
* Handle in all bones all `multiple`/`languages` cases
* Eliminate most of duplicate code

Solves #503

PR #1213, opened 2024-07-12
fix: `RecordBone.getReferencedBlobs` should collect references for all bones

### LINKED ISSUE #1211: `RecordBone` with a `FileBone` inside having weak file references looses its content
This is a known problem with stacked RelationalBones, which is now documented in form of an issue.

This data model
```py
class SkelPartSkel(RelSkel):
	headline = StringBone(
            descr="Headline",
            searchable=True,
            languages=conf["viur.availableLanguages"]
	)

	image = FileBone(
            descr="Image",
            derive=conf["derives"]
	)

class SkelSkel(Skeleton):
        parts = RecordBone(
            descr="Abschnitt",
            format="$(headline.de)",
            multiple=True,
            using=SkelPartSkel
	)
```
stacks a `FileBone` in a multiple `RecordBone`.

Files, that are directly uploaded to the `image` bone in this case (without choosing it from the File module) are indicated as `weak`. Because the current design of ViUR does not support stacked bones, the weak file referencing doesn't apply in this case, and the file is being removed after a while.

sveneberth:
The `if not bone.searchable` condition introduced in #528 due to refactorings was too much.

Without this fix, weak files are not locked and will be deleted.

Fixes #1211

### DIFF
--- a/src/viur/core/bones/record.py
+++ b/src/viur/core/bones/record.py
@@ -1,5 +1,10 @@
 import json
+import typing as t
+
 from viur.core.bones.base import BaseBone, ReadFromClientError, ReadFromClientErrorSeverity
+
+if t.TYPE_CHECKING:
+    from ..skeleton import SkeletonInstance
 
 
 class RecordBone(BaseBone):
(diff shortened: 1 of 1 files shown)


phorward (approved):
Nice! :+1:

PR #1227, opened 2024-08-05
feat: Retrieve default `descr` from bone's name in its Skeleton

phorward:
This PR allows to define a bone

```py
test = StringBone()
```

which is shown as "test" bone, instead of just an empty string.

image


src/viur/core/bones/base.py
```diff
@@ -319,6 +319,18 @@ def __init__(
...
 
+    @property
+    def descr(self):
+        if self._descr:
+            return self._descr
+
+        if self.skel_cls:
+            for name, bone in self.skel_cls.__boneMap__.items():
+                if bone is self:
+                    return name
```
sveneberth, line 330: Not tested, but if `self.skel_cls` is set, `self.name` should be set as well. That should held the attribute name. Can you confirm this?

```suggestion
        if self.name:
            return self.name
```

sveneberth (changes requested):

src/viur/core/bones/base.py
phorward: yes! I was a little bit lost yesterday...

src/viur/core/bones/base.py
```diff
@@ -319,6 +319,10 @@ def __init__(
 
         self.compute = compute
 
+    @property
+    def descr(self):
+        return self._descr or self.name or ""
```
sveneberth, line 324: Can you merge the main --> develop first? I miss https://github.com/viur-framework/viur-core/pull/1223


Then we could think about

```suggestion
        return self._descr or translate(self.name) or ""
```

sveneberth (changes requested):

src/viur/core/bones/base.py
phorward: ok will do.

phorward: I've fixed it differently, see current diff.

sveneberth (changes requested):
Creating every time u access `.descr` a `translate` instance is not really efficient.

I have another solution: https://github.com/phorward/viur-core/pull/7/files

The translate object in the structure is btw no problem and needs to extra `str`-cast since we use our own `CustomJsonEncoder`.

phorward:
> I have another solution: [phorward#7](https://github.com/phorward/viur-core/pull/7)

It's kinda over-engineered, but we can make it so.

sveneberth (approved):

PR #1228, opened 2024-08-05
fix: `MultipleConstraints` as intended

phorward:
Fixes `MultipleConstraints`
    
- requirement for a hashed tuple
- renamed dataclass members*
- required modularization for `RelationalBone` and `RecordBone`
    
*) yes, this is a breaking change, BUT this feature isn't used in any ViUR project so far, I'm sure.

### DIFF
--- a/src/viur/core/bones/base.py
+++ b/src/viur/core/bones/base.py
@@ -11,6 +11,7 @@ import inspect
 import logging
 from dataclasses import dataclass, field
 from datetime import datetime, timedelta
+from collections.abc import Iterable
 from enum import Enum
 import typing as t
 from viur.core import db, utils, i18n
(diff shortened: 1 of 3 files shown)


src/viur/core/bones/base.py
```diff
@@ -132,12 +133,12 @@ class MultipleConstraints:  # Used to define constraints on multiple bones
...
     """
-    minAmount: int = 0  # Lower bound of how many entries can be submitted
+    min: int = 0  # Lower bound of how many entries can be submitted
     """An integer representing the lower bound of how many entries can be submitted (default: 0)."""
-    maxAmount: int = 0  # Upper bound of how many entries can be submitted
+    max: int = 0  # Upper bound of how many entries can be submitted
     """An integer representing the upper bound of how many entries can be submitted (default: 0)."""
-    preventDuplicates: bool = False  # Prevent the same value of being used twice
-    """A boolean value indicating if the same value can be used twice (default: False)."""
+    duplicates: bool = True  # Prevent the same value of being used twice
```
sveneberth, line 140: According to _principle of least privilege_, i think forbiding duplicates as longs as you not explicitly allow it, would not that bad. I currently see any use-case to allow duplicates, so I would prefer to reject this as long you really need this and enable it.


Following the _principle of least privilege_, I think disallowing duplicates as long as you do not explicitly allow them would not be so bad. I currently see no use case for allowing duplicates, so I would prefer to deny it as long as you really need it and enable it.

sveneberth, line 140: Since you negated the flag, this should be flipped as well
```suggestion
    duplicates: bool = True  # Allows the same value of being used twice
```

---
Btw; IMO it's totally nonsense to put a comment behind it and having an docstring. Can we extend the docstring and remove these inline comments. Then they would appear in our docs too.

sveneberth (changes requested):

sveneberth (approved):

ArneGudermann (approved):

Issue #1229, opened 2024-08-08
Redefine, rewrite and refactor any renderer stuff


phorward:
The current renderer stuff is a buggy mess of ugly, old code.
It should be entirely redesigned with a clear and straight concept.

Some issues with the current implementation:
- Optional function `_postProcessAppObj` used by json and xml should be converted into a standardized `Index`-module
- The renderers are currently constructed from the `renders` folder. This brings in several disadvantages, as the declared symbols are bad to distinguish.
- It is possible to register a module called `html` or `json`, with messes up with the specific renderer without any hint or error message
- A render refers to a module and a module to a renderer. This makes it impossible to instantiate each module and each renderer only once
- The concept of defining a module specific renderer, like with `user`, is nowhere else used in any ViUR project. By the concept of `ActionSkel`s, module-specific renders become obsolete and therefore the possibility should entirely be removed.

Issue relates to #1071.

PR #1230, opened 2024-08-08
fix: Cleanly collect renders from Python module

phorward:
This is still a hack, but it doesn't map any declared symbols like jinja global functions or imported stuff into the renderers-map as before.


phorward:
Relates to #1229

sveneberth (approved):

PR #1232, opened 2024-08-08
refactor: `__build_app` function clean-up and make all modules accessible

phorward:
This is both a refactor and a fix for the `Modulename.html = True` default that was removed with viur-core 3.6.

The error `AttributeError: 'Index' object has no attribute '<modulename>'` is fixed with this improvement, and the workaround of `conf.main_app.vi.modulename` is not neccessary anymore.

Belongs to #1229.


phorward:
We think about putting this into main branch v3.6

phorward:
Replaced by #1240

PR #1237, opened 2024-08-15
fix: Inconsistency raises AssertionError

phorward:
This change is for better debugging, but the process might still hang.

I'm quite unsure if this thing should better be ignored, or if `entry` itself should be removed, as it is an inconsistency.

### DIFF
--- a/src/viur/core/skeleton.py
+++ b/src/viur/core/skeleton.py
@@ -1408,9 +1408,13 @@ def processRemovedRelations(removedKey, cursor=None):
         .filter("viur_relational_consistency >", 2)
     updateListQuery = updateListQuery.setCursor(cursor)
     updateList = updateListQuery.run(limit=5)
+
     for entry in updateList:
         skel = skeletonByKind(entry["viur_src_kind"])()
-        assert skel.fromDB(entry["src"].key)
+        if not skel.fromDB(entry["src"].key):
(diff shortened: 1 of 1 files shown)


src/viur/core/skeleton.py
```diff
@@ -1408,9 +1408,13 @@ def processRemovedRelations(removedKey, cursor=None):
         .filter("viur_relational_consistency >", 2)
     updateListQuery = updateListQuery.setCursor(cursor)
     updateList = updateListQuery.run(limit=5)
+
     for entry in updateList:
         skel = skeletonByKind(entry["viur_src_kind"])()
-        assert skel.fromDB(entry["src"].key)
+        if not skel.fromDB(entry["src"].key):
+            logging.error(f"processRemovedRelations detects inconsistency on src={entry['src'].key!r}")
+            raise ValueError()
```
sveneberth, line 1416: Can you add the message to the `ValueError` exception too?

sveneberth (changes requested):

src/viur/core/skeleton.py
phorward, line 1416:
```suggestion
            raise ValueError(f"processRemovedRelations detects inconsistency on src={entry['src'].key!r}")
```

phorward: Fixed!

sveneberth (approved):

Issue #1219, opened 2024-07-18
`URIBone` needed


phorward:
Similar to #1155, why isn't there an `URIBone` to specify valid URLs and URIs?

The requirements must be evaluated, here are some ideas:

- Accepted protocol (http, https...)
- Accept, or accepted ports?
- Local path, may start or not start with `/`
- Clean-up
  - accept or deny GET-parameters
  - domains disallow list

Relates to #1218, I already missed such a bone for several use-cases in the past.

phorward:
Fixed by #1254

PR #1241, opened 2024-08-20
feat: `FileBone(public=True)` for public files

akelch:
This PR adds public files. Public files are stored in an extra bucket that follows this name: public-dot-{__PROJECT_ID}
The bucket must be created with ACL rights.

With getUploadUrl, the parameter public is used to decide whether it is a public file or not.
Filebones with flag public=True can only refer to a public file

Public files again deliver a servingurl.


src/viur/core/modules/file.py
```diff
@@ -361,6 +368,17 @@ class FileLeafSkel(TreeSkel):
         readOnly=True,
     )
 
+    public = BooleanBone(
+        descr="Public File",
+        readOnly=True,
+        defaultValue=False,
+    )
+
+    serving_url = StringBone(
```
phorward, line 377: Shall we use a RawBone here, or URIBone #1219

src/viur/core/modules/file.py
```diff
@@ -656,7 +711,11 @@ def read(self, key: db.Key | int | str | None = None, path: str | None = None) -
             else:
                 path = f"""{skel["dlkey"]}/source/{skel["name"]}"""
 
-        blob = GOOGLE_STORAGE_BUCKET.blob(path)
+            bucket = File.get_bucket(skel["dlkey"])
+        else:
+            bucket = File.get_bucket(PUBLIC_DLKEY_POSTFIX if public else "")
```
phorward, line 716: This is my proposal, its a little bit ugly, but it's concise.

src/viur/core/modules/file.py
```diff
@@ -757,6 +821,7 @@ def getUploadURL(
         file_skel["pendingparententry"] = db.keyHelper(node, self.addSkel("node").kindName) if node else None
         file_skel["pending"] = True
         file_skel["weak"] = True
+        file_skel["public"] = public
```
phorward, line 824: This was 
```py
        file_skel["public"] = True
```
in the original PR, but I think it's wrong?! @akelch

src/viur/core/modules/file.py
```diff
@@ -869,6 +938,109 @@ def download(self, blobKey: str, fileName: str = "", download: bool = False, sig
...
 
+    @exposed
+    def serve(
+        self,
+        key: str,
+        size: t.Optional[int] = None,
+        filename: t.Optional[str] = None,  # random string with .ext
+        options: str = "",
+        download: bool = False,
+    ):
```
phorward, line 949: Refactored this. The type annotations are used by ViUR 3 to parse specific types. @akelch the original version looked a little bit old, I hope that I've refactored it correctly.

src/viur/core/modules/file.py
```diff
@@ -969,6 +1150,26 @@ def mark_for_deletion(self, dlkey: str) -> None:
...
+                )
+            except Exception as e:
+                logging.warning("Error while creating serving url")
+                if not conf.instance.is_dev_server:
+                    logging.exception(e)
+        return skel
+
 
 File.json = True
 File.html = True
```
phorward, line 1175: Why is the `File`-module open to html and json by default? I want to eliminate this, it depends on the project.

phorward (changes requested):
@akelch can you test this branch again with my refactoring?
@sveneberth @ArneGudermann looking phorward for your review!

src/viur/core/modules/file.py
```diff
@@ -677,22 +695,26 @@ def read(self, key: db.Key | int | str | None = None, path: str | None = None) -
...
                 if not path:
                     raise ValueError("This skeleton is not in the database!")
             else:
                 path = f"""{skel["dlkey"]}/source/{skel["name"]}"""
-            current_bucket = get_current_bucket(skel["dlkey"])
 
-        blob = current_bucket.blob(path)
+            bucket = File.get_bucket(skel["dlkey"])
+        else:
+            bucket = File.get_bucket(PUBLIC_DLKEY_POSTFIX if public else "")
```
akelch, line 718: This will not work because the prefix is not a dlkey. The public parameter is not necessary, as the type is defined by the dlkey in both the skel and the path

src/viur/core/modules/file.py
```diff
@@ -610,15 +654,20 @@ def write(self, filename: str, content: t.Any, mimetype: str = "text/plain", wid
...
         :return: Returns the key of the file object written. This can be associated e.g. with a FileBone.
         """
         if not File.is_valid_filename(filename):
             raise ValueError(f"{filename=} is invalid")
 
         dl_key = utils.string.random()
 
-        blob = GOOGLE_STORAGE_BUCKET.blob(f"{dl_key}/source/{filename}")
+        if public:
+            dl_key += "_pub"  # mark file as public
```
akelch, line 666: use PUBLIC_DLKEY_POSTFIX

src/viur/core/modules/file.py
```diff
@@ -742,7 +804,11 @@ def getUploadURL(
 
         # Create upload-URL and download key
         dlkey = utils.string.random()  # let's roll a random key
-        blob = GOOGLE_STORAGE_BUCKET.blob(f"{dlkey}/source/{filename}")
+
+        if public:
+            dlkey += "_pub"  # mark folder as public
```
akelch, line 809: use PUBLIC_DLKEY_POSTFIX

src/viur/core/modules/file.py
akelch: yes, this was wrong, this is how i tested it in the admin

src/viur/core/modules/file.py
phorward: Can you clarify with a suggestion? In case there is only a path provided, there is still no way to define whether the path belongs to the public or private bucket.

src/viur/core/modules/file.py
```diff
@@ -969,6 +1147,26 @@ def mark_for_deletion(self, dlkey: str) -> None:
...
+                bucket = File.get_bucket(skel['dlkey'])
+                skel["serving_url"] = images.get_serving_url(
+                    None,
+                    secure_url=True,
+                    filename=f"/gs/{bucket.name}/{skel['dlkey']}/source/{skel['name']}",
+                )
+            except Exception as e:
+                logging.warning("Error while creating serving url")
+                if not conf.instance.is_dev_server:
+                    logging.exception(e)
```
ArneGudermann, line 1167:
```suggestion
                logging.exception(e)
```
We don't need this check I think. Because we check this above.

src/viur/core/modules/file.py
```diff
@@ -422,6 +444,21 @@ class File(Tree):
...
+        """
+        if dlkey and dlkey.endswith(PUBLIC_DLKEY_POSTFIX):
+            if public_bucket := GOOGLE_STORAGE_CLIENT.lookup_bucket(f"""public-dot-{_PROJECT_ID}"""):
+                return public_bucket
+
+            raise ValueError(
+                f"""The bucket 'public-dot-{_PROJECT_ID}' does not exist! Please create it with ACL access."""
+            )
+
+        return GOOGLE_STORAGE_CLIENT.lookup_bucket(f"""{_PROJECT_ID}.appspot.com""")
```
ArneGudermann, line 460:
```suggestion
def get_bucket(dlkey: str) -> google.cloud.storage.bucket.Bucket:
	"""
	Retrieves a Google Cloud Storage bucket for the given dlkey.
	"""
	if dlkey and dlkey.endswith(PUBLIC_DLKEY_POSTFIX):
		if public_bucket := conf.file_buckets.get(f"""public-dot-{_PROJECT_ID}"""):
			return public_bucket
		if public_bucket := GOOGLE_STORAGE_CLIENT.lookup_bucket(f"""public-dot-{_PROJECT_ID}"""):
			conf.file_buckets[f"""public-dot-{_PROJECT_ID}"""] = public_bucket
			return public_bucket

		raise ValueError(
			f"""The bucket 'public-dot-{_PROJECT_ID}' does not exist! Please create it with ACL access."""
		)
	
	if private_bucket := conf.file_buckets.get(f"""{_PROJECT_ID}.appspot.com"""):
		return private_bucket
	
	private_bucket = conf.file_buckets.get(f"""{_PROJECT_ID}.appspot.com"""):
	conf.file_buckets[f"""{_PROJECT_ID}.appspot.com"""] = private_bucket
	return private_bucket
```
We can solve it like this than we can save us lookups we only need `file_buckets` in `conf`

ArneGudermann (changes requested):

src/viur/core/modules/file.py
```diff
@@ -34,15 +36,17 @@
     re.IGNORECASE
 )
 
-_CREDENTIALS, __PROJECT_ID = google.auth.default()
-GOOGLE_STORAGE_CLIENT = storage.Client(__PROJECT_ID, _CREDENTIALS)
-GOOGLE_STORAGE_BUCKET = GOOGLE_STORAGE_CLIENT.lookup_bucket(f"""{__PROJECT_ID}.appspot.com""")
+_CREDENTIALS, _PROJECT_ID = google.auth.default()
+GOOGLE_STORAGE_CLIENT = storage.Client(_PROJECT_ID, _CREDENTIALS)
+PUBLIC_DLKEY_POSTFIX = "_pub"
```
phorward, line 41: Rename to suffix.
```suggestion
PUBLIC_DLKEY_SUFFIX = "_pub"
```

src/viur/core/modules/file.py
phorward: Hello @ArneGudermann, we can either put it into conf or make it global in file.py... latter one is closer to the implementation before the change. I wouldn't put it into the config.

phorward: @ArneGudermann are you happy with e19a204c09ad2d5f399bfaba6f8e920b11ce15f1?

src/viur/core/modules/file.py
```diff
@@ -869,6 +944,106 @@ def download(self, blobKey: str, fileName: str = "", download: bool = False, sig
...
+        if any(c not in conf.search_valid_chars for c in host):
+            raise errors.BadRequest("key contains invalid characters")
+
+        # extract format from filename
+        file_fmt = "webp"
+
+        if filename:
+            fmt = filename.rsplit(".", 1)[-1].lower()
+            if fmt in VALID_FMTS:
+                file_fmt = fmt
```
sveneberth, line 1019: I know, who would debug and searching for hours why `avif` is always rendered as `webp`…


```suggestion
            if fmt in VALID_FMTS:
                file_fmt = fmt
            else:
                raise errors.UnprocessableEntity(f"Unsupported filetype {fmt}")
```

src/viur/core/modules/file.py
```diff
@@ -869,6 +944,106 @@ def download(self, blobKey: str, fileName: str = "", download: bool = False, sig
...
+            p - face crop
+            fv - vertrical flip
+            fh - horizontal flip
+            rXXX - rotate 90, 180, 270
+            nu - no upscale
+        :param download: if value = 1 force download, else not
+        :return:
+        """
+
+        VALID_OPTIONS = {
```
sveneberth, line 973: Why aren't these constants class variables, which would allow monkeypatching / subclasse overwriting for project requirements?

src/viur/core/modules/file.py
```diff
@@ -610,15 +660,20 @@ def write(self, filename: str, content: t.Any, mimetype: str = "text/plain", wid
...
         if not File.is_valid_filename(filename):
             raise ValueError(f"{filename=} is invalid")
 
         dl_key = utils.string.random()
 
-        blob = GOOGLE_STORAGE_BUCKET.blob(f"{dl_key}/source/{filename}")
+        if public:
+            dl_key += PUBLIC_DLKEY_SUFFIX  # mark file as public
+
+        bucket = File.get_bucket(dl_key)
```
sveneberth, line 674: Please use `self` instead of the class to allow subclassing.
```suggestion
        bucket = self.get_bucket(dl_key)
```

(Please change all `File.$any` to `self.$any`)

And I would change external `File.$any` to `conf.main_app.file.$any`

src/viur/core/modules/file.py
```diff
@@ -869,6 +944,106 @@ def download(self, blobKey: str, fileName: str = "", download: bool = False, sig
...
+            response.headers["Cache-Control"] = "public, max-age=604800"  # 7 Days
+            if download:
+                response.headers["Content-Disposition"] = f"attachment; filename={filename}"
+            else:
+                response.headers["Content-Disposition"] = f"filename={filename}"
+
+            return requests.get(url).content
+
+        except Exception as e:
+            raise errors.BadRequest("Invalid Url")
```
sveneberth, line 1045: what kind of poor error handling is that?

```suggestion
        response = current.request.get().response
        response.headers["Content-Type"] = f"image/{file_fmt}"
        response.headers["Cache-Control"] = "public, max-age=604800"  # 7 Days
        if download:
            response.headers["Content-Disposition"] = f"attachment; filename={filename}"
        else:
            response.headers["Content-Disposition"] = f"filename={filename}"
        
        answ = requests.get(url, timeout=20)
        if not answ.ok:
            raise errors.BadRequest("Unable to fetch a file with these parameters")
```

**Notes:**
- only as much as necessary into the `try`
- Limit the requests since we have limited frontend request

src/viur/core/modules/file.py
```diff
@@ -869,6 +944,106 @@ def download(self, blobKey: str, fileName: str = "", download: bool = False, sig
...
+        url += "=" + options
+
+        try:
+            response = current.request.get().response
+            response.headers["Content-Type"] = f"image/{file_fmt}"
+            response.headers["Cache-Control"] = "public, max-age=604800"  # 7 Days
+            if download:
+                response.headers["Content-Disposition"] = f"attachment; filename={filename}"
+            else:
+                response.headers["Content-Disposition"] = f"filename={filename}"
```
sveneberth, line 1040: Shouldn't this be better quoted to avoid problems with spaces?

```suggestion
            if download:
                response.headers["Content-Disposition"] = f'attachment; filename="{filename}"'
            else:
                response.headers["Content-Disposition"] = f'filename="{filename}"'
```

src/viur/core/modules/file.py
```diff
@@ -969,6 +1153,25 @@ def mark_for_deletion(self, dlkey: str) -> None:
...
+                and skel["mimetype"].startswith("image/") and not skel["serving_url"]:
+
+            try:
+                bucket = File.get_bucket(skel['dlkey'])
+                skel["serving_url"] = images.get_serving_url(
+                    None,
+                    secure_url=True,
+                    filename=f"/gs/{bucket.name}/{skel['dlkey']}/source/{skel['name']}",
+                )
+            except Exception as e:
```
sveneberth, line 1170: Why is it caught like this and not reraised or another return value used?

sveneberth (changes requested):

src/viur/core/modules/file.py
```diff
@@ -869,6 +944,106 @@ def download(self, blobKey: str, fileName: str = "", download: bool = False, sig
 
         raise errors.Redirect(signedUrl)
 
+    @exposed
+    def serve(
```
sveneberth, line 948: How can I generate these URLs? How do I know the _host prefix_?

src/viur/core/modules/file.py
```diff
@@ -869,6 +944,108 @@ def download(self, blobKey: str, fileName: str = "", download: bool = False, sig
...
+            raise errors.BadRequest("Invalid options provided")
+
+        options += f"-{VALID_FMTS[file_fmt]}"
+
+        if size and size in VALID_SIZES:
+            options = f"s{size}-" + options
+
+        url += "=" + options
+
+        try:
```
phorward, line 1035: Original comment by @sveneberth readded, as GitHub had a hicc-up with too many changes in a row...

---
what kind of poor error handling is that?

```suggestion
        response = current.request.get().response
        response.headers["Content-Type"] = f"image/{file_fmt}"
        response.headers["Cache-Control"] = "public, max-age=604800"  # 7 Days
        if download:
            response.headers["Content-Disposition"] = f"attachment; filename={filename}"
        else:
            response.headers["Content-Disposition"] = f"filename={filename}"
        
        answ = requests.get(url, timeout=20)
        if not answ.ok:
            raise errors.BadRequest("Unable to fetch a file with these parameters")
```

**Notes:**
- only as much as necessary into the `try`
- Limit the requests since we have limited frontend request

src/viur/core/modules/file.py
phorward: I would like to put this into a separate issue. There are also calls in render/html on this.

phorward: Referenced by #1252, we shall keep it as is here for now.

src/viur/core/modules/file.py
phorward: Constants moved to `File`-class.

src/viur/core/modules/file.py
phorward: Pull request #1253 fixes this, and other related issues.

src/viur/core/modules/file.py
```diff
@@ -869,6 +944,106 @@ def download(self, blobKey: str, fileName: str = "", download: bool = False, sig
...
+        filename: t.Optional[str] = None,  # random string with .ext
+        options: str = "",
+        download: bool = False,
+    ):
+        """
+        Requests an image using the serving url to bypass direct Google requests.
+
+        :param key: a string with one __ seperator. The first part is the host prefix, the second part is the key
+        :param size: the target image size, take a look at the VALID_SIZES
+        :param filename: a random string with an extention, valid extentions are jpg,png,webp
```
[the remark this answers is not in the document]
akelch: no

src/viur/core/modules/file.py
akelch: the serving url must be split. The only alternative could be to save the split in the skel.

for js frontends there will be a function in vue-utils. 
for jinja we could add a filter for this, if needed

src/viur/core/modules/file.py
```diff
@@ -869,6 +944,106 @@ def download(self, blobKey: str, fileName: str = "", download: bool = False, sig
...
+        key: str,
+        size: t.Optional[int] = None,
+        filename: t.Optional[str] = None,  # random string with .ext
+        options: str = "",
+        download: bool = False,
+    ):
+        """
+        Requests an image using the serving url to bypass direct Google requests.
+
+        :param key: a string with one __ seperator. The first part is the host prefix, the second part is the key
```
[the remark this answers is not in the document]
akelch: okay, what do you prefer 2 parameters or one @?

src/viur/core/modules/file.py
```diff
@@ -869,6 +944,106 @@ def download(self, blobKey: str, fileName: str = "", download: bool = False, sig
...
+            "p",
+            "fv",
+            "fh",
+            "r90",
+            "r180",
+            "r270",
+            "nu"
+        }
+
+        VALID_SIZES = {
```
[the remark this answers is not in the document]
akelch: goog point, i tested these again. it seams that these limits are not used anymore. I will replace this with a simple isdigit check
https://github.com/GoogleCloudPlatform/appengine-python-standard/blob/f10e3de6d53752b3aa6046e4c5436b6f81cd699f/src/google/appengine/api/images/__init__.py#L1844

src/viur/core/modules/file.py
akelch: The intention behind this was that an error during generation would not prevent saving.

src/viur/core/modules/file.py
sveneberth: okay

src/viur/core/modules/file.py
sveneberth: Okay, Can you extend `File.create_download_url()` or implement a similar method for this?

src/viur/core/modules/file.py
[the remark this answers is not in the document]
sveneberth: I would prefer a second parameter, so you don't have to worry about any separation and can pass the arguments in a known way.

src/viur/core/modules/file.py
akelch: There are now 2 new methods:
create_serve_parameters: takes a serving url and splits to serve parameters host and key
create_serve_url function: generates serve url from a google servingurl, structure will be like this:
/file/serve/lh3/SOMEKEY/2000/test.webp?options=r270-fv&download=true

src/viur/core/modules/file.py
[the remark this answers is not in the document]
akelch: Sizes are no longer limited

src/viur/core/modules/file.py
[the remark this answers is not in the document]
akelch: The serve function now takes a host and a key parameter

ArneGudermann (approved):

src/viur/core/modules/file.py
```diff
@@ -449,6 +493,53 @@ def hmac_sign(data: t.Any) -> str:
...
+
+        serve_url = f"{File.SERVE_URL_PREFIX}"
+
+        serving_params = File.create_serve_parameters(serving_url)
+        if not serving_params:
+            raise errors.UnprocessableEntity(f"Invalid serving_url {serving_url!r} provided")
+
+        serve_url += f"/{'/'.join(serving_params)}"
+
+        path_parameters = [f"{x}" for x in [size, filename] if x is not None]
```
sveneberth, line 529: This does not work, if `filename` is provided, but not `size`. Then, `filename` would be used as `size`.

I would suggest to switch the order of the `filename` and `size` parameter in `serve()` and here in `create_serve_url()` and append `size` always as keyword argument in the query string. 

To have the filename as last positional has btw. the advantage, that browsers use the last url path segment if a file is displayed inline for the save dialog and tab title.

src/viur/core/modules/file.py
```diff
@@ -449,6 +493,53 @@ def hmac_sign(data: t.Any) -> str:
...
+    @staticmethod
+    def create_serve_url(
+        serving_url: str,
+        size: t.Optional[int] = None,
+        filename: t.Optional[str] = None,
+        options: str = "",
+        download: bool = False
+    ) -> str:
+
+        serve_url = f"{File.SERVE_URL_PREFIX}"
```
sveneberth, line 521: It's already a string

```suggestion
        serve_url = File.SERVE_URL_PREFIX
```

src/viur/core/modules/file.py
```diff
@@ -449,6 +493,53 @@ def hmac_sign(data: t.Any) -> str:
...
+            serving_url
+        )
+
+        if regex_result:
+            return regex_result.groups()
+        return regex_result
+
+    @staticmethod
+    def create_serve_url(
+        serving_url: str,
```
sveneberth, line 514: Can we make the method a little bit more handy?

So that I only have to pass it a file skel (`SkeletonInstance` for a `FileSkel`) or dest object (`SkeletonInstance` for a `RefSkelForFile`), and it will automatically fill in the `servingurl` and `filename` from it.

So I either pass a servingurl or fileskel?!

src/viur/core/modules/file.py
sveneberth: But how do I know if everything was OK or not when I call this method if neither an exception is raised nor a suitable return value is returned?

sveneberth (changes requested):

src/viur/core/modules/file.py
```diff
@@ -449,6 +493,53 @@ def hmac_sign(data: t.Any) -> str:
     def hmac_verify(data: t.Any, signature: str) -> bool:
         return hmac.compare_digest(File.hmac_sign(data.encode("ASCII")), signature)
 
+    @staticmethod
+    def create_serve_parameters(serving_url: str) -> t.Iterable[str] | None:
```
phorward, line 497: Stay conform to your own suggestion below!
```suggestion
    def create_serve_parameters(serving_url: str) -> tuple[str, str]:
```

src/viur/core/modules/file.py
phorward: The `logging.exception(e)` is enough for now.

src/viur/core/modules/file.py
phorward: Either you create a proposal for the function, or we use it now as is, in the first iteration. This is a big feature and nice that we have a solution for it right now, so lets get things merged. v3.7 is still not Final.

phorward (dismissed):
I've tested together with @akelch.

ArneGudermann (dismissed):

src/viur/core/modules/file.py
phorward: You're right, need to fix this.

phorward: Fixed by a94b9226

phorward (approved):

PR #1244, opened 2024-08-21
fix: Re-Implement `RelationalConsistency.PreventDeletion` validation

sveneberth:
You can set `consistency` on a `RelationalBone` to `RelationalConsistency.PreventDeletion` to disallow the deletion of the referenced Skeleton (if there is a relation from this bone to it).

This works fine as long a Skeleton is referenced by only few other Skeleton. But if thousands of Skeletons has a reference to the same Skeleton (10000:1 relation) the `viur_incomming_relational_locks` list has to hold 10000 keys, which exceeds the size limit of the entity. Furthermore the transaction limit would be exceeded during writing on 10000 entites the `"{bonename}_outgoingRelationalLocks"` property too.
We cannot carry a list, neither in the one way nor in the other. (Yes, okay, theoretically you could write _outgoingRelationalLocks outside the transaction, maybe in a task. But even here it can be dangerous if a skeleton contains many (multiple) `RelationalBone`. Even if the number of outgoing ones will probably be smaller than the number of incoming ones.  But this is still an insane overhead).

That's why I thought about what I could use to solve this. And in my opinion, we can simply use what we already have: The `viur-relations` entities. The `consistency` is already stored in this. 
So I have removed everything that relates to the maintenance of these lock lists, ensured that these lists will no longer be present in the datastore entites in future and replaced the "can delete" check with a query on the `viur-relations`.
In my tests this works very well and we should even save some runtime and memory.

### DIFF
--- a/src/viur/core/bones/relational.py
+++ b/src/viur/core/bones/relational.py
@@ -869,8 +869,10 @@ class RelationalBone(BaseBone):
         origFilter = dbFilter.queries
         if origFilter is None or not "orderby" in rawFilter:  # This query is unsatisfiable or not sorted
             return dbFilter
-        if "orderby" in rawFilter and isinstance(rawFilter["orderby"], str) and rawFilter["orderby"].startswith(
-            f"{name}."):
+        if (
+            "orderby" in rawFilter and isinstance(rawFilter["orderby"], str)
+            and rawFilter["orderby"].startswith(f"{name}.")
+        ):
             if not dbFilter.getKind() == "viur-relations" and self.multiple:  # This query has not been rewritten (yet)
                 name, skel, dbFilter, rawFilter = self._rewriteQuery(name, skel, dbFilter, rawFilter)
             key = rawFilter["orderby"]
@@ -1143,16 +1145,15 @@ class RelationalBone(BaseBone):
         Sets the value of the specified bone in the given skeleton. Sanity checks are performed to ensure the
         value is valid. If the value is invalid, no modifications are made.
 
-        :param SkeletonInstance skel: Dictionary with the current values from the skeleton we belong to.
-        :param str boneName: The name of the bone to be modified.
+        :param skel: Dictionary with the current values from the skeleton we belong to.
+        :param boneName: The name of the bone to be modified.
         :param value: The value to be assigned. The type depends on the bone type.
-        :param bool append: If true, the given value is appended to the values of the bone instead of replacing it.
-                   Only supported on bones with multiple=True.
-        :param Union[None, str] language: Set/append for a specific language (optional). Required if the bone
+        :param append: If true, the given value is appended to the values of the bone instead of replacing it.
+            Only supported on bones with multiple=True.
+        :param language: Set/append for a specific language (optional). Required if the bone
             supports languages.
 
         :return: True if the operation succeeded, False otherwise.
-        :rtype: bool
         """
         assert not (bool(self.languages) ^ bool(language)), "Language is required or not supported"
         assert not append or self.multiple, "Can't append - bone is not multiple"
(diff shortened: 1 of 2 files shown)


src/viur/core/bones/relational.py
```diff
@@ -482,29 +477,8 @@ def serialize(self, skel: 'SkeletonInstance', name: str, parentIndexed: bool) ->
...
-                    referencedObj["viur_incomming_relational_locks"].append(skel["key"])
-                    db.Put(referencedObj)
-
-        for oldLock in oldRelationalLocks - newRelationalLocks:
-            # Remove Lock
-            if referencedObj := db.Get(oldLock):
-                if skel["key"] in referencedObj["viur_incomming_relational_locks"]:
-                    referencedObj["viur_incomming_relational_locks"].remove(skel["key"])
-                    db.Put(referencedObj)
+        # Delete legacy property
```
phorward, line 480:
```suggestion
        # FIXME VIUR4: Delete legacy property
```

src/viur/core/bones/relational.py
```diff
@@ -620,25 +566,25 @@ def postSavedHandler(self, skel, boneName, key):
...
-        dbVals.filter("src.__key__ =", key)
-        db.Delete([x for x in dbVals.run()])
-
-    def isInvalid(self, key):
+        query = db.Query("viur-relations")
+        query.filter("viur_src_kind =", skel.kindName)
+        query.filter("viur_dest_kind =", self.kind)
+        query.filter("viur_src_property =", boneName)
+        query.filter("src.__key__ =", key)
+        db.Delete([x for x in query.run()])
```
phorward, line 585:
```suggestion
        db.Delete([entity for entity in query.run()])
```

src/viur/core/bones/relational.py
```diff
@@ -923,8 +869,10 @@ def buildDBSort(
         origFilter = dbFilter.queries
         if origFilter is None or not "orderby" in rawFilter:  # This query is unsatisfiable or not sorted
             return dbFilter
-        if "orderby" in rawFilter and isinstance(rawFilter["orderby"], str) and rawFilter["orderby"].startswith(
-                f"{name}."):
+        if (
+            "orderby" in rawFilter and isinstance(rawFilter["orderby"], str)
+            and rawFilter["orderby"].startswith(f"{name}.")
+        ):
```
phorward, line 875: Can you please keep the ugly code, as it is refactored by #1077 already for viur-core 3.7 and results in a merge conflict.

```suggestion
        if "orderby" in rawFilter and isinstance(rawFilter["orderby"], str) and rawFilter["orderby"].startswith(
                f"{name}."):
```

src/viur/core/bones/relational.py
```diff
@@ -1217,27 +1163,33 @@ def setBoneValue(
...
+            ):
                 raise ValueError(f"You must supply a tuple of (Database-Key, relSkel) to {boneName}")
             realValue = value
         elif self.multiple and not self.using:
-            if not (isinstance(value, str) or isinstance(value, db.Key)) and not (isinstance(value, list)) \
-                    and all([isinstance(x, str) or isinstance(x, db.Key) for x in value]):
+            if (
+                not (isinstance(value, str) or isinstance(value, db.Key))
+                and not (isinstance(value, list))
+                and all(isinstance(x, (str, db.Key)) for x in value)
```
phorward, line 1177:
```suggestion
                and all(isinstance(k, (str, db.Key)) for k in value)
```

src/viur/core/skeleton.py
```diff
@@ -1048,6 +1049,9 @@ def txnUpdate(write_skel):
                         else:
                             logging.critical("Detected Database corruption! Could not delete stale lock-object!")
 
+            # Delete legacy property
```
phorward, line 1052:
```suggestion
            # FIXME VIUR4: Delete legacy property
```

src/viur/core/skeleton.py
```diff
@@ -1404,14 +1416,17 @@ def __init__(self, baseSkel=None):
...
+        .filter("dest.__key__ =", removedKey)
+        .filter("viur_relational_consistency >", RelationalConsistency.PreventDeletion.value)
+    )
     updateListQuery = updateListQuery.setCursor(cursor)
     updateList = updateListQuery.run(limit=5)
     for entry in updateList:
         skel = skeletonByKind(entry["viur_src_kind"])()
         assert skel.fromDB(entry["src"].key)
-        if entry["viur_relational_consistency"] == 3:  # Set Null
+        if entry["viur_relational_consistency"] == RelationalConsistency.SetNull.value:
```
phorward, line 1429: lol! :+1:

phorward (changes requested):
Hello @sveneberth, this is a very nice solution!
Thanks for kicking out this ugly and misspelled `viur_incomming_relational_locks`...

Can you please take a look to my comments, it's only minor stuff.

src/viur/core/bones/relational.py
sveneberth: well okay. The linter always disagrees with PyCharms, so I'll probably have to mediate manually again.

phorward: well okay, then leave it as is, I'll manually resolve the conflict.

phorward (dismissed):
I'm fine with this, although it are very huge change for a "bugfix release" 3.6.14 ;)

phorward (approved):

sveneberth:
> I'm fine with this, although it are very huge change for a "bugfix release" 3.6.14 ;)

#1240 and #1228 were also huge :) SemVer hasn't been here at viur-core for a long time anyway… :/

PR #973, opened 2023-12-07
refactor: `Skeleton.toDB` method

ArneGudermann:

PR #1248, opened 2024-08-28
fix: `Skeleton.toDB()`s `is_add` determined wrong

phorward:
This is a hotfix for the problem that `is_add` is being set based on if the key is set or not. Furthermore, it is determined twice, so that the "magic"-bullshit now was moved into the transaction. It is a fix for PR #973 but previous implementation was also wrong.

### DIFF
--- a/src/viur/core/skeleton.py
+++ b/src/viur/core/skeleton.py
@@ -949,7 +949,7 @@ class Skeleton(BaseSkeleton, metaclass=MetaSkel):
             logging.warning(msg, stacklevel=3)
             update_relations = not kwargs["clearUpdateTag"]
 
-        def txnUpdate(write_skel):
+        def __txn_update(write_skel):
             db_key = write_skel["key"]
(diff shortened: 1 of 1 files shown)


src/viur/core/skeleton.py
```diff
@@ -1166,23 +1172,15 @@ def fixDotNames(entity):
...
 
         # Run our SaveTxn
         if db.IsInTransaction():
-            key, db_obj, skel, change_list = txnUpdate(skel)
+            key, db_obj, skel, change_list, is_add = __txn_update(skel)
         else:
-            key, db_obj, skel, change_list = db.RunInTransaction(txnUpdate, skel)
-
-        # Perform post-save operations (postProcessSerializedData Hook, Searchindex, ..)
-        skel["key"] = key
```
phorward, line 1185: This is already done in __txn_update, therefore doppelt gemoppelt.

sveneberth (approved):

PR #1250, opened 2024-08-28
refactor: Send emails from `EmailTransport` instances instead of class

### LINKED ISSUE #813: `conf["viur.email.transport"]` with class instances for custom adapter configuration
> In general I would change the transport class to an instances.
> I think it is cleaner to configure an instance than the `conf` like this:
> ```py
> conf["viur.email.transport"] = EmailTransportSendInBlue(
> 	api_key="*************************",
> 	thresholds=(500, 300, 100),
> )
> ```

_Originally posted by @sveneberth in https://github.com/viur-framework/viur-core/issues/808#issuecomment-1675200774_

sveneberth:
This allows to store the credentials and other settings on the instance itself and reduce the config size.
Furthermore I think this is more pythonic, combines what belongs together and makes `abc` meaningful.

This a small **breaking change**.
Instead of 
```py
conf.email.mailjet_api_key = secret.get("api-mailjet-key")
conf.email.mailjet_api_secret = secret.get("api-mailjet-secret")
conf.email.transport_class = EmailTransportMailjet
```

you write now
```py
conf.email.transport_class = EmailTransportMailjet(
    api_key=secret.get("api-mailjet-key"),
    secret_key=secret.get("api-mailjet-secret"),
)
```

--- 
### Furthermore 
* Enforce pep8 compliant `snake_case` names as requested by @phorward 
* `email.sendEMail` is now `email.send_email`
* `email.sendEMailToAdmins` is now `email.send_email_to_admins`
* We use the [`deprecated`](https://deprecated.readthedocs.io/en/latest/) package, which is not working smooth as expected, but this can be done later ...
* Add and update a bunch of documentation
* Fixing not used `max_retries` variable
* Introduce some constants (so things can be changed with monkey patching or are clearly reserved and fixed names)

Resolves #813

### DIFF
--- a/docs/doc_tutorial/advanced/05_training_config.rst
+++ b/docs/doc_tutorial/advanced/05_training_config.rst
@@ -140,7 +140,7 @@ viur.emailRecipientOverride
 Override recipients for all outgoing email. This should be done for testing purposes only.
 
 If set, all outgoing emails will be send to this address
-(always overriding the ``dests``-parameter in `core.email.sendEMail`_).
+(always overriding the ``dests``-parameter in `core.email.send_email`_).
 
 ::
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,6 +7,7 @@ name = "viur-core"
 dynamic = ["version"]
 dependencies = [
     "appengine-python-standard~=1.0",
+    "Deprecated~=1.2",
     "google-api-core[grpc]~=2.0",
     "google-auth~=2.0",
     "google-cloud-datastore~=2.0",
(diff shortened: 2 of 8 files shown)


phorward (changes requested):
This pull request is in conflict with #1249, can you please integrate it here and fix the config first? IMHO we can make this into core-3.7.

Can you also generally rename the methods to become pep8-compliant, so that splitAddresses becomes split_addresses, etc.

sveneberth:
> This pull request is in conflict with #1249, can you please integrate it here and fix the config first

Yeah, I expected this. 

I will resolve the conflicts and do the renamings later.

sveneberth:
@phorward @ArneGudermann @KadirBalku I'm now done, you can review now :)

phorward (approved):

PR #1254, opened 2024-09-06
feat: `UriBone`

ArneGudermann:
Solution for https://github.com/viur-framework/viur-core/issues/1219

### DIFF
--- a/src/viur/core/bones/uri.py
+++ b/src/viur/core/bones/uri.py
@@ -27,16 +27,16 @@ class UriBone(BaseBone):
         **kwargs
     ):
         """
-        The URIBone is used for storing URI and URL.
+        The UriBone is used for storing URI and URL.
 
         :param accepted_protocols: The accepted protocols can be set to allow only the provide protocols.
         :param accepted_ports The accepted ports can be set to allow only the provide ports.
         ..  code-block:: python
             # Example
-            URIBone(accepted_ports=1)
-            URIBone(accepted_ports="2")
-            URIBone(accepted_ports="1-4")
-            URIBone(accepted_ports=(1,"2","4-10"))
+            UriBone(accepted_ports=1)
+            UriBone(accepted_ports="2")
+            UriBone(accepted_ports="1-4")
+            UriBone(accepted_ports=(1,"2","4-10"))
         :param clean_get_params: When set to True, the GET-parameter for the URL will be cleaned.
         :param domain_allowed_list: If set, only the URLs that are matched with an entry of this iterable
             will be accepted.
@@ -45,7 +45,7 @@ class UriBone(BaseBone):
         :param local_path_allowed: If True, the URLs that are local paths will be prefixed with "/".
         """
         super().__init__(**kwargs)
-        self.accepted_ports = set(sorted(URIBone._build_accepted_ports(accepted_ports), key=lambda rng: rng.start))
+        self.accepted_ports = set(sorted(UriBone._build_accepted_ports(accepted_ports), key=lambda rng: rng.start))
 
         if range(PORT_MIN, PORT_MAX + 1) in self.accepted_ports:
             self.accepted_ports = None  # all allowed
@@ -127,7 +127,7 @@ class UriBone(BaseBone):
         elif isinstance(accepted_ports, Iterable):
             accepted_ports_value = []
             for accepted_port in accepted_ports:
-                accepted_ports_value.extend(URIBone._build_accepted_ports(accepted_port))
+                accepted_ports_value.extend(UriBone._build_accepted_ports(accepted_port))
             return accepted_ports_value
 
         raise ValueError("accepted_ports must be a iterable or an integer or string")
--- a/src/viur/core/bones/uri.py
+++ b/src/viur/core/bones/uri.py
@@ -44,9 +44,11 @@ class URIBone(BaseBone):
         :param local_path_allowed: If True, the URLs that are local paths will be prefixed with "/".
         """
         super().__init__(**kwargs)
-        self.accepted_ports = sorted(URIBone._build_accepted_ports(accepted_ports), key=lambda rng: rng.start)
-        if self.accepted_ports == [range(PORT_MIN, PORT_MAX + 1)]:
+        self.accepted_ports = set(sorted(URIBone._build_accepted_ports(accepted_ports), key=lambda rng: rng.start))
+
+        if range(PORT_MIN, PORT_MAX + 1) in self.accepted_ports:
             self.accepted_ports = None  # all allowed
+
         self.accepted_protocols = accepted_protocols
         if self.accepted_protocols:
             if not isinstance(self.accepted_protocols, Iterable) or isinstance(self.accepted_protocols, str):
@@ -109,7 +111,7 @@ class URIBone(BaseBone):
                 return [range(start, end + 1)]
 
             else:
-                port = int(port)
+                port = int(accepted_ports)
                 return [range(port, port + 1)]
 
         elif isinstance(accepted_ports, int):
--- a/src/viur/core/bones/uri.py
+++ b/src/viur/core/bones/uri.py
@@ -1,10 +1,7 @@
 import fnmatch
-import sys
-
-import logging
+import typing as t
 from . import BaseBone, ReadFromClientError, ReadFromClientErrorSeverity
 from urllib.parse import urlparse, urlunparse
-import typing as t
 from collections.abc import Iterable
 from collections import namedtuple
 
@@ -45,10 +42,13 @@ class UriBone(BaseBone):
         :param local_path_allowed: If True, the URLs that are local paths will be prefixed with "/".
         """
         super().__init__(**kwargs)
-        self.accepted_ports = set(sorted(UriBone._build_accepted_ports(accepted_ports), key=lambda rng: rng.start))
+        if accepted_ports:
+            self.accepted_ports = set(sorted(UriBone._build_accepted_ports(accepted_ports), key=lambda rng: rng.start))
 
-        if range(PORT_MIN, PORT_MAX + 1) in self.accepted_ports:
-            self.accepted_ports = None  # all allowed
+            if range(PORT_MIN, PORT_MAX + 1) in self.accepted_ports:
+                self.accepted_ports = None  # all allowed
+        else:
+            self.accepted_ports = None
 
         self.accepted_protocols = accepted_protocols
         if self.accepted_protocols:
@@ -166,7 +166,7 @@ class UriBone(BaseBone):
 
     def singleValueFromClient(self, value, skel, bone_name, client_data) -> tuple:
         if err := self.isInvalid(value):
-            return self.getEmptyValue(), [ReadFromClientError(ReadFromClientErrorSeverity.Invalid, err)]
+            return value, [ReadFromClientError(ReadFromClientErrorSeverity.Invalid, err)]
 
         parsed_url = urlparse(value)
         if self.local_path_allowed and parsed_url.scheme == "":
--- a/src/viur/core/bones/uri.py
+++ b/src/viur/core/bones/uri.py
@@ -20,7 +20,8 @@ class URIBone(BaseBone):
             URIBone(accepted_ports=[1,"2","4-10"])
        :param clean_get_params: When set to Ture the GET Parameter for the URL will be cleand.
        :param domain_allowed_list: If set only the URLs that are match with an entry of this list will be accepted.
-       :param domain_disallowed_list: If set only the URLs that are not match with an entry of this list will be accepted.
+       :param domain_disallowed_list: If set only the URLs that are not match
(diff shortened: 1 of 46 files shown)


src/viur/core/bones/uri.py
```diff
@@ -0,0 +1,192 @@
...
+
+       :param accepted_protocols: The accepted protocols can be set to allow only the provide protocols.
+       :param accepted_ports The accepted ports can be set to allow only the provide ports.
+       ..  code-block:: python
+            # Example
+            URIBone(accepted_ports=1)
+            URIBone(accepted_ports="2")
+            URIBone(accepted_ports="1-4")
+            URIBone(accepted_ports=[1,"2","4-10"])
+       :param clean_get_params: When set to Ture the GET Parameter for the URL will be cleand.
```
phorward, line 21:
```suggestion
       :param clean_get_params: When set to True, the GET-parameter for the URL will be cleaned.
```

src/viur/core/bones/uri.py
```diff
@@ -0,0 +1,192 @@
...
+       The URIBone is used for storing URI and URL.
+
+       :param accepted_protocols: The accepted protocols can be set to allow only the provide protocols.
+       :param accepted_ports The accepted ports can be set to allow only the provide ports.
+       ..  code-block:: python
+            # Example
+            URIBone(accepted_ports=1)
+            URIBone(accepted_ports="2")
+            URIBone(accepted_ports="1-4")
+            URIBone(accepted_ports=[1,"2","4-10"])
```
phorward, line 20: Sorry, I'm a tuple fan
```suggestion
            URIBone(accepted_ports=(1,"2","4-10"))
```

src/viur/core/bones/uri.py
```diff
@@ -0,0 +1,192 @@
...
+       :param clean_get_params: When set to Ture the GET Parameter for the URL will be cleand.
+       :param domain_allowed_list: If set only the URLs that are match with an entry of this list will be accepted.
```
phorward, line 22:
```suggestion
       :param domain_allowed_list: If set, only the URLs that are matched with an entry of this iterable will be accepted.
```

src/viur/core/bones/uri.py
```diff
@@ -0,0 +1,192 @@
...
+       :param domain_disallowed_list: If set only the URLs that are not match
+            with an entry of this list will be accepted.
```
phorward, line 24:
```suggestion
       :param domain_disallowed_list: If set, only the URLs that are not matched
            with an entry of this iterable will be accepted.
```

src/viur/core/bones/uri.py
```diff
@@ -0,0 +1,192 @@
...
+       :param local_allowed_list: If True the URLs that are local paths will be prefixed with "/".
```
phorward, line 25:
```suggestion
       :param local_path_allowed: If True, the URLs that are local paths will be prefixed with "/".
```

src/viur/core/bones/uri.py
```diff
@@ -0,0 +1,192 @@
...
+
+        if not isinstance(local_path_allowed, bool):
+            raise ValueError("local_path_allowed must be a boolean")
+
+        self.clean_get_params = clean_get_params
+        self.domain_allowed_list = domain_allowed_list
+        self.domain_disallowed_list = domain_disallowed_list
+        self.local_path_allowed = local_path_allowed
+
+    def build_accepted_ports(self, accepted_ports):
```
phorward, line 80: This method should be a `@staticmethod` and additinally `_private`

src/viur/core/bones/uri.py
```diff
@@ -0,0 +1,192 @@
...
+        accepted_ports_value = set()
+        if isinstance(accepted_ports, str):
+            if accepted_ports == "*":
+                return "*"
```
phorward, line 84: This looks wrong, as your function returns a set on any other cases?

phorward (changes requested):
Hello @ArneGudermann, thanks for forcing this issue.

I think there are several issues which first need to be fixed. Can you take a look into my comments?

I do also see a problem with the ports set, which, in its worst case is a set from port 1 to 65535, which are 65535 Python int objects in one set. I had to deal with a similar problem when I developed a regular expressions library in C regarding unicode character-classes, for which I developed a special datatype which could handle ranges.

There is a C implementation https://github.com/phorward/libphorward/blob/master/src/ccl.c and also a Rust implementation https://github.com/phorward/charclass/blob/main/src/lib.rs.

Maybe there is some standard library thing in Python which also can handle this, you have to find out. But don't waste your time and develop it on your own, or you do it as a separate library if there is nothing similar.

Maybe @sveneberth can also help with his expertise here?

sveneberth (changes requested):
And

src/viur/core/bones/uri.py
```diff
@@ -0,0 +1,204 @@
+import fnmatch
+import sys
+
+import logging
+from . import BaseBone, ReadFromClientError, ReadFromClientErrorSeverity
+from urllib.parse import urlparse, urlunparse
+import typing as t
+from collections.abc import Iterable
+from collections import namedtuple
```
phorward, line 9: `logging` and `sys` are not used!
```suggestion
import fnmatch
import typing as t
from . import BaseBone, ReadFromClientError, ReadFromClientErrorSeverity
from urllib.parse import urlparse, urlunparse
from collections.abc import Iterable
from collections import namedtuple
```

src/viur/core/bones/uri.py
```diff
@@ -0,0 +1,204 @@
...
+
+            return [range(accepted_ports, accepted_ports + 1)]
+
+        elif isinstance(accepted_ports, Iterable):
+            accepted_ports_value = []
+            for accepted_port in accepted_ports:
+                accepted_ports_value.extend(UriBone._build_accepted_ports(accepted_port))
+            return accepted_ports_value
+
+        raise ValueError("accepted_ports must be a iterable or an integer or string")
```
phorward, line 133: `accepted_ports` defaults to `None`, so when not given, the instanciation fails with 
```
  File "~/myproject/deploy/skeletons/test.py", line 5, in <module>
    class TestSkel(skeleton.Skeleton):
  File "~/myproject/deploy/skeletons/test.py", line 17, in TestSkel
    url = UriBone(
          ^^^^^^^^
  File "~/mycore/src/viur/core/bones/uri.py", line 45, in __init__
    self.accepted_ports = set(sorted(UriBone._build_accepted_ports(accepted_ports), key=lambda rng: rng.start))
                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "~/mycore/src/viur/core/bones/uri.py", line 130, in _build_accepted_ports
    raise ValueError("accepted_ports must be a iterable or an integer or string")
ValueError: accepted_ports must be a iterable or an integer or string
```

src/viur/core/bones/uri.py
```diff
@@ -0,0 +1,204 @@
...
+            will be accepted.
+        :param domain_disallowed_list: If set, only the URLs that are not matched
+            with an entry of this iterable will be accepted.
+        :param local_path_allowed: If True, the URLs that are local paths will be prefixed with "/".
+        """
+        super().__init__(**kwargs)
+        self.accepted_ports = set(sorted(UriBone._build_accepted_ports(accepted_ports), key=lambda rng: rng.start))
+
+        if range(PORT_MIN, PORT_MAX + 1) in self.accepted_ports:
+            self.accepted_ports = None  # all allowed
```
phorward, line 51: Fix for ValueError, see below.

```suggestion
        if accepted_ports:
            self.accepted_ports = set(sorted(UriBone._build_accepted_ports(accepted_ports), key=lambda rng: rng.start))

            if range(PORT_MIN, PORT_MAX + 1) in self.accepted_ports:
                self.accepted_ports = None  # all allowed
        else:
            self.accepted_ports = None
```

src/viur/core/bones/uri.py
```diff
@@ -0,0 +1,204 @@
...
+                return f"""Url is not in the domain allowed list."""
+
+        if self.domain_disallowed_list is not None:
+            for domain in self.domain_disallowed_list:
+                if fnmatch.fnmatch(value, domain):
+                    return f"""Url is in the domain disallowed list."""
+
+    def singleValueFromClient(self, value, skel, bone_name, client_data) -> tuple:
+        if err := self.isInvalid(value):
+            return self.getEmptyValue(), [ReadFromClientError(ReadFromClientErrorSeverity.Invalid, err)]
```
phorward, line 169: Return `value` here, otherwise anything the user inputs is simply emptied, which is not user-friendly.
```suggestion
            return value, [ReadFromClientError(ReadFromClientErrorSeverity.Invalid, err)]
```

src/viur/core/bones/uri.py
```diff
@@ -0,0 +1,204 @@
...
+            for domain in self.domain_allowed_list:
+                if fnmatch.fnmatch(value, domain):
+                    break
+            else:
+                return f"""Url is not in the domain allowed list."""
+
+        if self.domain_disallowed_list is not None:
+            for domain in self.domain_disallowed_list:
+                if fnmatch.fnmatch(value, domain):
+                    return f"""Url is in the domain disallowed list."""
```
phorward, line 165: These tests are applied to the full value.

When I read `domain_allow_list` in the docstring and the parameter signature, I suggest that `["mausbrand.de"]` suffices for a domain allow list, like so:
```py
    url = UriBone(
        accepted_protocols=("https", "http"),
        domain_allowed_list=["mausbrand.de"],
    )
```
but instead, I must provide `["https://*.mausbrand.de/*", "https://mausbrand.de/*", "https://mausbrand.de", "http://*.mausbrand.de/*", "http://mausbrand.de/*", "http://mausbrand.de"]` to match all domains like subdomains, tld-only and all protocols (and this still doesn't match all cases).

This isn't cool. Maybe the bone should also accept regex here, but this is stage 2.

phorward (changes requested):

src/viur/core/bones/uri.py
ArneGudermann: You can provide it like 

    url = UriBone(
        accepted_protocols=("https", "http"),
        domain_allowed_list=["*mausbrand.de"],
    )
for all of this ?

phorward: > You can provide it like
> 
> ```python
>     url = UriBone(
>         accepted_protocols=("https", "http"),
>         domain_allowed_list=["*mausbrand.de"],
>     )
> ```
> 
> for all of this ?

Sure! But the protocol is not part of the domain, and shall not be handled as so. What if I want to accept only `mausbrand.de`? Then I must provide `*//mausbrand.de`? This notation is ulgy.

Oh right, and given `["*mausbrand.de"]`, I cannot provide a link `https://mausbrand.de/test` because then, the definition must be `["*mausbrand.de*"]`, but this also accepts `mausbrand.decoration` as accepted domain name.

Err no... please correct it so that the domain_allow/disalllow-lists are only matched against the true domain that is given, and not the entire value.

phorward (changes requested):

src/viur/core/bones/uri.py
```diff
@@ -45,10 +42,13 @@ def __init__(
         :param local_path_allowed: If True, the URLs that are local paths will be prefixed with "/".
         """
         super().__init__(**kwargs)
-        self.accepted_ports = set(sorted(UriBone._build_accepted_ports(accepted_ports), key=lambda rng: rng.start))
+        if accepted_ports:
+            self.accepted_ports = set(sorted(UriBone._build_accepted_ports(accepted_ports), key=lambda rng: rng.start))
```
sveneberth, line 46: The idea here was to have the ports sorted once to provide them in a friendly readable way for the structure.

```suggestion
            self.accepted_ports = sorted(set(UriBone._build_accepted_ports(accepted_ports)), key=lambda rng: rng.start)
```

src/viur/core/bones/uri.py
sveneberth: Can you implement some testcases based on this example?

tests/bones/test_uri_bone.py
```diff
@@ -0,0 +1,127 @@
...
+        url_value = "foo/bar/?a=b"
+        res = bone.singleValueFromClient(url_value, skel, self.bone_name, None)
+        self.assertEqual(("/foo/bar/?a=b", None), res)
+
+        bone = UriBone(local_path_allowed=False)
+        url_value = "/foo/bar/?a=b"
+        res = bone.singleValueFromClient(url_value, skel, self.bone_name, None)
+        self.is_invalid(res, url_value)
+
+        # Test for general schema, not valid URLs
```
sveneberth, line 102: I added some more tests for the schema, they seems to be identical to JS: https://jsfiddle.net/y03e26p5/5/

sveneberth (approved):

phorward (approved):

PR #1257, opened 2024-09-17
fix: @access-decorator

phorward:
- "root" is always allowed to pass
- the set().difference() method check didn't work as expected, therefore changed to an all(a in user["access"]) check.

### DIFF
--- a/src/viur/core/module.py
+++ b/src/viur/core/module.py
@@ -265,28 +265,30 @@ class Method:
 
                 raise errors.Unauthorized(self.access["message"]) if self.access["message"] else errors.Unauthorized()
 
-            ok = False
-            for acc in self.access["access"]:
-                if trace:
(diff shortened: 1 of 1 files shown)


ArneGudermann (approved):

src/viur/core/module.py
```diff
@@ -265,28 +265,30 @@ def parse_value_by_annotation(annotation: type, name: str, value: str | list | t
...
-                if isinstance(acc, str):
-                    acc = (acc, )
+                    # Otherwise, check for access rights
+                    if isinstance(acc, str):
+                        acc = (acc, )
 
-                assert isinstance(acc, (tuple, list, set))
+                    assert isinstance(acc, (tuple, list, set))
 
-                if not set(acc).difference(user["access"]):
```
sveneberth, line 287: @phorward  The `not` was too much.

Issue #1092, opened 2024-03-06
`compute` doesn't work in a `JsonBone`


sveneberth:
`compute` is implemented in `serialize` and `unserialize`, but the `JsonBone` overwrites this without a super call. So `bone._compute` is never called and the bone value is `None`.

ArneGudermann:
Solved by #1093

sveneberth:
Hey @ArneGudermann,

I tested it again and found a confusing behaivior.

I have this bone (has `row=True` by default):
```py
    price = JsonBone(  
        compute=Compute(lambda skel: skel.price_.to_dict(), ComputeInterval(ComputeMethod.Always))
    )
```

This will use `JsonBone.singleValueUnserialize` which tries to load a json string and fails (it's a dict):
```
[2024-09-03 16:49:37,983] /.../viur/core/request.py:332 [ERROR] the JSON object must be str, bytes or bytearray, not dict
Traceback (most recent call last):
  File "/.../viur-core/src/viur/core/request.py", line 306, in _process
    self._route(path)
  File "/.../viur-core/src/viur/core/request.py", line 565, in _route
    res = caller(*self.args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/.../viur-core/src/viur/core/module.py", line 299, in __call__
    return self._func(self._instance, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/.../viur-shop/src/viur/shop/modules/api.py", line 272, in cart_list
    child = self.json_renderer.renderSkelValues(child_skel)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/.../viur-core/src/viur/core/render/json/default.py", line 137, in renderSkelValues
    res[key] = self.renderBoneValue(bone, skel, key)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/.../viur-core/src/viur/core/render/json/default.py", line 102, in renderBoneValue
    boneVal = skel[key]
              ~~~~^^^^^
  File "/.../viur-core/src/viur/core/skeleton.py", line 226, in __getitem__
    boneInstance.unserialize(self, key)
  File "/.../viur-core/src/viur/core/bones/base.py", line 815, in unserialize
    skel.accessedValues[name] = self._compute(skel, name)
[...]

PR #1262, opened 2024-09-20
feat: Make SkeletonInstance json serializable

ArneGudermann:
This PR is a fix for #1092 and makes SkeletonInstances now json serializable.
Like:
```python
def get_skel():
	skel = conf.main_app.country.viewSkel()
	skel.fromDB("agxoaWVyLWdpYnQtZXNyFQsSDm5pY2h0c196dXNlaGVuGLkKDA")
	return skel

foo=JsonBone(compute=Compute(fn=get_skel,interval=ComputeInterval(ComputeMethod.OnWrite),raw=True))
```
Now get in json render 

```json
"c_s": {
               "key": "agxoaWVyLWdpYnQtZXNyFQsSDm5pY2h0c196dXNlaGVuGLkKDA",
                "name": null,
                "creationdate": "2022-10-05T06:20:38+00:00",
                "changedate": "2024-02-05T13:20:46+00:00",
                "viurCurrentSeoKeys": {
                    "de": "4757066702913536",
                    "_viurLanguageWrapper_": true,
                    "en": "4757066702913536"
                },
                "country": "Nauru",
                "continent": "Oceania",
                "sortindex": 1664950838.329367
            }
```

### DIFF
--- a/src/viur/core/bones/json.py
+++ b/src/viur/core/bones/json.py
@@ -49,7 +49,7 @@ class JsonBone(RawBone):
         if isinstance(val, (str, bytes, bytearray)):
             return utils.json.loads(val)
 
-        return val
+        return utils.json.loads(utils.json.dumps(val))
 
     def singleValueFromClient(self, value: str | list | dict, skel, bone_name, client_data):
         if value:
--- a/src/viur/core/bones/json.py
+++ b/src/viur/core/bones/json.py
@@ -46,7 +46,10 @@ class JsonBone(RawBone):
         return utils.json.dumps(skel.accessedValues[name])
 
     def singleValueUnserialize(self, val):
-        return utils.json.loads(val)
+        if isinstance(val, (str, bytes, bytearray)):
+            return utils.json.loads(val)
+
+        return utils.json.loads(utils.json.dumps(val))
 
     def singleValueFromClient(self, value: str | list | dict, skel, bone_name, client_data):
         if value:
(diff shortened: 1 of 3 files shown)


src/viur/core/bones/json.py
```diff
@@ -46,7 +46,10 @@ def singleValueSerialize(self, value, skel: 'SkeletonInstance', name: str, paren
         return utils.json.dumps(skel.accessedValues[name])
 
     def singleValueUnserialize(self, val):
-        return utils.json.loads(val)
+        if isinstance(val, (str, bytes, bytearray)):
+            return utils.json.loads(val)
+
+        return utils.json.loads(utils.json.dumps(val))
```
phorward, line 52: Hmm this looks a little crazy... doesn't just
```suggestion
        return val
```
work?

phorward (changes requested):

src/viur/core/utils/json.py
```diff
@@ -47,6 +47,9 @@ def preprocess(obj: t.Any) -> t.Any:
         elif isinstance(obj, (list, tuple)):
             return tuple(ViURJsonEncoder.preprocess(value) for value in obj)
 
+        elif hasattr(obj, "boneMap"):  # SkeletonInstance
```
phorward, line 50: I think this test is necessary because of circular imports. Can you give a hint in form of a comment. I wonder if there isn't a better method, like inspect.

src/viur/core/bones/json.py
phorward: Can you anyway give a short hint in form of a comment why it must be done this way?

phorward (changes requested):
Please clarify the code-snippets in the comments in the source code.

phorward (approved):

PR #1264, opened 2024-09-20
refactor: `Skeleton`-API rework

phorward:
- [x] Rename `Skeleton.toDB()` into `Skeleton.write()`
- [x] Rename `Skeleton.fromDB()` into `Skeleton.read()`
- [x] Rename any `skelValues` variable into `skel`
- [x] Clean-up `Skeleton.delete()`
- [x] Improve `Skeleton.read()`, `Skeleton.write()` and `Skeleton.delete()` to accept key-parameter
- [x] Deprecation handling
- [x] Added support for legacy fromDB/toDB definitions in project-specific Skeletons

This pull request is a first, backward-compatible milestone on #630.

Issue #991, opened 2023-12-12
Context-manager for transactions with Skeletons `with MySkel().transact(): ...`


phorward:
It would be nice to do something like this:

```python
with self.editSkel().transact(key) as skel:
    skel["counter"] += 1
```

Currently, it is impossible, as `viur-datastore` needs to be enhanced.

phorward:
Relates to #853

phorward:
Hello @ArneGudermann, thanks for taking a closer look on this.

IMHO this could be achieved by modifying and/or modularization of code in viur-datastore, around here: https://github.com/viur-framework/viur-datastore/blob/master/src/viur/datastore/transport.pyx#L806
There is a beginTransaction and a commit, which must be turned into the ContextManager object.

Nevertheless, I don't want to heavily change viur-datastore for features in viur-core, as we still plan to integrate other database adapters in far future.

As an alternative for this issue, this is my most-current version of the `set_status` function I'm using throughout several projects:
```py
"""
This module contains some project-specific helper functions.
"""

import logging
import time
from viur.core import db, errors


def set_status(key, values=None, check=None, func=None, skel=None, retry=3, update_relations=False):
    """
    Universal function to set a status of an db.Entity or Skeleton within a transaction.

    :param key: Entity/skeleton key to change
    :param values: A dict of key-values to update on the entry
    :param check: An optional dict of key-values to check on the entry before
    :param func: A function that is called inside the transaction
    :param skel: Use assigned skeleton instead of low-level DB-API
    :param retry: On BadRequestError, retry for this amount of times.
    :param update_relations: Instruct ViUR to update relations (normally not required)

    If the function does not raise an Exception, all went well.
    It returns either the assigned skel, or the db.Entity on success.
    """
    if callable(values):
        assert not func, "'values' is a callable, but func is also set. Either set values or func in this case."
        func = values
        values = None
    else:
        assert isinstance(values, dict) or values is None, "'values' has to be a dict when set"

    def transaction():
        if skel:
            if not skel.fromDB(key):
                raise errors.NotFound()

            obj = skel
        else:
            obj = db.Get(key)

        if check:
            if callable(check):
                assert check(obj)
            else:
                assert isinstance(check, dict), "'check' has to be a dict, you diggi!"

                for bone, value in check.items():
                    assert obj[bone] == value, "%r contains %r, expecting %r" % (bone, obj[bone], value)

        if values:
            for bone, value in values.items():
                if bone[0] == "+":
[...]

PR #1267, opened 2024-09-24
feat: `Skeleton.patch()` for transactional read/write

phorward:
Provides a `Skeleton.patch()` function based on `viur.toolkit.db.set_status()`.

This improves the Skeleton API to a better and more integrated CRUD-API (Create-Read-Update-Delete). It also introduces a ReadFromClientException that can be raised internally for client parsing error reporting.

This pull request is part of #630 and a follow-up on #1264, which ~should be~ was merged first.

This is also an alternative replacement on #991.

### DIFF
--- a/src/viur/core/skeleton.py
+++ b/src/viur/core/skeleton.py
@@ -7,12 +7,12 @@ import logging
 import os
 import string
 import sys
+import time
 import typing as t
 import warnings
 from deprecated.sphinx import deprecated
 from functools import partial
 from itertools import chain
-from time import time
 from viur.core import conf, current, db, email, errors, translate, utils
 from viur.core.bones import (
     BaseBone,
@@ -76,6 +76,7 @@ class MetaBaseSkel(type):
         "style",
         "toDB",
         "unserialize",
+        "update",
         "values",
         "write",
     }
@@ -199,9 +200,6 @@ class SkeletonInstance:
 
         return self[item]
 
-    def update(self, *args, **kwargs) -> None:
-        self.__ior__(dict(*args, **kwargs))
-
     def __setitem__(self, key, value):
         assert self.renderPreparation is None, "Cannot modify values while rendering"
         if isinstance(value, BaseBone):
@@ -270,6 +268,7 @@ class SkeletonInstance:
             "setBoneValue",
             "toDB",
             "unserialize",
+            "update",
             "write",
         }:
             return partial(getattr(self.skeletonCls, item), self)
@@ -1244,7 +1243,7 @@ class Skeleton(BaseSkeleton, metaclass=MetaSkel):
             skel.dbEntity["viur"]["viurLastRequestedSeoKeys"] = current_seo_keys
 
             # mark entity as "dirty" when update_relations is set, to zero otherwise.
-            skel.dbEntity["viur"]["delayedUpdateTag"] = time() if update_relations else 0
+            skel.dbEntity["viur"]["delayedUpdateTag"] = time.time() if update_relations else 0
 
             skel.dbEntity = skel.preProcessSerializedData(skel.dbEntity)
 
@@ -1326,9 +1325,9 @@ class Skeleton(BaseSkeleton, metaclass=MetaSkel):
         if update_relations and not is_add:
             if change_list and len(change_list) < 5:  # Only a few bones have changed, process these individually
                 for idx, changed_bone in enumerate(change_list):
-                    updateRelations(key, time() + 1, changed_bone, _countdown=10 * idx)
+                    updateRelations(key, time.time() + 1, changed_bone, _countdown=10 * idx)
             else:  # Update all inbound relations, regardless of which bones they mirror
-                updateRelations(key, time() + 1, None)
+                updateRelations(key, time.time() + 1, None)
 
         # Trigger the database adapter of the changes made to the entry
         for adapter in skel.database_adapters:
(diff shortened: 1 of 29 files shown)


src/viur/core/skeleton.py
```diff
@@ -1422,6 +1425,117 @@ def __txn_delete(skel: SkeletonInstance, key: db.Key):
...
 
+    @classmethod
+    def edit(
+        cls,
+        skel: SkeletonInstance,
+        values: t.Optional[dict | t.Callable[[SkeletonInstance], None]] = {},
+        *,
+        key: t.Optional[db.Key | int | str] = None,
+        check: t.Optional[dict | t.Callable[[SkeletonInstance], None]] = None,
+        create: t.Optional[dict | t.Callable[[SkeletonInstance], None]] = None,
```
sveneberth, line 1436:
```suggestion
        create: t.Optional[bool | dict | t.Callable[[SkeletonInstance], None]] = None,
```

src/viur/core/skeleton.py
```diff
@@ -1422,6 +1425,117 @@ def __txn_delete(skel: SkeletonInstance, key: db.Key):
...
+            elif callable(check):
+                check(skel)
+
+            # Set values
+            if isinstance(values, dict):
+                if values and not skel.fromClient(values, amend=True):
+                    raise ReadFromClientException(skel.errors)
+
+                # Special-feature: "+" and "-" prefix for simple calculations
+                # FIXME: This can maybe integrated into skel.fromClient() later...
```
sveneberth, line 1506: That would be a feature and is not a bug, isn't it?
```suggestion
                # TODO: This can maybe integrated into skel.fromClient() later...
```

src/viur/core/skeleton.py
```diff
@@ -1422,6 +1425,117 @@ def __txn_delete(skel: SkeletonInstance, key: db.Key):
         for adapter in skel.database_adapters:
             adapter.delete(skel)
 
+    @classmethod
+    def edit(
```
sveneberth, line 1429: I don't like the name. `edit` is already an action in the prototypes. But it has nothing to do with this `edit` method here. To clearly separate them, I suggest calling it [`upsert`](https://en.wiktionary.org/wiki/upsert), as in the context of relational DBs.


```suggestion
    def upsert(
```

sveneberth (changes requested):

src/viur/core/skeleton.py
phorward: I don't like the name `upsert` either. Never heard this before.
I see the `edit` in context of *editing* a Skeleton.
You may apply other alternative names.

sveneberth: Hmm, okay. Let's brainstorm together in the viur-meeting on Friday. If nobody has a better name, we can merge this as `edit`.

phorward: What about `Skeleton.apply()`?

src/viur/core/skeleton.py
```diff
@@ -1422,6 +1425,117 @@ def __txn_delete(skel: SkeletonInstance, key: db.Key):
...
+        def __update_txn():
+            # Try to read the skeleton, create on demand
+            if not skel.read(key):
+                if not (key or skel["key"]):
+                    return ValueError("No valid key provided")
+
+                skel["key"] = db.keyHelper(key or skel["key"], skel.kindName)
+
+                if create is None or create is False:
+                    raise ValueError("Creation during update is forbidden - explicitly provide `create=True` to allow.")
```
sveneberth, line 1562: I called this `edit` method on an empty method to test the create part, but got `ValueError('No valid key provided')`. I think claiming automatically a not provided key should be supported as well. 

```suggestion
                if create is None or create is False:
                    raise ValueError("Creation during update is forbidden - explicitly provide `create=True` to allow.")
                
                if not (key or skel["key"]) and create in (False, None):
                    return ValueError("No valid key provided")

                skel["key"] = db.keyHelper(key or skel["key"], skel.kindName)
```

src/viur/core/skeleton.py
```diff
@@ -1422,6 +1425,117 @@ def __txn_delete(skel: SkeletonInstance, key: db.Key):
...
+        Raises:
+            ValueError: In case parameters where given wrong or incomplete.
+            AssertionError: In case an asserted check parameter did not match.
+            ReadFromClientException: In case a skel.fromClient() failed with a high severity.
+        """
+
+        # Transactional function
+        def __update_txn():
+            # Try to read the skeleton, create on demand
+            if not skel.read(key):
```
sveneberth, line 1555:
```suggestion
            if not skel.read(key or skel["key"]):
```

sveneberth (changes requested):

src/viur/core/skeleton.py
sveneberth: _apply_ sounds vague in this context. i would expect apply after a sequence of operations to finally apply them to an object.

Further options for our brainstorming / discussion could be:

- change
- modify
- alter

phorward: Let's name it `patch`?

src/viur/core/skeleton.py
```diff
@@ -1503,6 +1506,117 @@ def __txn_delete(skel: SkeletonInstance, key: db.Key):
         for adapter in skel.database_adapters:
             adapter.delete(skel)
 
+    @classmethod
+    def edit(
```
phorward, line 1510:
```suggestion
    def patch(
```

src/viur/core/skeleton.py
```diff
@@ -1503,6 +1506,117 @@ def __txn_delete(skel: SkeletonInstance, key: db.Key):
...
+        def __update_txn():
+            # Try to read the skeleton, create on demand
+            if not skel.read(key):
+                if not (key or skel["key"]):
+                    raise ValueError("No valid key provided")
+
+                skel["key"] = db.keyHelper(key or skel["key"], skel.kindName)
+
+                if create is None or create is False:
+                    raise ValueError("Creation during update is forbidden - explicitly provide `create=True` to allow.")
```
phorward, line 1562: Previously provided by @sveneberth 
```suggestion
                if create is None or create is False:
                    raise ValueError("Creation during update is forbidden - explicitly provide `create=True` to allow.")
                
                if not (key or skel["key"]) and create in (False, None):
                    return ValueError("No valid key provided")
                
                if key or skel["key"]:
                    skel["key"] = db.keyHelper(key or skel["key"], skel.kindName)
```

src/viur/core/skeleton.py
```diff
@@ -54,6 +55,7 @@ class MetaBaseSkel(type):
         "clone",
         "cursor",
         "delete",
+        "edit",
```
ArneGudermann, line 58:
```suggestion
        "patch",
```

src/viur/core/skeleton.py
```diff
@@ -277,6 +279,7 @@ def __getattr__(self, item: str):
         elif item in {
             "all",
             "delete",
+            "edit",
```
ArneGudermann, line 282:
```suggestion
            "patch",
```

src/viur/core/skeleton.py
```diff
@@ -1503,6 +1506,118 @@ def __txn_delete(skel: SkeletonInstance, key: db.Key):
...
+                if key or skel["key"]:
+                    skel["key"] = db.keyHelper(key or skel["key"], skel.kindName)
+
+                if isinstance(create, dict):
+                    if create and not skel.fromClient(create, amend=True):
+                        raise ReadFromClientException(skel.errors)
+                elif callable(create):
+                    create(skel)
+                elif create is not True:
+                    raise ValueError("'create' must either be dict or a callable.")
```
ArneGudermann, line 1571:
```suggestion
                else:
                    raise ValueError("'create' must either be dict or a callable.")
```
This not raise if `create` is True

ArneGudermann (changes requested):

src/viur/core/skeleton.py
phorward: No, it can either be `True`, `dict` or `callable`. Anything else shall raise the Exception.

sveneberth (approved):

PR #1277, opened 2024-10-02
feat: Load Session only when needed

### LINKED ISSUE #1276: Set ViUR session cookie only when necessary
There are considerations about providing a ViUR system that does not rely on a cookie that is always set.

An idea of this consideration and little discussion was, that ViUR should only set a session cookie in case something is written into the session. If this was once the case, the session should be kept, even if its content is being cleared.

This would allow to provide ViUR systems that don't use cookies, and only sets the cookie when really necessary.

ArneGudermann:
This PR fix #1276.

1. Language is only stored in the Session when a Session exist. 
2. When no Session exists the `Accept-Language` Header is used to determine the language.
3. `clear`, `popitem`, `setdefault` added to the Session class
4. The Session is only fetched from db if a Cookie is provided

### DIFF
--- a/src/viur/core/config.py
+++ b/src/viur/core/config.py
@@ -596,7 +596,7 @@ class User(ConfigType):
     session_persistent_fields_on_login: Multiple[str] = ["language"]
     """If set, these Fields will survive the session.reset() called on user/login"""
 
-    session_persistent_fields_on_logout: Multiple[str] = []
+    session_persistent_fields_on_logout: Multiple[str] = ["language"]
     """If set, these Fields will survive the session.reset() called on user/logout"""
 
     max_password_length: int = 512
--- a/src/viur/core/config.py
+++ b/src/viur/core/config.py
@@ -544,11 +544,12 @@ class I18N(ConfigType):
     language_alias_map: dict[str, str] = {}
     """Allows mapping of certain languages to one translation (i.e. us->en)"""
 
-    language_method: t.Literal["session", "url", "domain"] = "session"
+    language_method: t.Literal["session", "url", "domain","header"] = "session"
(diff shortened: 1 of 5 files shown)


phorward (changes requested):
Hello @ArneGudermann, thanks for opening this pull request, and take a look into my suggestions.

It looks already promising. Beside my comments, I found out during little tests that the cookie is not set until I logged in. When logged it, the cookie is set, which is expected. Then, when I logout again, the cookie is set with a new cookie-key, so a user who was logged in but is now logged out again has a separate session by which he can be identified.

Can you respect this case in your pull request, too?

src/viur/core/config.py
```diff
@@ -593,7 +594,7 @@ class User(ConfigType):
     session_persistent_fields_on_login: Multiple[str] = ["language"]
     """If set, these Fields will survive the session.reset() called on user/login"""
 
-    session_persistent_fields_on_logout: Multiple[str] = ["language"]
+    session_persistent_fields_on_logout: Multiple[str] = []
```
phorward, line 597: Did this change in accordance with @ArneGudermann to leave an empty (non-cookie-requiring) session after user logged out.

phorward (dismissed):
Well, for me it works so far as expected.
I think that more tests are needed with greater projects afterwards.

src/viur/core/config.py
```diff
@@ -595,7 +596,7 @@ class User(ConfigType):
...
```
phorward, line 599:
```suggestion
    session_persistent_fields_on_logout: Multiple[str] = ["language"]
```

phorward (approved):

PR #1292, opened 2024-10-22
fix: Test `user["status"]` at one place

### LINKED ISSUE #1082: chore: `self._user_module.render.login` gets `loginFailed` and `accountStatus` parameters - why?
This is surely a little stupid and IMHO not necessary anymore.

> But this is stupid, since you can the bone from the skeleton (which is passed).

_Originally posted by @sveneberth in https://github.com/viur-framework/viur-core/pull/997#discussion_r1502598382_ and https://github.com/viur-framework/viur-core/pull/997#discussion_r1502596979_

phorward:
This change avoids special-cases as described in #1082 and also fixes a little bug in the password recovery process. Closes #1082.

### DIFF
--- a/src/viur/core/modules/user.py
+++ b/src/viur/core/modules/user.py
@@ -336,18 +336,9 @@ class UserPassword(UserPrimaryAuthentication):
         # next, check if the password hash matches
         is_okay &= secrets.compare_digest(password_data.get("pwhash", b"-invalid-"), password_hash)
 
-        # next, check if the user account is active
-        is_okay &= (user_skel["status"] or 0) >= Status.ACTIVE.value
-
         if not is_okay:
             self.loginRateLimit.decrementQuota()  # Only failed login attempts will count to the quota
-            skel = self.LoginSkel()
-            return self._user_module.render.login(
-                skel,
-                action="login",
-                loginFailed=True,  # FIXME: Is this still being used?
-                accountStatus=user_skel["status"]  # FIXME: Is this still being used?
-            )
+            return self._user_module.render.login(self.LoginSkel(), action="login")
 
         # check if iterations are below current security standards, and update if necessary.
         if iterations < PBKDF2_DEFAULT_ITERATIONS:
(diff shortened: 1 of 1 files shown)


src/viur/core/modules/user.py
```diff
@@ -336,18 +336,9 @@ def login(self, *, name: str | None = None, password: str | None = None, **kwarg
         # next, check if the password hash matches
         is_okay &= secrets.compare_digest(password_data.get("pwhash", b"-invalid-"), password_hash)
 
-        # next, check if the user account is active
-        is_okay &= (user_skel["status"] or 0) >= Status.ACTIVE.value
```
sveneberth, line 340: But where will the status check at login be in future? I only see the one at pwrecover and this one has been removed.
It should then either be moved to `next_or_finish` or remain here.

src/viur/core/modules/user.py
```diff
@@ -450,7 +441,7 @@ def pwrecover(self, recovery_key: str | None = None, skey: str | None = None, *a
                 )
             )
 
-        if user_skel["status"] != Status.ACTIVE:  # The account is locked or not yet validated. Abort the process.
+        if user_skel["status"] < Status.ACTIVE:  # The account is locked or not yet validated. Abort the process.
```
sveneberth, line 444: We have never specified anywhere which value ranges mean what. As far as I'm OK with that, it can be checked that way. But then there should be an explanation somewhere.

sveneberth (changes requested):

src/viur/core/modules/user.py
phorward: Hi @sveneberth, regardless of the authentication method, the user's status is always checked in [`User.authenticateUser()`](https://github.com/viur-framework/viur-core/blob/develop/src/viur/core/modules/user.py#L1397).

This check here is redundant now, and a relict of the first ViUR version.

sveneberth: Okay, I don't know why I overlooked `authenticateUser` earlier ...

sveneberth (approved):

PR #1295, opened 2024-10-23
fix: Handle `RefSkel`s in `unserialize_compute` differently

ArneGudermann:
This PR fix the following bug:

If you have a relation in skeleton A to a skeleton B that has a compute bone with a lifetime. If you then call the skeleton A but the lifetime of the bone from the skeleton B is not included in the call because it is not in the RefKeys.

### DIFF
--- a/src/viur/core/bones/base.py
+++ b/src/viur/core/bones/base.py
@@ -889,7 +889,7 @@ class BaseBone(object):
             skel.accessedValues[name] = self.getDefaultValue(skel)
             return False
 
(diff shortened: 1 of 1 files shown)


phorward (changes requested):
It seems that `loaded_value` was an unused parameter. Please do also remove it from the docstring then (generally search for it if there are more occurrences)

phorward (approved):

sveneberth (approved):

PR #1305, opened 2024-10-28
feat: Add `File.get_download_url()`

ArneGudermann:


src/viur/core/modules/file.py
```diff
@@ -1161,6 +1161,56 @@ def add(self, skelType: SkelType, node: db.Key | int | str | None = None, *args,
...
+
+        if not self.canView("leaf", skel):
+            raise errors.Unauthorized()
+
+        dlkey = skel["dlkey"]
+
+        if derived and filename is None:
+            res = {}
+            for filename in skel["derived"]["files"]:
+                res[filename] = {"downloadUrl": self.create_download_url(dlkey, filename, derived)}
```
phorward, line 1198:
```suggestion
                res[filename] = self.create_download_url(dlkey, filename, derived)
```

src/viur/core/modules/file.py
```diff
@@ -1161,6 +1161,56 @@ def add(self, skelType: SkelType, node: db.Key | int | str | None = None, *args,
...
+                # Check if Filename exist in the Derives. We sign nothing that not exist.
+                if filename not in skel["derived"]["files"]:
+                    raise errors.NotFound("File not in derives")
+            else:
+                if filename is None:
+                    filename = skel["name"]
+                elif filename != skel["name"]:
+                    raise errors.NotFound("Filename not match")
+
+            res = {"downloadUrl": self.create_download_url(dlkey, filename, derived)}
```
phorward, line 1210:
```suggestion
            res = self.create_download_url(dlkey, filename, derived)
```

phorward (changes requested):
@ArneGudermann looks much better. Why this `{"downloadUrl": ...}`-thing everywhere, can we just use the url or the url per filename as suggested?

phorward (approved):
I'm fine with it!

sveneberth (approved):

PR #1313, opened 2024-11-06
feat: Support enum type

sveneberth:
People came up with the idea of using the available parameters ... Never noticed so far
https://github.com/viur-framework/viur-shop/blob/ff6ea1721c7e09f4c3a9cd3346541f071884c2fb/src/viur/shop/modules/api.py#L176

Please patch this in 3.7 and later as well.

### DIFF
--- a/src/viur/core/module.py
+++ b/src/viur/core/module.py
@@ -82,7 +82,7 @@ class Method:
             Tries to parse a value according to a given type.
             May be called recursively to handle unions, lists and tuples as well.
             """
-            logging.debug(f"{annotation=} | {name=} | {value=}")
+            # logging.debug(f"{annotation=} | {name=} | {value=}")
 
             # simple types
             if annotation is str:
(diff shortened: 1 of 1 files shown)


src/viur/core/module.py
```diff
@@ -81,6 +82,8 @@ def parse_value_by_annotation(annotation: type, name: str, value: str | list | t
             Tries to parse a value according to a given type.
             May be called recursively to handle unions, lists and tuples as well.
             """
+            # logging.debug(f"{annotation=} | {name=} | {value=}")
+
```
ArneGudermann, line 86:
```suggestion
```
I think we dont't need this.

ArneGudermann (changes requested):

src/viur/core/module.py
sveneberth: I don't care, I find it quite practical if you have to debug something and only have to comment it in. But it can also go away. Let @phorward  decide.

phorward (approved):

ArneGudermann (approved):

Issue #1035, opened 2024-01-22
Rename `server.modules.user.accessright.` translation key


phorward:
> Hmm this looks awkward, as "server." is referred.

_Originally posted by @phorward in https://github.com/viur-framework/viur-core/pull/1025#discussion_r1456106198_

PR #1315, opened 2024-11-08
ref: Replace `server` in Translations

ArneGudermann:
Fix for: #1035


phorward (changes requested):
@ArneGudermann can you do this PR against branch 3.6 please?

PR #1317, opened 2024-11-12
fix: Name `f"server.modules.user.accessright.{right}"` correctly

ArneGudermann:
Fix for: https://github.com/viur-framework/viur-core/issues/1035 and Replacement for #1315

### DIFF
--- a/src/viur/core/modules/moduleconf.py
+++ b/src/viur/core/modules/moduleconf.py
@@ -49,7 +49,7 @@ class ModuleConfScriptSkel(skeleton.RelSkel):
     access = SelectBone(
         descr="Required access rights",
         values=lambda: {
-            right: i18n.translate(f"server.modules.user.accessright.{right}", defaultText=right)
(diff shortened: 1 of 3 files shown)


sveneberth (changes requested):
I would this name `viur.core.*` since ViUR is ecosystem and has several components

ArneGudermann:
> I would name this `viur.core.*` since ViUR is ecosystem and has several components

I have taken an example from https://github.com/search?q=repo%3Aviur-framework%2Fviur-core%20%20viur.modules.user&type=code

phorward:
@ArneGudermann @sveneberth how important is this now? Is it a required feature? Is it really necessary for 3.6? Please agree.

sveneberth:
> @ArneGudermann @sveneberth how important is this now? Is it a required feature? Is it really necessary for 3.6? Please agree.

No, not necessary for me, neither in 3.6 nor in 3.7. And I don't care what we call it. 

@ArneGudermann simply works only on prio low tickets ...

phorward (approved):

PR #1330, opened 2024-11-21
feat: Provide `ignore`-parameter for `Skeleton.fromClient`

phorward:
- Provides a way to parse data into readonly bones as well
- Default for `Skeleton.patch()`'ed skeletons is to also allow to write into readonly bones


sveneberth (approved):
But is this `ignore` now _trusted_ or _untrusted_?

PR #1331, opened 2024-11-22
fix: `User.is_active()` without status-bone

phorward:
When the UserSkel has no status-bone, it should default to is_active=True.

It occured in a subskel, that didn't contain the "status"-bone, the default
```py
    def onEdited(self, skel):
        super().onEdited(skel)
        # In case the user is set to inactive, kill all sessions
        if not self.is_active(skel):
            session.killSessionByUser(skel["key"])
```

always kicks the user - just because the "status" bone was unavailable. The previous definition before #1309 was
```py
    def onEdited(self, skel):
        super().onEdited(skel)
        # In case the user is set to inactive, kill all sessions
        if "status" in skel and skel["status"] < Status.ACTIVE.value:
            session.killSessionByUser(skel["key"])
```

### DIFF
--- a/src/viur/core/modules/user.py
+++ b/src/viur/core/modules/user.py
@@ -439,7 +439,7 @@ class UserPassword(UserPrimaryAuthentication):
             )
 
         # If the account is locked or not yet validated, abort the process.
-        if not self._user_module.is_active(user_skel):
+        if self._user_module.is_active(user_skel) is False:
             raise errors.NotFound(
                 i18n.translate(
                     key="viur.modules.user.passwordrecovery.accountlocked",
@@ -1377,7 +1377,7 @@ class User(List):
 
         return self.authenticateUser(user_key)
 
-    def is_active(self, skel: skeleton.SkeletonInstance) -> bool:
+    def is_active(self, skel: skeleton.SkeletonInstance) -> bool | None:
         """
         Hookable check if a user is defined as "active" and can login.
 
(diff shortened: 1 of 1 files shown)


ArneGudermann (dismissed):

sveneberth (changes requested):
I see your use case, but I don't actually find it right to return `True` neither `False`. The state here is simply indeterminable. I would return a `None` here and adjust the if accordingly

src/viur/core/modules/user.py
```diff
@@ -439,7 +439,7 @@ def pwrecover(self, recovery_key: str | None = None, skey: str | None = None, *a
             )
 
         # If the account is locked or not yet validated, abort the process.
-        if not self._user_module.is_active(user_skel):
+        if self._user_module.is_active(user_skel) is False:
```
sveneberth, line 442: In this case I would keep the old behaviour. It's necessary to include the status in the viewSkel, otherwise you can't determine if it's allowed and it should fail.

src/viur/core/modules/user.py
```diff
@@ -1409,7 +1409,7 @@ def authenticateUser(self, key: db.Key, **kwargs):
             raise ValueError(f"Unable to authenticate unknown user {key}")
 
         # Verify that this user account is active
-        if not self.is_active(skel):
+        if self.is_active(skel) is False:
```
sveneberth, line 1412: In this case I would keep the old behaviour. It's necessary to include the status in the viewSkel, otherwise you can't determine if it's allowed and it should fail.

sveneberth (changes requested):

sveneberth (approved):

PR #1332, opened 2024-11-22
fix: Revert changes from #1323

sveneberth:
Ensure key is a str for every usage

Also here:
https://github.com/viur-framework/viur-core/blob/ec2c09e29adaedd490a2b7032590c6a91be78a3e/src/viur/core/i18n.py#L185

### DIFF
--- a/src/viur/core/i18n.py
+++ b/src/viur/core/i18n.py
@@ -180,6 +180,8 @@ class translate:
         :param public: Flag for public translations, which can be obtained via /json/_translate/get_public.
         """
         super().__init__()
+        if not isinstance(key, str):
+            logging.warning(f"Got non-string (type {type(key)}) as {key=}!", exc_info=True)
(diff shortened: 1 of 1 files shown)


src/viur/core/i18n.py
```diff
@@ -180,8 +180,8 @@ def __init__(
         :param public: Flag for public translations, which can be obtained via /json/_translate/get_public.
         """
         super().__init__()
-
-        self.key = str(key).lower()
+        key = str(key)  # ensure key is a str
```
sveneberth, line 183: maybe we can add this warning
```suggestion
        if not isinstance(key, str):
            logging.warning(f"Got non-string (type {type(key)}) as {key=}!")
        key = str(key)  # ensure key is a str
```

phorward: yes please

phorward (approved):

PR #1333, opened 2024-11-22
fix: Store written db_obj in `toDB` on source skel

sveneberth:
If you create a new skel, call `toDB` and then access `skel.dbEntity` it was still None, this made problems with compute bones


phorward (approved):
I think this is completely rewritten in core 3.7, and I got over this, too.

PR #1341, opened 2024-12-04
fix+refactor: html-render `getSkel()` and `getList()`

phorward:
This is a general refactoring and some issue fixing with ViURs ugly and ridden-through jinja functions `getSkel()` and `getList()`.

Fixes bug #1343

### DIFF
--- a/src/viur/core/render/html/env/viur.py
+++ b/src/viur/core/render/html/env/viur.py
@@ -177,53 +177,19 @@ def getSkel(
     elif not key:
         raise ValueError("getSkel has to be called with a valid key!")
 
-<<<<<<< Updated upstream
-        if "canView" in dir(obj):
-            if not skel.fromDB(key):
-                logging.info(f"getSkel: Entry {key} not found")
-                return None
-            if isinstance(obj, prototypes.singleton.Singleton):
-                isAllowed = obj.canView()
-            elif isinstance(obj, prototypes.tree.Tree):
-                k = db.Key(key)
-                if k.kind.endswith("_rootNode"):
-                    isAllowed = obj.canView("node", skel)
-                else:
-                    isAllowed = obj.canView("leaf", skel)
-            else:  # List and Hierarchies
-                isAllowed = obj.canView(skel)
-            if not isAllowed:
-                logging.error(f"getSkel: Access to {key} denied from canView")
-                return None
-        elif "listFilter" in dir(obj):
-            qry = skel.all().mergeExternalFilter({"key": str(key)})
-            qry = obj.listFilter(qry)
-            if not qry:
-                logging.info("listFilter permits getting entry, returning None")
-                return None
-=======
     if "canView" in dir(obj):
         if not skel.read(key):
             logging.info(f"getSkel: Entry {key!r} not found")
             return None
->>>>>>> Stashed changes
 
         if isinstance(obj, prototypes.singleton.Singleton):
             isAllowed = obj.canView()
 
-<<<<<<< Updated upstream
-        else:  # No Access-Test for this module
-            if not skel.fromDB(key):
-                return None
-        skel.renderPreparation = render.renderBoneValue
-        return skel
-=======
         elif isinstance(obj, prototypes.tree.Tree):
             if skel["key"].kind == obj.nodeSkelCls.kindName:
                 isAllowed = obj.canView("node", skel)
             else:
                 isAllowed = obj.canView("leaf", skel)
->>>>>>> Stashed changes
 
         else:  # List and Hierarchies
             isAllowed = obj.canView(skel)
--- a/src/viur/core/render/html/env/viur.py
+++ b/src/viur/core/render/html/env/viur.py
@@ -191,7 +191,7 @@ def getSkel(
             else:
                 isAllowed = obj.canView("leaf", skel)
 
-        else:  # List and Hierarchies
+        else:
             isAllowed = obj.canView(skel)
 
         if not isAllowed:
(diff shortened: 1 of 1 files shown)


src/viur/core/render/html/env/viur.py
```diff
@@ -150,63 +156,66 @@ def getSkel(render: Render, module: str, key: str = None, skel: str = "viewSkel"
...
-                return None
-        skel.renderPreparation = render.renderBoneValue
-        return skel
+    if isinstance(obj, prototypes.singleton.Singleton) and not key:
+        # We fetching the entry from a singleton - No key needed
+        key = db.Key(skel.kindName, obj.getKey())
 
-    return False
+    elif not key:
+        raise ValueError("getSkel has to be called with a valid key!")
```
sveneberth, line 179:
```suggestion
        raise ValueError(f"getSkel has to be called with a valid key! Got {key!r}")
```

src/viur/core/render/html/env/viur.py
```diff
@@ -150,63 +156,66 @@ def getSkel(render: Render, module: str, key: str = None, skel: str = "viewSkel"
...
+
+    if "canView" in dir(obj):
```
sveneberth, line 181: This is more performant, I think.

```suggestion
    if hasattr(obj, "canView"):
```

src/viur/core/render/html/env/viur.py
```diff
@@ -150,63 +156,66 @@ def getSkel(render: Render, module: str, key: str = None, skel: str = "viewSkel"
...
+        if not skel.read(key):
+            logging.info(f"getSkel: Entry {key!r} not found")
+            return None
+
+        if isinstance(obj, prototypes.singleton.Singleton):
+            isAllowed = obj.canView()
```
sveneberth, line 187: And following 
```suggestion
            is_allowed = obj.canView()
```

src/viur/core/render/html/env/viur.py
```diff
@@ -150,63 +156,66 @@ def getSkel(render: Render, module: str, key: str = None, skel: str = "viewSkel"
...
+                isAllowed = obj.canView("leaf", skel)
+
+        else:
+            isAllowed = obj.canView(skel)
+
+        if not isAllowed:
+            logging.error(f"getSkel: Access to {key} denied from canView")
+            return None
+
+    elif "listFilter" in dir(obj):
```
sveneberth, line 202:
```suggestion
    elif hasattr(obj, "listFilter"):
```

src/viur/core/render/html/env/viur.py
```diff
@@ -150,63 +156,66 @@ def getSkel(render: Render, module: str, key: str = None, skel: str = "viewSkel"
...
+        qry = skel.all().mergeExternalFilter({"key": str(key)})
+        qry = obj.listFilter(qry)
+        if not qry:
+            logging.info("listFilter permits getting entry, returning None")
+            return None
+
+        skel = qry.getSkel()
+        if not skel:
```
sveneberth, line 210: 🦭
```suggestion
        if not (skel := qry.getSkel()):
```

src/viur/core/render/html/env/viur.py
```diff
@@ -295,40 +304,59 @@ def modulePath(render: Render) -> str:
...
+        raise RuntimeError("getList: Invalid skel name provided")
+
+    # Test if any value of kwargs is an empty list
+    if _noEmptyFilter and any(isinstance(value, list) and not value for value in kwargs.values()):
+        return []
+
+    # Create initial query
+    query = skel.all().mergeExternalFilter(kwargs)
+
     if "listFilter" in dir(caller):
```
sveneberth, line 348:
```suggestion
    if hasattr(obj, "listFilter"):
```

sveneberth (changes requested):

src/viur/core/render/html/env/viur.py
phorward: Hello @sveneberth, this ValueError will only raise in case it is not a Singleton and no key was given. Therefore it makes no sense to output the key. If the key is valid or not is not part of this function.

src/viur/core/render/html/env/viur.py
```diff
@@ -150,63 +156,65 @@ def getSkel(render: Render, module: str, key: str = None, skel: str = "viewSkel"
...
+                isAllowed = obj.canView("node", skel)
+            else:
+                isAllowed = obj.canView("leaf", skel)
+
+        else:
+            isAllowed = obj.canView(skel)
+
+        if not isAllowed:
+            logging.error(f"getSkel: Access to {key} denied from canView")
+            return None
```
phorward, line 200:
```suggestion
            is_allowed = obj.canView()

        elif isinstance(obj, prototypes.tree.Tree):
            if skel["key"].kind == obj.nodeSkelCls.kindName:
                is_allowed = obj.canView("node", skel)
            else:
                is_allowed = obj.canView("leaf", skel)

        else:
            is_allowed = obj.canView(skel)

        if not is_allowed:
            logging.error(f"getSkel: Access to {key} denied from canView")
            return None
```

src/viur/core/render/html/env/viur.py
sveneberth: But if I only get this error message, I don't know without looking in the code whether the key I have transmitted is empty. After all, this is possible because the default argument is None. So it is possible to call the function without a key, but if I then get _"has to be called with a valid key"_, I would understand this to mean that the key I have submitted is invalid (wrong format or entry does not exist, ...). 

Then please change the message to: "Key is empty" or something like that.

sveneberth (approved):

sveneberth (approved):

PR #1346, opened 2024-12-06
feat: add more filter-options to `SelectCountryBone`

bb-mausbrand:
Refactoring and extension of the `SelectCountryBone`

- Moved code constants into the `SelectCountryBone` class
- `values` parameter either accepts a str representing a predefined group (currently "dach" and "eu"), a list of ISO-Codes (must match the Code-Length from `codes`), or a dict with matching iso-codes as keys and the full country-names as values. In case of None, the default (=all countries) will be used.


phorward (approved):

sveneberth (approved):
I would like to have type hints, but well ...

phorward:
> I would like to have type hints, but well ...

Next time...

PR #1354, opened 2024-12-18
feat: Add `*args` and `**kwars` to callable acess func

ArneGudermann:
Can you solve the following problem: If you have a skel in a list that only certain people should have access to.
Like:
```python
def bar(key, *args, **kwargs):
    user = current.user.get()
    if "manager" in user["access"]:
        return key == "example_key"
    return False
 
@access(bar)
def foo(key, param1, parma2):
    ....
```

### DIFF
--- a/src/viur/core/module.py
+++ b/src/viur/core/module.py
@@ -288,7 +288,7 @@ class Method:
                     # Callable directly tests access
                     if callable(acc):
                         try:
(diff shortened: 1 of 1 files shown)


ArneGudermann:
Can be resolved with `current.request.get().args` `current.request.get().kwargs`

PR #1358, opened 2025-01-03
fix: `BooleanBone.setBoneValue` doesn't respect language

phorward:
Also added some log output to `BaseBone.setBoneValue()` for better debugging of complex data.


ArneGudermann (approved):

sveneberth (approved):

sveneberth:
Can we backpatch this to `3.6`?

phorward:
> Can we backpatch this to `3.6`?

Yes we can, but I find it exhausting to manage two versions all the time. We should talk about this personally.

PR #1362, opened 2025-01-08
fix: `FileLeafSkel._inject_serving_url()` is the better choice

phorward:
`File.write()` didn't make serving_url injection.
Therefore, this refactoring is a much better solution.

### DIFF
--- a/src/viur/core/modules/file.py
+++ b/src/viur/core/modules/file.py
@@ -764,7 +764,7 @@ class File(Tree):
         :param public: True if the file should be publicly accessible.
         :return: Returns the key of the file object written. This can be associated e.g. with a FileBone.
         """
-        logging.info(f"{filename=} {mimetype=} {width=} {height=} {public=}")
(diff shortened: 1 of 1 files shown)


src/viur/core/modules/file.py
```diff
@@ -392,6 +392,31 @@ class FileLeafSkel(TreeSkel):
         }
     )
 
+    @classmethod
+    def __inject_serving_url(cls, skel: SkeletonInstance) -> None:
```
sveneberth, line 396:
```suggestion
    def _inject_serving_url(cls, skel: SkeletonInstance) -> None:
```

The name mangling make is uncomfortable if you are subclassing things (Access to `Skeleton._Skeleton__seo_key_trans` is already enough anoying for me ...). I would avoid it unless you have a real reason to keep things inside the same class and not make them available to sub or other classes. "Private" should be enough to signal internal use.

sveneberth (changes requested):

sveneberth (approved):

PR #1364, opened 2025-01-09
fix: General fixing and clean-up on `translate()`

phorward:
This pull request summarizes several problems:

- [x] Turn `tr_key` into `name`
- [x] Make `name` part of the db.Key for faster retrieval
- [x] Let all ViUR-core related translations have a `viur.core`-name-prefix
- [x] Make `name` (`tr_key`) required on creation (otherwise, it is kept readonly)
- [x] Provide a way for `conf.i18n.add_missing_translations` to only add or ignore a specific pattern (fnmatch)
- [x] Provide `translation.dump()` function for JSON-based retrieval of both public and private translations and to handle multiple patterns
  - [x] Provide a canView-like hook (`conf.i18n.dump_can_view`)
- [x] `SelectBone` with auto-create-translation-flag
- [x] Implement `default_variables` #1379 
- [x] In `translate`-Module a standard view for System-translations + Public translations
- [x] Don't add translation for error messages
- [x] Don't do a str-cast on an existing `translate` to use it as `key` for the new object

### DIFF
--- a/src/viur/core/bones/select.py
+++ b/src/viur/core/bones/select.py
@@ -46,6 +46,7 @@ class SelectBone(BaseBone):
         ] = None,
         values: dict | list | tuple | t.Callable | enum.EnumMeta = (),
         translation_key_prefix: str | t.Callable[[t.Self], str] = "",
+        add_missing_translations: bool = False,
         **kwargs
     ):
         """
@@ -61,6 +62,7 @@ class SelectBone(BaseBone):
         """
         super().__init__(defaultValue=defaultValue, **kwargs)
         self.translation_key_prefix = translation_key_prefix
+        self.add_missing_translations = add_missing_translations
 
         # handle list/tuple as dicts
         if isinstance(values, (list, tuple)):
@@ -92,19 +94,19 @@ class SelectBone(BaseBone):
 
                 assert isinstance(values, (dict, OrderedDict))
 
-            if self.languages:
-                prefix = self.translation_key_prefix
-                if callable(prefix):
-                    prefix = prefix(self)
+            prefix = self.translation_key_prefix
+            if callable(prefix):
+                prefix = prefix(self)
 
-                values = {
-                    key: label if isinstance(label, translate) else translate(
-                        f"{prefix}{label}", str(label),
-                        f"value {key} for {self.name}<{type(self).__name__}> "
-                        + f"in {self.skel_cls.__name__} in {self.skel_cls}"
-                    )
-                    for key, label in values.items()
-                }
+            values = {
+                key: label if isinstance(label, translate) else translate(
+                    f"{prefix}{key}", str(label),
+                    f"value {key} for {self.name}<{type(self).__name__}> "
+                    + f"in {self.skel_cls.__name__} in {self.skel_cls}",
+                    add_missing=self.add_missing_translations,
+                )
+                for key, label in values.items()
+            }
 
             return values
 
--- a/src/viur/core/bones/select.py
+++ b/src/viur/core/bones/select.py
@@ -92,17 +92,18 @@ class SelectBone(BaseBone):
 
                 assert isinstance(values, (dict, OrderedDict))
 
-            prefix = self.translation_key_prefix
-            if callable(prefix):
-                prefix = prefix(self)
+            if self.languages:
+                prefix = self.translation_key_prefix
+                if callable(prefix):
+                    prefix = prefix(self)
 
-            values = {
-                key: label if isinstance(label, translate) else translate(
-                    f"{prefix}{label}", str(label),
-                    f"value {key} for {self.name}<{type(self).__name__}> in {self.skel_cls.__name__} in {self.skel_cls}"
-                )
-                for key, label in values.items()
-            }
+                values = {
+                    key: label if isinstance(label, translate) else translate(
+                        f"{prefix}{label}", str(label),
+                        f"value {key} for {self.name}<{type(self).__name__}> in {self.skel_cls.__name__} in {self.skel_cls}"
+                    )
+                    for key, label in values.items()
+                }
 
             return values
 
--- a/src/viur/core/bones/select.py
+++ b/src/viur/core/bones/select.py
@@ -100,7 +100,8 @@ class SelectBone(BaseBone):
                 values = {
                     key: label if isinstance(label, translate) else translate(
                         f"{prefix}{label}", str(label),
-                        f"value {key} for {self.name}<{type(self).__name__}> in {self.skel_cls.__name__} in {self.skel_cls}"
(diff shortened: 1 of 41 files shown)


src/viur/core/i18n.py
```diff
@@ -205,7 +205,12 @@ def __str__(self) -> str:
 
             from viur.core.render.html.env.viur import translate as jinja_translate
 
-            if self.key not in systemTranslations and conf.i18n.add_missing_translations:
+            if (
+                not self.key.startswith("core.")
+                and not self.key.startswith("viur.")
```
phorward, line 210: wip

src/viur/core/bones/select.py
```diff
@@ -92,17 +92,18 @@ def __getattribute__(self, item):
...
+                if callable(prefix):
+                    prefix = prefix(self)
+
+                values = {
+                    key: label if isinstance(label, translate) else translate(
+                        f"{prefix}{label}", str(label),
+                        f"value {key} for {self.name}<{type(self).__name__}> in {self.skel_cls.__name__} in {self.skel_cls}"
+                    )
+                    for key, label in values.items()
+                }
```
phorward, line 106: wip

src/viur/core/modules/translation.py
```diff
@@ -21,25 +21,29 @@ class TranslationSkel(Skeleton):
     kindName = KINDNAME
 
     tr_key = StringBone(
+        required=True,
```
phorward, line 24: important!

src/viur/core/i18n.py
```diff
@@ -227,38 +235,56 @@ def __str__(self) -> str:
...
-                    key=self.key,
-                    hint=self.hint,
-                    default_text=self.defaultText,
-                    filename=self.filename,
-                    lineno=self.lineno,
-                    public=self.public,
-                )
+            if self.key not in systemTranslations:
+                # either the translate()-object has add_missing set
+                if not (add_missing := self.add_missing):
```
sveneberth, line 240: I find this behaviour confusing.

- So I set `add_missing` set to `False`, but if `conf.i18n.add_missing_translations` is `True`, it is added anyway.
If `add_missing` is set to `True`, it doesn't matter what is in `conf.i18n.add_missing_translations` and it's added too.


I would see this as a common condition:
If `self.add_missing` and `conf.i18n.add_missing_translations` are both `True`, add, otherwise not.

phorward: we can make it so.

phorward: Hi @sveneberth, sorry for the confusion. The initial consideration about this flag was, that either the translate-object has the `add_missing`-flag on first order, and when this is not True, the globally defined logic of `conf.i18n.add_missing_translations` takes place, which can either be just True, but also one or many fnmatch-style pattern to match the given translation key.

phorward:
@sveneberth I'm done with my requests.

Can you please add changes for your requests:

- [ ] Don't add translation for error messages
- [ ] Don't do a str-cast on an existing `translate` to use it as `key` for the new object

This issue is urgent, waiting for approval.

phorward:
@sveneberth thanks, this looks fine so far. Can you please approve the PR, I cannot self-approve.

src/viur/core/modules/translation.py
```diff
@@ -132,24 +133,29 @@ class TranslationSkel(Skeleton):
...
-    def preProcessSerializedData(cls, skelValues: SkeletonInstance, entity: db.Entity) -> db.Entity:
-        # Backward-compatibility: re-add the key for viur-core < v3.6
-        # TODO: Remove in ViUR4
-        entity["key"] = skelValues["tr_key"]
-        return super().preProcessSerializedData(skelValues, entity)
+    def write(cls, skel, **kwargs):
+
+        # Create the key from the name on initial write!
+        if not skel["key"]:
+            skel["key"] = db.Key(KINDNAME, skel["name"].lower())
```
sveneberth, line 156: This would overwrite data.

If there's already a complete skeleton with the key `"foo"`, and someone clicks on add and use the same key, it would overwrite the other skel without any warning etc.

src/viur/core/i18n.py
```diff
@@ -442,11 +476,11 @@ def add_missing_translation(
 
     # Ensure lowercase key
     key = key.lower()
-    entity = db.Query(KINDNAME).filter("tr_key =", key).getEntry()
-    if entity is not None:
+
+    # Check if key already exists
+    if db.Get(db.Key(KINDNAME, key)):
```
sveneberth, line 481: Because existing entities would still keep their numeric id as key, this would not work

```suggestion    
    entity = db.Query(KINDNAME).filter("name =", key).getEntry()
    if entity is not None:
```

src/viur/core/modules/user.py
```diff
@@ -799,7 +799,7 @@ def start(self):
         return self._user_module.render.edit(
             self.OtpSkel(),
             params={
-                "name": i18n.translate(self.NAME),
+                "name": i18n.translate("viur.core.modules.user." + self.NAME),
```
sveneberth, line 802:
```suggestion
                "name": i18n.translate(f"viur.core.modules.user.{self.NAME}"),
```

and for others please use f-string too

src/viur/core/i18n.py
sveneberth: it seems we have different use cases for this and other exceptions (you want an `or` logic, I an `and`). Let's talk about this in a meet.

sveneberth (changes requested):

src/viur/core/i18n.py
sveneberth: As discussed yesterday, I have implemented Always and Never in 58d2e3057.

src/viur/core/modules/translation.py
sveneberth: I didn't tested this, and this is kinda hakkish, but this might solve this issue:

```python
    @classmethod
    def write(cls, skel, **kwargs):

        def write_with_claimed_key():
            skel["key"] = db.Key(KINDNAME, skel["name"].lower())
            if db.Get(skel["key"]):
                raise ValueError(f'An TranslationSkel with {skel["name"]=} exists already!')
            return super().write(skel, **kwargs)

        # Create the key from the name on initial write!
        if not skel["key"]:
            if db.IsInTransaction():
                return write_with_claimed_key()
            else:
                return db.RunInTransaction(write_with_claimed_key)

        return super().write(skel, **kwargs)
```

phorward: I did some further tests on this: Because `name` is unique, `Skeleton.fromClient()` already reports the problem and the write() is not called. Therefore the behavior is safe here.

sveneberth (approved):

PR #1365, opened 2025-01-09
feat: Provide default `index`-function for `Tree` and `Singleton`

phorward:


src/viur/core/prototypes/tree.py
```diff
@@ -286,6 +286,20 @@ def pathToKey(self, key: db.Key):
...
 
+    @exposed
+    def index(self, skelType: SkelType = "node", parententry: t.Optional[db.Key | int | str] = None, **kwargs):
+        if not parententry:
+            repos = self.getAvailableRootNodes(**kwargs)
+            match len(repos):
+                case 0:
+                    raise errors.Unauthorized()
+                case 1:
+                    parententry = repos.pop()["key"]
```
sveneberth, line 297: I wouldn't pop it without a copy of it. If `getAvailableRootNodes` provides anything cached this would be modified.

sveneberth (changes requested):

ArneGudermann (approved):

sveneberth (approved):

PR #1366, opened 2025-01-09
feat: Provide a generic `Tree.dump()` function

phorward:
In many use-cases, a tree must be fetched for e.g. a workflow, a menu or similar. This helper function directly dumps a hierarchy/tree to a given level and with a given limit as JSON.

Its better, to provide a simple function for this, instead of copy a bad function multiple times in several projects.

### DIFF
--- a/src/viur/core/prototypes/tree.py
+++ b/src/viur/core/prototypes/tree.py
@@ -1,4 +1,5 @@
 import logging
+import json
 import typing as t
 from deprecated.sphinx import deprecated
 from viur.core import utils, errors, db, current
@@ -7,6 +8,7 @@ from viur.core.bones import KeyBone, SortIndexBone
 from viur.core.cache import flushCache
 from viur.core.skeleton import Skeleton, SkeletonInstance
 from viur.core.tasks import CallDeferred
+from viur.core.render.json.default import CustomJsonEncoder
 from .skelmodule import SkelModule
 
 
@@ -1049,6 +1051,69 @@ class Tree(SkelModule):
 
             if self.leafSkelCls:
                 self._clone_recursive("leaf", src_skel["key"], skel["key"], skel["parentrepo"])
+
+    # Helpers
+
+    @exposed
+    def dump(
+        self,
+        parententry: t.Optional[db.Key | int | str] = None,
+        depth: int = 3, limit:
+        int = 30,
+        **kwargs
+    ) -> str:
+        """
+        Dumps a Tree's content as JSON, based on a parententry.
+
+        :param parententry: Parententry to dump; If not given, the fuction tries to figure out a single parent entry.
+        :param depth: Depth to dump children. Can only be a maxiumum of 3.
+        :param limit: Limit of entries on each level.
+        """
+        if not utils.string.is_prefix(self.render.kind, "json"):
+            raise errors.BadRequest("Can only use this function on JSON-based renders")
+
+        current.request.get().response.headers["Content-Type"] = "application/json"
+
+        if depth < 1 or depth > 3:
+            raise errors.NotAcceptable(f"{'limit'!r} out of bounds, must be between 1 and 99")
+
+        if limit < 1 or limit > 30:
+            raise errors.NotAcceptable(f"{'limit'!r} out of bounds, must be between 1 and 99")
+
+        if not parententry:
+            repos = self.getAvailableRootNodes(**kwargs)
+
+            match len(repos):
+                case 0:
+                    raise errors.Unauthorized()
+                case 1:
+                    parententry = repos.pop()["key"]
+                case _:
+                    raise errors.NotAcceptable(f"Missing required parameter {'parententry'!r}")
+
+        def query(parententry):
+            q = self.viewSkel("node").all().filter("parententry", parententry)
+            if not (q := self.listFilter(q)):
+                raise errors.Unauthorized()
+
+            ret = [
+                self.render.renderSkelValues(skel) | {"_skeltype": "node", "_children": query(skel["key"])}
+                for skel in q.fetch(limit=limit)
+            ]
+
+            if self.leafSkelCls:
+                q = self.viewSkel("node").all().filter("parententry", parententry)
+                if not (q := self.listFilter(q)):
+                    raise errors.Unauthorized()
+
+                ret += [
+                    self.render.renderSkelValues(skel) | {"_skeltype": "leaf"}
+                    for skel in q.fetch(limit=limit)
+                ]
+
+            return ret
+
+        return json.dumps(query(parententry), cls=CustomJsonEncoder)
 
 
 Tree.vi = True


src/viur/core/prototypes/tree.py
```diff
@@ -1050,6 +1052,77 @@ def onCloned(self, skelType: SkelType, skel: SkeletonInstance, src_skel: Skeleto
...
                 self._clone_recursive("leaf", src_skel["key"], skel["key"], skel["parentrepo"])
 
+    # Helpers
+
+    @exposed
+    def dump(
+        self,
+        parententry: t.Optional[db.Key | int | str] = None,
+        depth: int = 3, limit:
+        int = 30,
```
ArneGudermann, line 1062:
```suggestion
        depth: int = 3, 
        limit: int = 30,
```

src/viur/core/prototypes/tree.py
```diff
@@ -1050,6 +1052,77 @@ def onCloned(self, skelType: SkelType, skel: SkeletonInstance, src_skel: Skeleto
...
+        :param depth: Depth to dump children. Can only be a maxiumum of 3.
+        :param limit: Limit of entries on each level.
+        """
+        if not utils.string.is_prefix(self.render.kind, "json"):
+            raise errors.BadRequest("Can only use this function on JSON-based renders")
+
+        current.request.get().response.headers["Content-Type"] = "application/json"
+
+        if depth < 1 or depth > 3:
+            raise errors.NotAcceptable(f"{'limit'!r} out of bounds, must be between 1 and 99")
```
ArneGudermann, line 1078:
```suggestion
            raise errors.NotAcceptable(f"'depth' out of bounds, must be between 1 and 3")
```

src/viur/core/prototypes/tree.py
```diff
@@ -1050,6 +1052,77 @@ def onCloned(self, skelType: SkelType, skel: SkeletonInstance, src_skel: Skeleto
...
+
+        if limit < 1 or limit > 30:
+            raise errors.NotAcceptable(f"{'limit'!r} out of bounds, must be between 1 and 99")
```
ArneGudermann, line 1081:
```suggestion
            raise errors.NotAcceptable(f"'limit' out of bounds, must be between 1 and 30")
```

src/viur/core/prototypes/tree.py
```diff
@@ -1050,6 +1052,77 @@ def onCloned(self, skelType: SkelType, skel: SkeletonInstance, src_skel: Skeleto
...
+
+            match len(repos):
+                case 0:
+                    raise errors.Unauthorized()
+                case 1:
+                    parententry = repos[0]["key"]
+                case _:
+                    raise errors.NotAcceptable(f"Missing required parameter {'parententry'!r}")
+
+        def query(parententry):
```
ArneGudermann, line 1094:
```suggestion
        def query(parententry, _depth : int = 0):
```

src/viur/core/prototypes/tree.py
```diff
@@ -1050,6 +1052,77 @@ def onCloned(self, skelType: SkelType, skel: SkeletonInstance, src_skel: Skeleto
...
+            q = self.viewSkel("node").all()
+            q.mergeExternalFilter(kwargs | {"parententry": parententry})
+
+            if not (q := self.listFilter(q)):
+                raise errors.Unauthorized()
+
+            self._apply_default_order(q)
+
+            ret = [
+                self.render.renderSkelValues(skel) | {"_skeltype": "node", "_children": query(skel["key"])}
```
ArneGudermann, line 1104:
```suggestion
                self.render.renderSkelValues(skel) |({"_skeltype": "node", "_children": query(skel["key"], _depth + 1)} if _depth + 1 <= depth else {})
```

ArneGudermann (changes requested):

src/viur/core/prototypes/tree.py
```diff
@@ -1050,6 +1052,88 @@ def onCloned(self, skelType: SkelType, skel: SkeletonInstance, src_skel: Skeleto
...
+        :param depth: Depth to dump children. Can only be a maxiumum of 3.
+        :param limit: Limit of entries on each level, can be maximum of 30 per level and type.
+        """
+        if not utils.string.is_prefix(self.render.kind, "json"):
+            raise errors.BadRequest("Can only use this function on JSON-based renders")
+
+        current.request.get().response.headers["Content-Type"] = "application/json"
+
+        if depth < 1 or depth > 3:
+            raise errors.NotAcceptable("'depth' out of bounds")
```
ArneGudermann: Why we not print out the borders?

ArneGudermann (changes requested):

phorward:
@ArneGudermann I've updated the PR with your change request, can you please re-review?

src/viur/core/prototypes/tree.py
```diff
@@ -1064,6 +1066,105 @@ def onCloned(self, skelType: SkelType, skel: SkeletonInstance, src_skel: Skeleto
...
```
ArneGudermann, line 1161: I have already mentioned it, why not print the borders here?

src/viur/core/prototypes/tree.py
```diff
@@ -1064,6 +1066,105 @@ def onCloned(self, skelType: SkelType, skel: SkeletonInstance, src_skel: Skeleto
...
+
+        if limit < 1 or limit > 30:
+            raise errors.NotAcceptable("'limit' out of bounds")
```
ArneGudermann, line 1164: And here ?

ArneGudermann (changes requested):

src/viur/core/prototypes/tree.py
```diff
@@ -1064,6 +1066,105 @@ def onCloned(self, skelType: SkelType, skel: SkeletonInstance, src_skel: Skeleto
...
+        """
+        if not utils.string.is_prefix(self.render.kind, "json"):
+            raise errors.BadRequest("Can only use this function on JSON-based renders")
+
+        current.request.get().response.headers["Content-Type"] = "application/json"
+
+        if depth < 1 or depth > 3:
+            raise errors.NotAcceptable("'depth' out of bounds")
+
+        if limit < 1 or limit > 30:
```
sveneberth, line 1163: in my opinion, this is an arbitrary limit without any logical derivation or basis for calculation.

I assume you want to prevent the request from becoming too large with large trees. But that is simply too arbitrary for any context.

Furthermore, you underestimate polynomial growth. Your limit of $30^3 = 27000$ is, with the right data, beyond what should be processed within a request.
Whereas the extension by one more level, but a reduction of the limit to 5 would still be a perfectly conceivable amount: $5^4 = 625$. 
While the theoretical worst-case scenario is unlikely to be used in practice, it is quite realistic to have more than 3 levels and more than 30 entries on at least one (probably the last level; the leafs).

I would therefore suggest offering the limitation in case you only want to display the first n entries per level in your view and want to reload the rest, but not limiting it so strictly.
In addition, I would add an absolute limit so that a maximum of, for example, 1000 entries, regardless of their level and number, can be loaded with one request.

sveneberth (changes requested):

phorward:
Can we put this into viur-toolkit @sveneberth ?

PR #1367, opened 2025-01-10
fix: determine a better path for a new `TranslateSkel`

sveneberth:
The problem was on the one hand that the path was first determined in `__str__`, which means that it is not the position where the translate object was created, but where it is converted to a string (this can also be in the renderer). I have therefore moved this to `__init__`.

Furthermore, the stack loaded in `walk_stack()` skipped too many frames (we actually need the first frame directly), so this couldn't work either -- except for Jinja. Now we load the stack explicitly from the top.

### DIFF
--- a/src/viur/core/i18n.py
+++ b/src/viur/core/i18n.py
@@ -74,11 +74,13 @@ Of course you can create skeletons / entries in the datastore in your project
 on your own. Just use the TranslateSkel).
 """  # FIXME: grammar, rst syntax
 import datetime
-import jinja2.ext as jinja2
 import logging
+import sys
 import traceback
 import typing as t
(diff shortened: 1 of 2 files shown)


sveneberth:
> @sveneberth can you please integrate this as part of #1364? I've already been on this, too, last week.

But #1364 address core 3.7. and is a wip draft. This is a working fix for 3.6

phorward:
@sveneberth ok for me. Would you integrate this afterwards into #1364?

sveneberth:
> @sveneberth ok for me. Would you integrate this afterwards into #1364?

sure

phorward (approved):

PR #1372, opened 2025-01-17
doc: Fix `SyntaxWarning: invalid escape sequence '\*'`

phorward:
The previous docstring caused InvalidEscapeSequence warnings with Python 3.12

### DIFF
--- a/src/viur/core/skeleton.py
+++ b/src/viur/core/skeleton.py
@@ -1760,7 +1760,7 @@ class RefSkel(RelSkel):
     @classmethod
     def fromSkel(cls, kindName: str, *args: list[str]) -> t.Type[RefSkel]:
         """
-            Creates a relSkel from a skeleton-class using only the bones explicitly named in args.
+            Creates a ``RefSkel`` from a skeleton-class using only the bones explicitly named in ``args``.
 
             :param args: List of bone names we'll adapt
             :return: A new instance of RefSkel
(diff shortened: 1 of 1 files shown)


src/viur/core/skeleton.py
```diff
@@ -1760,8 +1760,7 @@ class RefSkel(RelSkel):
     @classmethod
     def fromSkel(cls, kindName: str, *args: list[str]) -> t.Type[RefSkel]:
         """
-            Creates a relSkel from a skeleton-class using only the bones explicitly named
-            in \*args
+            Creates a relSkel from a skeleton-class using only the bones explicitly named in args.
```
sveneberth, line 1763: Or code inline formatting?
```suggestion
            Creates a relSkel from a skeleton-class using only the bones explicitly named in ``args``.
```

sveneberth, line 1763:
```
            :return: A new instance of RefSkel
```
does not match with `RelSkel` :thinking: 
```suggestion
            Creates a ``RefSkel`` from a skeleton-class using only the bones explicitly named in args.
```

sveneberth (changes requested):

src/viur/core/skeleton.py
phorward, line 1763:
```suggestion
            Creates a ``RefSkel`` from a skeleton-class using only the bones explicitly named in ``args``.
```

sveneberth (approved):

PR #1373, opened 2025-01-17
fix: `SkelModule` not able to handle empty index definitions

phorward:


sveneberth (approved):

sveneberth:
Since we decided to have 3.6 LTS we should start the fix there, or not?

PR #1376, opened 2025-01-22
feat: Header to render only specific bones in the json renderer

ArneGudermann:
This PR adds a check for a Header Field  `x-viur-bonelist`. The names of the bones to be rendered can be entered in this field.

### DIFF
--- a/src/viur/core/request.py
+++ b/src/viur/core/request.py
@@ -205,7 +205,7 @@ class Router:
         return path
 
     def _process(self):
-        if self.method not in ("get", "post", "head"):
+        if self.method not in ("get", "post", "head", "options"):
             logging.error(f"{self.method=} not supported")
             return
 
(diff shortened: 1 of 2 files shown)


akelch (approved):

src/viur/core/request.py
```diff
@@ -295,7 +295,8 @@ def _process(self):
             path = self._select_language(path)[1:]
 
             # Check for closed system
-            if conf.security.closed_system:
+            # preflight requests send without cookies
+            if conf.security.closed_system and self.method != "options":
```
phorward, line 299:
```suggestion
            if self.method != "options" and conf.security.closed_system:
```

phorward (changes requested):
So my suggestion would be the following:

- Evaluate an `x-viur-bonelist` in the viewSkel and create a dynamic SubSkel based on it
- The SubSkel is then integrated into the normal process in ViUR, data is retrieved with it and it only ever affects the bones that are in the subskel
- Nothing has to be adapted to any renderers
- If this works well, it could be extended to the BaseBone() as a general option.

Dynamic SubSkels are only available from viur-core 3.7 and I'm not patching that down to 3.6.

sveneberth:
I am completely on @phorward's side.

A renderer should transform data into a desired format. But it should not apply any logic to this data, except that which concerns the output format. Reducing or expanding the data does not belong in there, but in the module.
In addition, data should be filtered out as early as possible in order to be as efficient as possible for further processing.

PR #1378, opened 2025-01-22
fix: a `not caseSensitive` bone should lock the lower value

sveneberth:
The idea of _caseINSensitive_ bones is that the upper/lower case does not matter and `"Test"`, `"test"`, and `"TEST"` are equivalent.

If you now have a unique contstraint `UniqueLockMethod.SameValue`, it is still possible to create the three entries.

---

Alternatively, we could of course also introduce `UniqueLockMethod.EqualValue`, which controls this even more precisely.

### DIFF
--- a/src/viur/core/bones/base.py
+++ b/src/viur/core/bones/base.py
@@ -9,12 +9,13 @@ import copy
 import hashlib
 import inspect
 import logging
-from dataclasses import dataclass, field
-from datetime import datetime, timedelta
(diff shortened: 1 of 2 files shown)


phorward (approved):
ok for me

sveneberth:
Sorry, forgot the `multiple` case first. This makes the changes more complex due to the poor `_hashValueForUniquePropertyIndex` implementation.

PR #1384, opened 2025-01-23
feat: Evaluate `x-viur-bonelist` on default `viewSkel()`

phorward:
- Introduces generic `SkelModule.skel()` function to obtain a skeleton instance.
- Modifies all prototypes to accept `bones_from_request`

Replacement for #1376


src/viur/core/prototypes/skelmodule.py
```diff
@@ -108,7 +108,45 @@ def baseSkel(self, *args, **kwargs) -> SkeletonInstance:
...
+            # and (bonelist := current.request.get().kwargs.get("x-viur-bonelist"))  # param must be given (DEBUG!)
+            and (bonelist := current.request.get().request.headers.get("x-viur-bonelist"))  # header must be given
+        ):
+            bones |= {bone.strip() for bone in bonelist.split(",")}
+
+        # Return a subskel?
+        if bones:
+            # When coming from outside of a request, "*" must always be contained.
+            if bones_from_request:
+                return skel_cls.subskel("*", bones=bones)
```
ArneGudermann, line 144:
```suggestion
                bones.add("key")
                bones.add("name")
                return skel_cls.subskel("*", bones=bones)
```

phorward: as privately mentioned, "key" is always part of a subskel, and I wouldn't like to enforce "name".

src/viur/core/prototypes/list.py
```diff
@@ -37,7 +37,7 @@ def viewSkel(self, *args, **kwargs) -> SkeletonInstance:
 
             :return: Returns a Skeleton instance for viewing an entry.
         """
-        return self.baseSkel(*args, **kwargs)
+        return self.skel(bones_from_request=True, **kwargs)
```
sveneberth, line 40: - `*args` get lost
- `bones_from_request` cannot be conntrolled by calling `viewSkel()`? --> Please add in `viewSkel`s signature and pass it.
- 
```suggestion
        return self.skel(*args, bones_from_request=True, **kwargs)
```

sveneberth (changes requested):
Now we have a header that controls a bonelist. The header is evaluated by default for all viewSkels.
But what happens if I load another skeleton within the same request? Then the same bonelist is also used for the other skel. You don't want that ...


Furthermore, the [Vary header](https://httpwg.org/specs/rfc9110.html#field.vary) should be set in the response after (and if) the bonelist has been evaluated and applied.

src/viur/core/prototypes/skelmodule.py
```diff
@@ -108,7 +108,45 @@ def baseSkel(self, *args, **kwargs) -> SkeletonInstance:
...
+        The parameters `bones` and `bones_from_request` can be combined.
+        """
+        skel_cls = self._resolveSkelCls(**kwargs)
+        bones = set(bones) if bones else set()
+
+        if (
+            bones_from_request  # feature generally enabled?
+            and skel_cls.subSkels.get("*")  # a named subSkel "*"" must exist
+            # and (bonelist := current.request.get().kwargs.get("x-viur-bonelist"))  # param must be given (DEBUG!)
+            and (bonelist := current.request.get().request.headers.get("x-viur-bonelist"))  # header must be given
```
sveneberth, line 136: This is NOT restricted by a renderer or permission or admin context. You could also do this in the Html Render and remove bones, that are used in the templates. 
Since errors are not cached you could create an amount of intensive renderings (depending on the templates, module, skeleton for sure) and exhaust resources (front end instances, read ops, ...).

src/viur/core/prototypes/skelmodule.py
sveneberth, line 144: How does this affect our computed fields?

If I have a computed field A, which is `Always` calculated from another bone B and B is no longer included in the skeleton, we have the same problem as with the relskels

sveneberth (changes requested):

src/viur/core/prototypes/list.py
phorward: > * `*args` get lost

I want to get rid of this. It is a relic from days gone by and should be eliminated. To do it, we have to refactor all *Skel()-related stuff in prototypes, and it will also become a breaking change. Therefore, just throw it away for now here.

> * `bones_from_request` cannot be controlled by calling `viewSkel()`? --> Please add in `viewSkel`s signature and pass it.

We can do it this, do you want it to overwrite this as a default behavior?

src/viur/core/prototypes/skelmodule.py
phorward: Yes, this must be considered by the developer currently, who defines the star-subskel ("*").

akelch (approved):

src/viur/core/prototypes/skelmodule.py
phorward: This should be fixes, see latest version.

PR #1394, opened 2025-01-29
fix: Custom decorators do not work

### LINKED ISSUE #1381: Custom decorators do not work
Problem:

```python
@foo
@exposed
def bar(key):
...
```
The `foo` decorator  is never called

ArneGudermann:
Resolves #1381

### DIFF
--- a/src/viur/core/module.py
+++ b/src/viur/core/module.py
@@ -45,7 +45,7 @@ class Method:
         # Guards
         self.skey = False
         self.access = None
-        self.decorators = []
+        self.guards = []
 
     def __get__(self, obj, objtype=None):
         """
@@ -228,7 +228,7 @@ class Method:
         if conf.debug.trace:
             logging.debug(f"calling {self._func=} with cleaned {args=}, {kwargs=}")
         # call decorators in reversed because they are added in the reversed order
-        for func in reversed(self.decorators):
+        for func in reversed(self.guards):
             func(args=args, kwargs=kwargs, varargs=varargs, varkwargs=varkwargs)
 
         # call with instance when provided
--- a/src/viur/core/module.py
+++ b/src/viur/core/module.py
@@ -43,8 +43,9 @@ class Method:
         self.signature = inspect.signature(self._func)
 
         # Guards
-        self.skey = None
+        self.skey = False
         self.access = None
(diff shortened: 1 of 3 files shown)


src/viur/core/module.py
```diff
@@ -43,8 +43,9 @@ def __init__(self, func: t.Callable):
         self.signature = inspect.signature(self._func)
 
         # Guards
-        self.skey = None
+        self.skey = False
         self.access = None
+        self.decorators = []
```
phorward, line 48: I would like to call these "guards", instead of "decorators".
```suggestion
        self.guards = []
```

phorward (changes requested):
Hello @ArneGudermann,

Thanks for facing this issue. Can you please do the following:
1. Rename `decorators` into `guards`
2. Move the "validate"-Functions into the decorators and turn these into closures. This would make the Method members `access` and `skey` useless, as they are part of the closure function, which is then put into the `guards` list.

phorward:
>     2. Move the "validate"-Functions into the decorators and turn these into closures. This would make the Method members `access` and `skey` useless, as they are part of the closure function, which is then put into the `guards` list.

I've seen, that this is already the case. Therefore, Method.access and Method.skey are obsolete, and you just can remove them. I think the code is then okay so far. Can you test it?

src/viur/core/module.py
```diff
@@ -43,8 +43,9 @@ def __init__(self, func: t.Callable):
         self.signature = inspect.signature(self._func)
 
         # Guards
-        self.skey = None
+        self.skey = False
         self.access = None
```
phorward, line 47: I think these are obsolete, as the configuration is taken from the closure.

ArneGudermann: We use the `skey` and `access` vars on some ponits  like
https://github.com/viur-framework/viur-core/blob/3c8680fde2c95ba67392486d23cf6209641e7fab/src/viur/core/module.py#L339-L343

ArneGudermann:
I test it and it looks all good.

phorward (approved):

sveneberth (approved):
I have not noticed any problems with the changes in my project. Everything is running.

Issue #1385, opened 2025-01-24
Content of a `SortIndexBone` is being cloned when using `clone`


phorward:
The problem described in the title happens when you create an entry containing a SortIndexBone (and sorting by this) and afterwards clone this entry. The saved value created by SortIndexBone's `defaultValue` is just copied, and calculates always into the same value, so that sorting is impossible.

Workaround: Manually overwrite the cloned value to get unique values.

PR #1399, opened 2025-01-31
fix: Set default value on clone for Sortindex bones

ArneGudermann:
Fix for #1385.


phorward:
Closed in favor of #1401

PR #1403, opened 2025-02-05
feat: `RebuildSearchIndex` also with a specific key

phorward:
This is a refactoring and improvement of the RebuildSearchIndex task.

Sometimes it is necessary to refresh a single entry with a specific key, especially because the `refresh()`-logic is being used. Instead of adding a new task for this, this pull requests adds the ability to either update all kinds, all entries of a specific kind or a single entry with a given key.

This PR also refactors the naming, because "kindnames" where referred as "modules" before.

### DIFF
--- a/src/viur/core/skeleton.py
+++ b/src/viur/core/skeleton.py
@@ -1968,13 +1968,23 @@ class TaskUpdateSearchIndex(CallableTaskBase):
         return user is not None and "root" in user["access"]
 
     def dataSkel(self):
-        modules = ["*"] + listKnownSkeletons()
-        modules.sort()
         skel = BaseSkeleton().clone()
-        skel.module = SelectBone(descr="Module", values={x: translate(x) for x in modules}, required=True)
+
+        skel.skelname = SelectBone(
+            descr="Kind",
+            values={"*": "*"} | {name: translate(name) for name in sorted(listKnownSkeletons())},
+        )
+
+        skel.specific_key = StringBone(
+            descr="Key",
+            params={
+                "tooltip": "Optional: Key of one entry to refresh, ignores the kind selected above."
+            }
+        )
+
         return skel
 
-    def execute(self, module, *args, **kwargs):
+    def execute(self, skelname, specific_key, *args, **kwargs):
         usr = current.user.get()
         if not usr:
             logging.warning("Don't know who to inform after rebuilding finished")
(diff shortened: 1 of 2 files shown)


src/viur/core/skeleton.py
```diff
@@ -1968,34 +1968,60 @@ def canCall(self) -> bool:
...
         else:
             notify = usr["name"]
 
-        if module == "*":
-            for module in listKnownSkeletons():
-                logging.info("Rebuilding search index for module %r", module)
-                self._run(module, notify)
+        if skelname == "*":
+            if specific_key:
+                raise errors.BadRequest(f"Cannot run on all kinds with {specific_key=}")
```
ArneGudermann, line 1997: Why we raise an error here in the tooltip above is written that the kind is ignored.

ArneGudermann (changes requested):

src/viur/core/skeleton.py
phorward: Because skelname == "*" defines "all kinds"... an iteration over all available kinds. Therefore, it should not be usable with a key. I also made it this way because I was lazy... there could be more effort done into rewriting this task, but we have so many construction sites currently...

src/viur/core/skeleton.py
```diff
@@ -1968,34 +1968,60 @@ def canCall(self) -> bool:
...
-        modules.sort()
         skel = BaseSkeleton().clone()
-        skel.module = SelectBone(descr="Module", values={x: translate(x) for x in modules}, required=True)
+
+        skel.skelname = SelectBone(
+            descr="Kind",
+            values={"*": "*"} | {name: translate(name) for name in sorted(listKnownSkeletons())},
+        )
+
+        skel.specific_key = StringBone(
```
sveneberth, line 1978: How about making this multiple?

And if you are already on this point, I would like to have module multiple too. I often start this for several but not all modules.

src/viur/core/skeleton.py
```diff
@@ -1968,34 +1968,60 @@ def canCall(self) -> bool:
         return user is not None and "root" in user["access"]
 
     def dataSkel(self):
-        modules = ["*"] + listKnownSkeletons()
-        modules.sort()
         skel = BaseSkeleton().clone()
-        skel.module = SelectBone(descr="Module", values={x: translate(x) for x in modules}, required=True)
+
+        skel.skelname = SelectBone(
```
sveneberth, line 1973: It's the kindname, not the skelname. By skelname I understand the name of the class, i.e. FileLeafSkel, while the child name is file. It is also consistent with the attribute in the `Skeleton` / `SkeletonInstance`.

```suggestion
        skel.kindname = SelectBone(
```

src/viur/core/skeleton.py
sveneberth: Why don't we just ignore the kind as soon as a `specific_key` is provided?

With `params={"logic.visibleIf": "bool(specific_key)"},` we could even display this in the admin.

src/viur/core/skeleton.py
```diff
@@ -1968,34 +1968,60 @@ def canCall(self) -> bool:
...
+                try:
+                    specific_key = db.Key.from_legacy_urlsafe(specific_key)
+                    skelname = specific_key.kind
+                except Exception:
+                    pass  # just ignore, and use value of specific_key further.
+
+            if not skelname:
+                raise errors.BadRequest("Cannot run on unknown kind")
+
+            self._run(skelname, notify, **{"key": str(specific_key)})
```
sveneberth, line 2013: I think it's superfluous to allocate a dict and then unpack it directly, even though it can also be used as a direct argument.
```suggestion
            self._run(skelname, notify, key=str(specific_key))
```

sveneberth (changes requested):

src/viur/core/skeleton.py
phorward: It is implemented as such. The "*" must be explicitly selected, I removed required from skelname within this PR, so that either a skelname or a key can be provided.

sveneberth (changes requested):
.

src/viur/core/skeleton.py
phorward: Sure, this can be done in a separate PR by someone who invests a day or so into this.

src/viur/core/skeleton.py
sveneberth: well okay.

src/viur/core/skeleton.py
sveneberth: Why should we touch it twice when you're already working on the code?

It's just unnecessary extra work for everyone to create, test and review two PRs.

phorward:
Superseded by #1502.

PR #1404, opened 2025-02-05
fix: `FileBone.refresh()` should fix `serving_url`

phorward:
This is part of the "public-Files"-feature introduced with viur-core 3.7. In case of a RelationalBone configured as public, and an assigned image that is public but has no serving_url, the referenced entry should be refreshed as well.

### DIFF
--- a/src/viur/core/bones/relational.py
+++ b/src/viur/core/bones/relational.py
@@ -1027,11 +1027,8 @@ class RelationalBone(BaseBone):
         if not skel[name] or self.updateLevel == RelationalUpdateLevel.OnValueAssignment:
             return
 
-        for idx, lang, value in self.iter_bone_value(skel, name):
-            if value is None:
-                continue
-
-            if value["dest"]:
+        for _, _, value in self.iter_bone_value(skel, name):
+            if value and value["dest"]:
                 try:
                     target_skel = value["dest"].read()
                 except ValueError:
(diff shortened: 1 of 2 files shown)


src/viur/core/bones/relational.py
```diff
@@ -1027,11 +1027,8 @@ def refresh(self, skel: "SkeletonInstance", name: str) -> None:
         if not skel[name] or self.updateLevel == RelationalUpdateLevel.OnValueAssignment:
             return
 
-        for idx, lang, value in self.iter_bone_value(skel, name):
-            if value is None:
-                continue
-
-            if value["dest"]:
+        for _, _, value in self.iter_bone_value(skel, name):
+            if value and value["dest"]:
```
ArneGudermann, line 1031:
```suggestion
            if not (value and value["dest"]):
                continue
```
And indent everything underneath to the left, I think that always looks nicer, don't you?

ArneGudermann (changes requested):

src/viur/core/bones/relational.py
phorward: @ArneGudermann we can do it on a follow-up PR. I already wrote code but I cannot test it correctly as #1403 is not there right now, and I don't have time now.

ArneGudermann (approved):

PR #1406, opened 2025-02-05
fix: Load user in a deferred task

### LINKED ISSUE #1405: User inside a `CallDeferred` task is ALWAYS `None`
```
>>>  log.info(f"{current.user.get()=}")
[INFO] current.user.get()=None
```

sveneberth:
Fixes #1405

### DIFF
--- a/src/viur/core/modules/user.py
+++ b/src/viur/core/modules/user.py
@@ -1312,7 +1312,7 @@ class User(List):
     def getCurrentUser(self):
         session = current.session.get()
 
-        if session and session.loaded and (user := session.get("user")):
+        if session and (user := session.get("user")):
             skel = self.baseSkel()
             skel.setEntity(user)
             return skel
--- a/src/viur/core/modules/user.py
+++ b/src/viur/core/modules/user.py
@@ -1312,7 +1312,8 @@ class User(List):
     def getCurrentUser(self):
         session = current.session.get()
 
-        if session and (user := session.get("user")):
+        req = current.request.get()
+        if session and (session.loaded or req.is_deferred) and (user := session.get("user")):
             skel = self.baseSkel()
             skel.setEntity(user)
             return skel
--- a/src/viur/core/tasks.py
+++ b/src/viur/core/tasks.py
@@ -202,6 +202,9 @@ class TaskHandler(Module):
         if env:
             if "user" in env and env["user"]:
                 current.session.get()["user"] = env["user"]
+                # FIXME: We do not have a fully loaded session from the cookie here,
+                #        but only a partial session.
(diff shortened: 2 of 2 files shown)


ArneGudermann (commented):
Can we Check If the Task ist deferred an the loaf the User all the time ?

sveneberth:
> Can we Check If the Task ist deferred an the loaf the User all the time ?

Sorry, but I don't understand what you want to do?

phorward:
> > Can we Check If the Task ist deferred an the loaf the User all the time ?
> 
> Sorry, but I don't understand what you want to do?

He means, that in case the request is on a deferred task, the current user is automatically loaded. IMHO the user is provided together with the deferred call request. @ArneGudermann's proposal is useful.

sveneberth:
> > > Can we Check If the Task ist deferred an the loaf the User all the time ?
> > 
> > 
> > Sorry, but I don't understand what you want to do?
> 
> He means, that in case the request is on a deferred task, the current user is automatically loaded. IMHO the user is provided together with the deferred call request. @ArneGudermann's proposal is useful.

I still have no idea what I should change in this PR. 
Please give me an example, suggestion or alternative PR.

src/viur/core/modules/user.py
```diff
@@ -1312,7 +1312,7 @@ def secondFactorProviderByClass(self, cls) -> UserSecondFactorAuthentication:
     def getCurrentUser(self):
         session = current.session.get()
 
-        if session and session.loaded and (user := session.get("user")):
+        if session and (user := session.get("user")):
```
ArneGudermann, line 1315:
```suggestion
        req = current.request.get()
        if session and (session.loaded or req.is_deferred) and (user := session.get("user")):
```

sveneberth: We can do it that way if you like.

But that doesn't really make any difference. If the session is not `loaded`, then `session.get("user")` is `None` anyway and the condition is falsy.

ArneGudermann: Now it works in your case i think. If the session is not loaded we can get the current user if the request is deferred.

ArneGudermann (approved):

src/viur/core/modules/user.py
sveneberth: no, I mean in the non-deferred case it wouldn't matter whether you have
```py
if session and (user := session.get("user")):
```
or
```py
if session and (session.loaded or req.is_deferred) and (user := session.get("user")):
``` 
because if the session is not loaded (`session.loaded` is then `False`) there is no content in the unloaded session object and the `session.get()` cannot load anything and returns the default value `None`.
This means that both statements are logically the same.

phorward (approved):

PR #1407, opened 2025-02-05
fix: `BooleanBone.refresh()` doesn't respect language

phorward:
- Fixed `BooleanBone.refresh()` to respect a language setting
- Improved `BooleanBone.singleValueFromClient()` to use conf-values for booleans
- ~Fixed `Skeleton.write()` to not raise a catched Exception when it occurs (in this case, came from the resolved bug: Value of type bool is not iterable)~

### DIFF
--- a/src/viur/core/bones/boolean.py
+++ b/src/viur/core/bones/boolean.py
@@ -68,7 +68,10 @@ class BooleanBone(BaseBone):
 
             :param name: The property-name this bone has in its Skeleton (not the description!)
         """
-        if not isinstance(skel[boneName], bool):
+        if self.languages:
+            for lang in self.languages:
+                skel[boneName][lang] = utils.parse.bool(skel[boneName][lang], conf.bone_boolean_str2true)
+        else:
             skel[boneName] = utils.parse.bool(skel[boneName], conf.bone_boolean_str2true)
 
     def setBoneValue(
@@ -114,7 +117,7 @@ class BooleanBone(BaseBone):
             values.
             The serialized value should be suitable for storage in the database.
         """
-        return utils.parse.bool(value)
+        return utils.parse.bool(value, conf.bone_boolean_str2true)
 
     def buildDBFilter(
         self,
--- a/src/viur/core/bones/boolean.py
+++ b/src/viur/core/bones/boolean.py
@@ -60,7 +60,7 @@ class BooleanBone(BaseBone):
             return True
         return not bool(value)
 
-    def refresh(self, skel: 'viur.core.skeleton.SkeletonInstance', boneName: str) -> None:
+    def refresh(self, skel: 'viur.core.skeleton.SkeletonInstance', name: str) -> None:
         """
             Inverse of serialize. Evaluates whats
             read from the datastore and populates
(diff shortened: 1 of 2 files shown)


src/viur/core/bones/boolean.py
```diff
@@ -68,7 +68,10 @@ def refresh(self, skel: 'viur.core.skeleton.SkeletonInstance', boneName: str) ->
 
             :param name: The property-name this bone has in its Skeleton (not the description!)
         """
-        if not isinstance(skel[boneName], bool):
+        if self.languages:
+            for lang in self.languages:
+                skel[boneName][lang] = utils.parse.bool(skel[boneName][lang], conf.bone_boolean_str2true)
```
sveneberth, line 73: if `skel[boneName]` does not contain `lang` is not set, this will clash.

sveneberth (changes requested):

src/viur/core/skeleton.py
```diff
@@ -1258,9 +1258,8 @@ def __txn_write(write_skel):
                     # Serialize bone into entity
                     try:
                         bone.serialize(skel, bone_name, True)
-                    except Exception:
-                        logging.error(f"Failed to serialize {bone_name} {bone} {skel.accessedValues[bone_name]}")
-                        raise
+                    except Exception as e:
+                        logging.exception(e)
```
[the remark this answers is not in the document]
phorward: The error message is not useful. I think it's a problem with viur-toolkits Importer. We can leave this as is, if wanted.

src/viur/core/bones/boolean.py
phorward: Please clarify.

sveneberth: Not every `lang` must be inside the dict `skel[boneName]` (maybe someone added a new lang, or this was just omitted before). But if you iterate over every `lang in self.languages` you get a `KeyError` on `skel[boneName][lang]`.

phorward: Ok, so your suggestion for this is simply
```suggestion
                skel[boneName][lang] = utils.parse.bool(skel[boneName].get(lang), conf.bone_boolean_str2true)
```
and we're done?

sveneberth: I think this must be rather the default value
```suggestion
                skel[boneName][lang] = utils.parse.bool(skel[boneName].get(lang, self.getDefaultValue(skel)), conf.bone_boolean_str2true)
```

---

or 
```python
if lang not in skel[name]:
    continue
```


Not sure what the ViUR specification says and how this is handled in other bones...

src/viur/core/skeleton.py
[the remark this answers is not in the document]
sveneberth: Please keep the old behavior instead of silently accept errors.


Or was this a problem for you?
```suggestion                   
                        logging.error(f"Failed to serialize {bone_name} {bone} {skel.accessedValues.get(bone_name, '<<MISSING>>')}")
                        raise
```

KadirBalku (approved):

ArneGudermann (approved):

PR #1408, opened 2025-02-07
fix: Make translations usable

sveneberth:
* `tr_key` must be required, otherwise this makes no sense for mappings
* longer `translations` should be supported too
* `tr_key`, `translations` and `default_text` should not be HTML escaped, sometimes they contains links, or words that should be highlighted with e.g. `<strong/>`
* `tr_key`s are often longer than 50 char, but they should still be searchable
* Use custom `name` implementation, instead of the `key`. (E.g. used for tabs in admin)


phorward (approved):
Finally somebody who faught the StringBone-ugliness with the HTML!

ArneGudermann (approved):

PR #1413, opened 2025-02-14
chore: Disable all compatibility flags by default

sveneberth:
Like #1411 
and make things for the admin and @akelch easier in the future


phorward (changes requested):
Hello @sveneberth, I would like to propose the general removal of any old behavior from the system, not to just "disable" it. This relates to VIUR4.

phorward (approved):

PR #1415, opened 2025-02-17
feat: Evaluate `x-viur-bonelist` on default `viewSkel()` (light-version)

phorward:
This is a lighter version of #1384, integrating the feature only for `Singleton` and `List`.

### DIFF
--- a/src/viur/core/prototypes/skelmodule.py
+++ b/src/viur/core/prototypes/skelmodule.py
@@ -149,8 +149,7 @@ class SkelModule(Module):
             return skel_cls.subskel(bones=bones)
 
         # Otherwise, return full skeleton
-        # return skel_cls()  # FIXME: VIUR4...
-        return self.baseSkel(**kwargs)
+        return skel_cls()
 
     def _apply_default_order(self, query: db.Query):
         """
--- a/src/viur/core/prototypes/skelmodule.py
+++ b/src/viur/core/prototypes/skelmodule.py
@@ -47,8 +47,9 @@ def __load_indexes_from_file() -> dict[str, list]:
 
 DATASTORE_INDEXES = __load_indexes_from_file()
 
-X_VIUR_BONELIST = "X-VIUR-BONELIST"
+X_VIUR_BONELIST: t.Final[str] = "X-VIUR-BONELIST"
 """Defines the header parameter that might contain a client-defined bone list."""
+
 
 class SkelModule(Module):
     """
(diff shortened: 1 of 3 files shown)


ArneGudermann (dismissed):

src/viur/core/prototypes/skelmodule.py
```diff
@@ -110,6 +112,46 @@ def baseSkel(self, *args, **kwargs) -> SkeletonInstance:
         """
         return self._resolveSkelCls(*args, **kwargs)()
 
+    def skel(
```
sveneberth, line 115: I've in a project already a custom quick & dirty solution, based on #1384.

```py
    def skel_from_header(
        self,
        *args,
        bones_exclude: tuple[str, ...] | t.List[str] = (),
        **kwargs,
    ) -> SkeletonInstance:
        X_VIUR_BONELIST: t.Final[str] = "X-VIUR-BONELIST"

        skel_cls = self._resolveSkelCls(**kwargs)
        bones = set()

        if bonelist := current.request.get().request.headers.get(X_VIUR_BONELIST):
            if "*" not in skel_cls.subSkels:  # a named star-subskel "*"" must exist!
                raise errors.BadRequest(f"Use of {X_VIUR_BONELIST!r} requires for a star-subskel")
            bones |= {bone.strip() for bone in bonelist.split(",")}

        bones.difference_update(bones_exclude)

        if bones:  # Return a subskel?
            log.debug(f"{current.request.get().response.vary=}")
            current.request.get().response.vary = (X_VIUR_BONELIST, *(current.request.get().response.vary or ()))
            # When coming from outside of a request, "*" is always involved.
            return skel_cls.subskel("*", bones=bones)

        elif bones_exclude:  # Return full skel, without generally excluded bones
            bones.update(skel_cls.__boneMap__.keys())
            bones.difference_update(bones_exclude)
            return skel_cls(bones=bones)

        # Otherwise, return full skeleton
        return skel_cls()
```

Can you please integrate the `bones_exclude` param and the vary header in your method too?

sveneberth, line 115: Why don't you implement this directly in `baseSkel`? Now you have to customize every call (`baseSkel()` --> `skel()`). If you would implement this in `baseSkel`, all calls could remain the same and you would be more backwards compatible.

With yet another `*skel` method, it only gets more complicated.

sveneberth (changes requested):

src/viur/core/prototypes/skelmodule.py
phorward: Can you shortly give a use-case for `bones_exclude`?

phorward: @sveneberth I've added your proposal into my code, can you test and review?

src/viur/core/prototypes/skelmodule.py
```diff
@@ -108,7 +111,59 @@ def baseSkel(self, *args, **kwargs) -> SkeletonInstance:
...
+        :param bones: Allows to specify a list of bones to form a subskel.
+        :param exclude_bones: Provide a list of bones which are always excluded.
+
+        The parameters `bones` and `allow_client_defined` can be combined.
+        """
+        skel_cls = self._resolveSkelCls(**kwargs)
+        bones = set(bones) if bones else set()
+
+        if allow_client_defined:
+            # if bonelist := current.request.get().kwargs.get(X_VIUR_BONELIST.lower()):  # DEBUG
```
phorward, line 139: @sveneberth I wonder if this could also be useful for debugging, what do you think?

sveneberth: I don't need that. Sending the header works perfectly for me, even when testing.

sveneberth, line 139:
```suggestion
```

sveneberth (approved):
Works in general for me

src/viur/core/prototypes/skelmodule.py
phorward: no, let the comment stay.

Issue #1426, opened 2025-02-25
bug: `RecordBone` in `RecordBone` is not handled


phorward:
ViUR currently is not able to handle RecordBones in RecordBones.
```py
[2025-02-25 14:01:02,112] /.../viur/core/request.py:375 [ERROR] ViUR has caught an unhandled exception!
[2025-02-25 14:01:02,112] /.../viur/core/request.py:376 [ERROR] Read something from the datastore thats not a dict: <class 'viur.core.skeleton.SkeletonInstance'> <SkeletonInstance of subBoneRelation with {'subbone_name': 'blub'}> subrel
Traceback (most recent call last):
  File "~/myproject/.venv/lib/python3.12/site-packages/viur/core/request.py", line 350, in _process
    self._route(path)
  File "~/myproject/.venv/lib/python3.12/site-packages/viur/core/request.py", line 625, in _route
    res = caller(*self.args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "~/myproject/.venv/lib/python3.12/site-packages/viur/core/module.py", line 315, in __call__
    return self._func(self._instance, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "~/myproject/.venv/lib/python3.12/site-packages/viur/core/prototypes/list.py", line 280, in add
    skel.write()
  File "~/myproject/.venv/lib/python3.12/site-packages/viur/core/skeleton.py", line 1460, in write
    bone.postSavedHandler(skel, bone_name, key)
  File "~/myproject/.venv/lib/python3.12/site-packages/viur/core/bones/record.py", line 112, in postSavedHandler
    bone.postSavedHandler(using, bone_name, None)
  File "~/myproject/.venv/lib/python3.12/site-packages/viur/core/bones/record.py", line 108, in postSavedHandler
    for idx, lang, value in self.iter_bone_value(skel, boneName):
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "~/myproject/.venv/lib/python3.12/site-packages/viur/core/bones/base.py", line 1385, in iter_bone_value
    value = skel[name]
            ~~~~^^^^^^
  File "~/myproject/.venv/lib/python3.12/site-packages/viur/core/skeleton.py", line 264, in __getitem__
    boneInstance.unserialize(self, key)
[...]

PR #1427, opened 2025-02-25
fix: Check for SkeletonInstance in  singleValueUnserialize of `RecordBone`

ArneGudermann:
Fix for #1426.


ArneGudermann:
fixed by https://github.com/viur-framework/viur-core/pull/1453

PR #1429, opened 2025-02-26
fix: `None` is now correclty supported and value-check and support "null"

### LINKED ISSUE #1423: type cast `"None"` to `None` in request parameters
```python
@exposed
def my_method(self, a_key: str | None): ...
```

sveneberth:
Resolves #1423

- `"null"` is helpful if you build requests in JS have `null` values
- If a annotation is missing, it's `param.empty` but this is difference to a explicit set `None` type annotation and needs to processed separately as `empty`
- The other "simple types" do a type-cast and would raise a `ValueError` in case of invalid data, but `None` was always taken, regardless of the value?!

Here some examples of definition which had some issues before and are now working as expected:
```py
    @exposed
    def my_method(
        self,
        *,
        test0: None | db.Key | str = None,
        test1: None = None,
        test2: None | str = None,
        test3: str | None = None,
        test4=None,
        test5: int = None,
    ):
``` 

Of course `test1: None = None` is a little bit impractical, but since it's a valid syntax it should also work in this way and accept only None, "None" or "null". But test4 has no type annotation and accepts any value as-is.
providing `"None"` for `test2` would result into `None`, while `"None"` for `test3` would result into `"None"` (due to the order).  
However, my initially problem parameter `test0: None | db.Key | str = None,` is now working as expected: I can provide explicit `"None"` or `"null"` and it's no longer always `None`.


src/viur/core/module.py
```diff
@@ -96,8 +96,10 @@ def parse_value_by_annotation(annotation: type, name: str, value: str | list | t
                 return float(value)
             elif annotation is bool:
                 return utils.parse.bool(value)
-            elif annotation is types.NoneType:
-                return None
+            elif annotation is types.NoneType or annotation is None:
+                if value in (None, "None", "null"):
```
phorward, line 100: Should we turn this into something like with `utils.parse.bool`?

sveneberth: It think this case is too special, I never needed this before.

phorward (approved):

ArneGudermann (approved):

PR #1431, opened 2025-02-26
feat: Re-integration of Google Datastore API

phorward:
Removes the viur-datastore dependency from viur-core and continues using Google Datastore API calls again directly from viur-core.

This branch integrates the stuff from viur-datastore that was already initially creatred by @akelch. Brings back the entire tooling previously cut out from viur-core.

This pull request provides a working version already, which is a drop-in replacement for viur-core 3.7 projects. There might still be bugs or parts of the code to be improved, but the basic operations look fine.


src/viur/core/db/transport.py
```diff
@@ -0,0 +1,226 @@
...
+    return __client__.delete(keys)
+
+
+@deprecated(version="3.8.0", reason="Use 'db.delete' instead")
+def Delete(keys: t.Union[Entity, t.List[Entity], Key, t.List[Key]]):
+    return delete(keys)
+
+
+@deprecated(version="3.8.0", reason="Use 'db.utils.is_in_transaction' instead")
+def IsInTransaction() -> bool:
```
sveneberth, line 108: @ArneGudermann We've this backward method already in the utils. Do we need this twice?

ArneGudermann (approved):

PR #1440, opened 2025-03-13
feat: Add `DeleteEntitiesIter` for `viur-transactionmarker`

ArneGudermann:


src/viur/core/tasks.py
```diff
@@ -869,3 +869,14 @@ def handleEntry(cls, entry, customData):
...
+
+
+@PeriodicTask(interval=datetime.timedelta(hours=4))
+def start_clear_transaction_marker():
+    """
+        Removes old (expired) Transaction marker
+        https://cloud.google.com/datastore/docs/concepts/transactions?hl=en#using_transactions
+    """
+    query = db.Query("viur-transactionmarker").filter("creationdate <",
+                                                      datetime.datetime.now() - datetime.timedelta(seconds=270))
```
sveneberth, line 881: This time is too short.


If you delete this before the task run the task would be abortet.

https://github.com/viur-framework/viur-core/blob/b617b713c2e4959584100d8f13c201287f5e2a63/src/viur/core/tasks.py#L215-L218
We need a delta of a month or something like that because you can delay tasks in the future and tasks may fail a few times.

sveneberth (changes requested):

src/viur/core/tasks.py
ArneGudermann: Okay I set this to one month

ArneGudermann: Okay the problem is the max countdown for a Transaction is 270 seconds and only 90 in Idle so we can delete this  after 270 seconds imo.

phorward: Hi @ArneGudermann, can you comment this thing in the source-code? A reference to this issue would also be well.

sveneberth: Yes, the transaction times are correct. But they are totally irrelevant for the transaction marker.


The reason is that a deferred task was called within a transaction:
https://github.com/viur-framework/viur-core/blob/fa3f977e2778e61089881f4bc69514b91a3601a8/src/viur/core/tasks.py#L574-L578

And the tasks should only run if the transaction was successful and that is **when the marker entity exists**. If the transaction fails, the marker is also removed in the rollback. 
**AFTER** the transaction has been successful, this marker object **must now exist at least until the task has been fully executed**. And when the task is executed can be at the earliest 90 seconds after the call, but also any time in the future. The task can be called with an eta for the next month or a countdown of 90 hours. But the task queue could also be so full that the task is simply only executed after thousands of others. There is **no fixed time** here. You can't predict that. Actually, the transaction marker can therefore only be safely removed after the task has been fully executed and not at any fixed date. In reality, however, tasks are completed within 30 days of the call deferred call and anything after that can be considered a corpse IMO.

ArneGudermann: Okay I think I missunderstood this Feature im the Frist Place. Thank you for the explaing. We can set the time to one month or we can create a conf variable for that. Or we can set a success to the Marker after the transaction.

sveneberth: I think using a large value is the best option. Because the same transaction marker can be used by multiple tasks (if they're started in the same transaction). Synchronising this could be complex.

ArneGudermann: Okay think 1 month is okay.

phorward (approved):

PR #1441, opened 2025-03-17
feat: History module

bb-mausbrand:
adds the history module to enable logging of changes of data and actions


src/viur/core/modules/history.py
```diff
@@ -0,0 +1,742 @@
...
+        updateLevel=RelationalUpdateLevel.OnValueAssignment,
+        searchable=True,
+        refKeys=[
+            "key",
+            "name",
+            "lastname",
+            "firstname",
+            "personnelnumber",
+            "last_audit",
+            "company",
```
sveneberth, line 56: These bones does not exist in the `UserSkel`
```suggestion
```

Please remove them from the  related mapping and schema too

src/viur/core/modules/history.py
```diff
@@ -0,0 +1,742 @@
...
+
+            schema = bone_to_schema(name, bone)
+
+            if isinstance(schema, dict):
+                self.schema.append(schema)
+            else:
+                self.schema.extend(schema)
+
+    def _get_table(self, skel):
+        # FIXME: Currently not in use!
```
sveneberth, line 385: then please remove this.

This looks identical like `select_or_create_table` anyway

src/viur/core/modules/history.py
```diff
@@ -0,0 +1,742 @@
...
+        return self.client.insert_rows(self.table, [skel_to_bigquery(skel)])
+
+
+class ViurHistory(List):
+    """
+    ViUR history module
+    """
+    kindName = "viur-history"
+
+    adminInfo = {
```
sveneberth, line 442: English please

src/viur/core/modules/history.py
```diff
@@ -0,0 +1,742 @@
...
+
+    def canDelete(self, _skel):
+        return False
+
+    def canAdd(self):
+        return False
+
+    # Module-specific functions
+
+    def _create_diff(self, new: dict, old: dict, diff_excludes: {str} = set()):
```
sveneberth, line 486: Are you sure this is a valid annotation? At least PyCharm does not understand it :/

```suggestion
    def _create_diff(self, new: dict, old: dict, diff_excludes: set[str] = set()):
```

src/viur/core/modules/history.py
```diff
@@ -0,0 +1,742 @@
...
+        key = "-".join(
+            part for part in (
+                entry["action"],
+                entry["current_kind"],
+                entry["timestamp"].isoformat()
+            ) if part
+        )
+
+        # write into datastore via history module
+        if "viur" in conf.get("viur.history.database"):
```
sveneberth, line 687: This wouldn't be a valid syntax in config strict mode

src/viur/core/modules/history.py
```diff
@@ -0,0 +1,742 @@
...
+            # need to do this as biquery functions modifies
+            # entry and seems to be called first
+            if conf.instance.is_dev_server:
+                entry = entry.copy()  # need to do this as biquery functions modifiy entry
+
+            conf.main_app.viur_history.write_to_bigquery_deferred(key, entry)
+
+        return key
+
+    def write(self, key: str, entry: dict, deferred: bool = False):
```
sveneberth, line 701: What's the purpose of `deferred` ?

src/viur/core/skeleton.py
```diff
@@ -905,6 +905,84 @@ def fulltextSearch(self, queryString: str, databaseQuery: db.Query) -> list[db.E
         return [resultEntryMap[x[0]] for x in resultList[:databaseQuery.queries.limit]]
 
 
+class HistoryAdapter(DatabaseAdapter):
```
sveneberth, line 908: Please place this in the history module. This is to special for the skeleton.

sveneberth (changes requested):

phorward (changes requested):
I have same concerns as @sveneberth. Additionally, I would like to move this huger thing into viur-core 3.8, so develop should be the target branch.

src/viur/core/modules/history.py
```diff
@@ -0,0 +1,596 @@
...
+    }
+
+    HISTORY_VERSION = 1
+    """
+    History format version.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.bigquery = BigQueryHistory()
```
phorward, line 345: This should be customizable and sub-classable via a `History.BigQueryHistoryCls` member.
```py
   BigQueryHistoryCls = BigQueryHistory
```
and then
```suggestion
        self.bigquery = self.BigQueryHistoryCls and self.BigQueryHistoryCls()
```
which will also allow to entirely unset the BigQueryHistory-adapter.

src/viur/core/modules/history.py
```diff
@@ -0,0 +1,596 @@
...
+        self.bigquery.write_row(entry)
+        logging.info(f"History entry {key=} written to biquery")
+
+    @tasks.CallDeferred
+    def write_to_bigquery_deferred(self, key: str, entry: dict):
+        self.write_to_bigquery(key, entry)
+
+
+ViurHistory.json = True
+ViurHistory.admin = True
```
phorward, line 596:
```suggestion
History.json = True
History.admin = True
```

src/viur/core/modules/history.py
```diff
@@ -0,0 +1,596 @@
...
+                entry["timestamp"].isoformat()
+            ) if part
+        )
+
+        # write into datastore via history module
+        if "viur" in conf.history.databases:
+            self.write_deferred(key, entry)
+
+        # write into BigQuery
+        if "bigquery" in conf.history.databases:
```
phorward, line 554:
```suggestion
        if self.bigquery and "bigquery" in conf.history.databases:
```

phorward (changes requested):
Hi @ArneGudermann, we're on a good way, I need some more customization features here.

src/viur/core/modules/history.py
```diff
@@ -0,0 +1,594 @@
...
+    )
+
+
+class BigQueryHistory:
+    """
+    Schema and connector for BigQuery history entries.
+    """
+    IGNORE_BONES = (
+        "viurCurrentSeoKeys",
+    )
```
phorward, line 86: This seems to be unused. Can you please re-check?
```suggestion
```

phorward (changes requested):
Hello, this looks good already :+1:, one small issue, see comment below.

phorward (approved):

PR #1442, opened 2025-03-19
fix: `singleValueUnserialize` is now working more stable

sveneberth:
PR #1425 introduced a `singleValueUnserialize` implementation that has no robustness against any other formats.

We still have old StringBone values in the database (`val={'val': '0.5', 'idx': '0.5'}`). This was never noticed, as there was never an error before. Now the value can neither be loaded nor ignored, nor does the `getDefaultValue` in except work without parameters. So no rebuild search index can be made with `refresh()` if the value is not loaded in the first place.


phorward (approved):
OK, thank you. Didn't got around this case.

PR #1445, opened 2025-03-25
feat: Allow `db.Key` in the `index`-method

ArneGudermann:
With this now you can call the index Methode direct with a db.Key like:
`/foo/agNmb29yCQsSA2Jhchh7DA`
instead of 
`/foo/view/agNmb29yCQsSA2Jhchh7DA`

Second if you have a Method like:
`def index(self, key: db.Key | str | None = None)`
the `key` is now parsed to a `db.Key` if this is possible.


phorward (changes requested):
Hi @ArneGudermann, this is a breaking change, as the index-function now behaves differently than before. Anyway, as the SEO-stuff is being reworked, we can do it so. You should handle the commented stuff. I would like to discuss this in a viur-meeting.

phorward (changes requested):
I'm still not completely satisfied with this. I think it could be solved more simply, like with the SEO keys: Wouldn't it be enough to simply try the following if the SEO key is not found (here in context, example from List.index):

```py
skel = self.viewSkel(
 allow_client_defined=utils.string.is_prefix(self.render.kind, "json"),
 _excludeFromAccessLog=True,
)

#---
if skel := (
    # Try to match the key
    skel.all().filter("__key__ =", args[0]).getSkel()
    # Try to match a SEO key...
    or skel.all().filter("viur.viurActiveSeoKeys =", str(args[0]).lower()).getSkel()
):
#---
 db.current_db_access_log.get(set()).add(skel["key"])
 if not self.canView(skel):
 raise errors.Forbidden()
 seoUrl = utils.seoUrlToEntry(self.moduleName, skel)
    # Check whether this is the current seo-key, otherwise redirect to it

    if current.request.get().request.path.lower() != seoUrl:
 raise errors.Redirect(seoUrl, status=301)
 self.onView(skel)
 return self.render.view(skel)
```

that would be all... so it's only about the if... that would just try one query with the key and one with the SEO key. You could also reverse the order.

phorward (approved):

PR #1447, opened 2025-03-27
fix: Save the session in a transaction

ArneGudermann:


phorward:
Hi @ArneGudermann we should discuss this in the viur-meeting. IMHO this doesn't change anything, as the entire session entry is being rewritten on every write, so no transaction is required. @skoegl had a similar problem here.

phorward:
Rejected.

PR #1448, opened 2025-03-27
feat: Allow `timedelta` for `session_life_time`

ArneGudermann:


sveneberth (changes requested):
Doesn't it make more sense to use timedelta by default, like introduced in #1086 and #1083 and https://github.com/viur-framework/viur-core/pull/1078 ?

sveneberth (approved):

phorward (approved):

PR #1457, opened 2025-04-10
feat: Apply `escape_html`-settings in `StringBone.refresh()` 

ArneGudermann:
With this PR the value of a Stringbone is unescaped or escaped on an refresh depends on `escape_html`

### DIFF
--- a/src/viur/core/bones/string.py
+++ b/src/viur/core/bones/string.py
@@ -80,25 +80,19 @@ class StringBone(BaseBone):
         In addition, it is checked whether the variable must be escaped or not.
         """
         if isinstance(value, str):
-            pass  # Nothing todo here
+            return value
         elif isinstance(value, Number):
-            value = str(value)
+            return str(value)
         elif isinstance(value, (datetime.datetime, datetime.date, datetime.time)):
-            value = value.isoformat()
+            return value.isoformat()
         elif isinstance(value, db.Key):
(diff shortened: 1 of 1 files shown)


phorward (approved):

phorward (dismissed):

src/viur/core/bones/string.py
```diff
@@ -331,7 +331,13 @@ def refresh(self, skel: "SkeletonInstance", bone_name: str) -> None:
         # TODO: duplicate code, this is the same iteration logic as in NumericBone
         new_value = {}
         for _, lang, value in self.iter_bone_value(skel, bone_name):
-            new_value.setdefault(lang, []).append(self.type_coerce_single_value(value))
+            value = self.type_coerce_single_value(value)
+            if self.escape_html:
+                value = utils.string.escape(value)
+            else:
```
sveneberth, line 337: But wouldn't this lead to 

"hello & bye" --> "hello &amp; bye" --> "hello &amp;amp; bye" --> "hello &amp;amp; bye"


```suggestion
            if not self.escape_html:
```

sveneberth (changes requested):

src/viur/core/bones/string.py
ArneGudermann: `&` is not translated in `escape`

PR #1458, opened 2025-04-16
fix: Update user's data in session on `user/edit`

ArneGudermann:
This PR fixes the following problem. When you edit a user, the user's session is not edited. This means, for example, that if you revoke a right from a user, the user retains this right until the session expires. Or if you add one, it only becomes usable when you log out and log in again.

### DIFF
--- a/src/viur/core/modules/user.py
+++ b/src/viur/core/modules/user.py
@@ -1579,7 +1579,8 @@ class User(List):
         # In case the user is set to inactive, kill all sessions
         if self.is_active(skel) is False:
             session.killSessionByUser(skel["key"])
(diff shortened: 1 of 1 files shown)


src/viur/core/modules/user.py
```diff
@@ -1579,6 +1579,10 @@ def onEdited(self, skel):
         # In case the user is set to inactive, kill all sessions
         if self.is_active(skel) is False:
             session.killSessionByUser(skel["key"])
+        # Update all Sessions
```
phorward, line 1582:
```suggestion

        # Update user setting in all sessions
```

PR #1460, opened 2025-04-16
fix: `dump` does not yield anything if there's no pattern filter given

sveneberth:


phorward (changes requested):
@sveneberth the behavior is correct! There should be no output when no pattern was given.

sveneberth:
> @sveneberth the behavior is correct! There should be no output when no pattern was given.

And how do I get all (public) translations via this great method that does not allow `*` patterns, but still has the parameter mandatory?

phorward:
> > @sveneberth the behavior is correct! There should be no output when no pattern was given.
> 
> And how do I get all (public) translations via this great method that does not allow `*` patterns, but still has the parameter mandatory?

You have to use a pattern. That's the point.

sveneberth:
> > > @sveneberth the behavior is correct! There should be no output when no pattern was given.
> > 
> > 
> > And how do I get all (public) translations via this great method that does not allow `*` patterns, but still has the parameter mandatory?
> 
> You have to use a pattern. That's the point.

So I should use this pattern instead of just taking everything directly?

```
/json/_translation/get_public?languages=de&languages=fr&pattern=a*&pattern=b*&pattern=c*&pattern=d*&pattern=e*&pattern=f*&pattern=g*&pattern=h*&pattern=i*&pattern=j*&pattern=k*&pattern=l*&pattern=m*&pattern=n*&pattern=o*&pattern=p*&pattern=q*&pattern=r*&pattern=s*&pattern=t*&pattern=u*&pattern=v*&pattern=w*&pattern=x*&pattern=y*&pattern=z*&pattern=A*&pattern=B*&pattern=C*&pattern=D*&pattern=E*&pattern=F*&pattern=G*&pattern=H*&pattern=I*&pattern=J*&pattern=K*&pattern=L*&pattern=M*&pattern=N*&pattern=O*&pattern=P*&pattern=Q*&pattern=R*&pattern=S*&pattern=T*&pattern=U*&pattern=V*&pattern=W*&pattern=X*&pattern=Y*&pattern=Z*&pattern=0*&pattern=1*&pattern=2*&pattern=3*&pattern=4*&pattern=5*&pattern=6*&pattern=7*&pattern=8*&pattern=9*
```

Something like this can only exist in ViUR ...

phorward (approved):

PR #1463, opened 2025-04-17
fix: Fix `access_log` for `db.delete`

ArneGudermann:
This PR fix this Problem.
image

### DIFF
--- a/src/viur/core/db/transport.py
+++ b/src/viur/core/db/transport.py
@@ -42,7 +42,7 @@ def get(keys: t.Union[Key, t.List[Key]]) -> t.Union[t.List[Entity], Entity, None
     """
     _write_to_access_log(keys)
 
-    if isinstance(keys, (list,set,tuple)):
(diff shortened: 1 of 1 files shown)


src/viur/core/db/transport.py
```diff
@@ -40,18 +40,13 @@ def get(keys: t.Union[Key, t.List[Key]]) -> t.Union[t.List[Entity], Entity, None
...
-    access_log = current_db_access_log.get()
-
-    if isinstance(keys, list):
-        if isinstance(access_log, set):
-            access_log.update(set(keys))
+    _write_to_access_log(keys)
 
+    if isinstance(keys, (list, set, tuple)):
         res_list = list(__client__.get_multi(keys))
         res_list.sort(key=lambda x: keys.index(x.key) if x else -1)
```
phorward, line 47:
```suggestion
        res_list.sort(key=lambda k: keys.index(k.key) if k else -1)
```

phorward (approved):

PR #1477, opened 2025-05-21
fix: Add check to allow override the `script` module

ArneGudermann:
With this PR it's not possible to override the `script` module. 
If you need for example an extra hook for an `onEdit`.

### DIFF
--- a/src/viur/core/__init__.py
+++ b/src/viur/core/__init__.py
@@ -146,7 +146,9 @@ def __build_app(modules: ModuleType | object, renderers: ModuleType | object, de
     modules._tasks = TaskHandler
     modules._moduleconf = ModuleConf
     modules._translation = Translation
-    modules.script = Script
+    # Check whether “script” is contained in modules so that it can be overwritten
+    if "script" not in vars(modules).keys() or not inspect.isclass(modules.script):
+        modules.script = Script
 
     # Resolver defines the URL mapping
     resolver = {}


phorward (changes requested):
I would like to improve this generally for any system-module:

```py
    for name, cls in {
        "_tasks: TaskHandler,
        "_moduleconf: ModuleConf,
        "_translation: Translation,
        "script: Script,
    }.items():
        # Check whether name is contained in modules so that it can be overwritten
        if name not in vars(modules):
            setattr(modules, name, cls)
            
        assert inspect.isclass(getattr(modules, name))
```

Can you test this?

src/viur/core/__init__.py
```diff
@@ -143,10 +143,17 @@ def __build_app(modules: ModuleType | object, renderers: ModuleType | object, de
...
+        "_tasks": TaskHandler,
+        "_moduleconf": ModuleConf,
+        "_translation": Translation,
+        "script": Script,
+    }.items():
+        # Check whether name is contained in modules so that it can be overwritten
+        if name not in vars(modules):
+            setattr(modules, name, cls)
+
+        assert issubclass(getattr(modules, name), Module)
```
sveneberth, line 156: We could enforce subclasses only
```suggestion
        assert issubclass(getattr(modules, name), cls)
```

sveneberth (changes requested):
In general, this is fine for me. However, this does not ensure that the mount point has the same name. If you do not mount `Task` under `_tasks`, you will have massive problems.
Either we exclude this or leave it in the full responsibility of the developer.

ArneGudermann:
@sveneberth @phorward Let's discuss this in the viur-meeting.

phorward (approved):

sveneberth (approved):

PR #1492, opened 2025-05-28
feat: `conf.db_query_max_limit` and `conf.db_query_default_limit`

phorward:
Allows to configure defaults and maximum for query limits.

Removes `db_engine` as this config switch becomes obsolete in viur-core 3.8.


phorward:
@sveneberth: I've removed any limit checking now; Limit is only enforced in `Query.mergeExternalFilter()`.

ArneGudermann (approved):

PR #1497, opened 2025-06-18
feat: Change user selections regarding login/secondfactor-providers to `ActionSkel`-paradigm

phorward:
This is a breaking change when invoking the `User.index()` function (which is not the case AFAIK). There may be bugs included.


src/viur/core/modules/user.py
```diff
@@ -1452,6 +1488,60 @@ def authenticateUser(self, key: db.Key, **kwargs):
...
+            provider = SelectBone(
+                descr="Authentication method",
+                required=True,
+                values=providers,
+            )
+
+        return SelectAuthenticationProviderSkel()
+
+    @exposed
+    def selectauthenticationprovider(self, **kwargs):
```
ArneGudermann, line 1509:
```suggestion
    def select_authentication_provider(self, **kwargs):
```
I find this version cleaner what do you think?

ArneGudermann (changes requested):

src/viur/core/modules/user.py
phorward: we can do it. The reason why I chose this name is, that a "selectauthenticationprovider"-action will return a "selectauthenticationproviderSuccess" (for legacy reasons). A "select_authentication_provider" will return "select_authentication_providerSuccess", which looks a little bit dumb.

PR #1502, opened 2025-06-25
feat: `SkeletonMaintenanceTask`

phorward:
Implements a SkeletonMaintenanceTask that can be used for

- refresh
- delete
- count

Specific skeletons, which can be filtered either via query or via Logics expressions.

This pull request entirely substitutes "RebuildSearchIndex". Use Skeletion Maintainnce Task with Mode Refresh for that now.

### DIFF
--- a/src/viur/core/skeleton/tasks.py
+++ b/src/viur/core/skeleton/tasks.py
@@ -257,7 +257,7 @@ class SkeletonMaintenanceTask(tasks.CallableTaskBase):
         condition = RawBone(
             descr="Condition",
             required=True,
-            defaultValue="False  # fused: by default, doesn't affect anything.\n",
+            defaultValue="False  # fused: by default, doesn't affect anything.",
             params={
                 "tooltip": "Enter a Logics expression here to filter entries by specific skeleton values."
             },


src/viur/core/skeleton/tasks.py
```diff
@@ -129,151 +141,150 @@ def __txn_update(_skel, key, srcRelKey):
...
+    def execute(self, task, kinds, filters, condition):
         try:
-            email.send_email(dests=notify, stringTemplate=txt, skel=None)
-        except Exception as exc:  # noqa; OverQuota, whatever
-            logging.exception(f"Failed to notify {notify}")
+            logics.Logics(condition)
+        except logics.ParseException as e:
+            raise errors.BadRequest(f"Error parsing condition {e}")
+
+        notify = current.user.get()["name"]
```
ArneGudermann, line 272: Maybe we can add Relationbone where you can pick an user or multipe to notify ? What do you think ?

src/viur/core/skeleton/tasks.py
```diff
@@ -129,151 +141,150 @@ def __txn_update(_skel, key, srcRelKey):
...
+
+        for kind in kinds:
```
ArneGudermann, line 274: What if I want to run a task for all kinds the I musst select all and send it ? Can we use a wildcard here ?

ArneGudermann (approved):

src/viur/core/skeleton/tasks.py
ArneGudermann: Good Idea i Take a Look in that tomorrow

PR #1514, opened 2025-07-09
feat: Add cache layer for the datastore adapter

ArneGudermann:


src/viur/core/db/transport.py
```diff
@@ -83,10 +99,23 @@ def delete(keys: t.Union[Entity, t.List[Entity], Key, t.List[Key]]):
...
+        for key in keys:
+            if isinstance(key, Key):
+                keys_data.append(key)
+            elif isinstance(key, Entity):
+                keys_data.append(key.key)
+    cache.delete(keys_data)
+    if single_request:
+        return __client__.delete(keys_data)
+
+    return __client__.delete_multi(keys_data)
```
phorward, line 118: Same here. Assume keys is always an iterable.

```suggestion
    keys = (k.key if isinstance(k, Entity) else k for k in utils.ensure_iterable(keys))
    if not keys:
        return
    
    cache.delete(keys)
    if len(keys) == 1:
        return __client__.delete(keys)

    return __client__.delete_multi(keys)
```

phorward (changes requested):
Thanks for facing this.

I'm a little bit astonished how bloated this code is, especially regarding this multiple-key feature. Is this anywhere used in ViUR, except in legacy code? 

Anyway. I think this code might be improved in the cache-module, too.

My recommendation: When code that either handles one or a list of keys is written, write the code as it always handles a list of keys, so that the `len(keys) == 1`-case becomes the special case. This eliminates lots of bloating regarding isinstance()-checks and iterable-checking.

Can you please test my suggestions, and please don't forget to import `utils` to use `ensure_iterable`.

src/viur/core/db/transport.py
ArneGudermann: @phorward I use get and not get_multi because it's faster.

phorward:
> @phorward we can not use `utils` due a circular import

I've fixed that and also a a bug and update the code to use `utils.ensure_iterable()` can you please have a look and - if you have test-cases - test it?

phorward (changes requested):

src/viur/core/db/cache.py
```diff
@@ -48,23 +49,24 @@ def get(keys: t.Union[Key, list[Key]], namespace: t.Optional[str] = None) -> t.U
...
         while keys:
-            cached_data |= conf.db.memcache_client.get_multi(keys[:MEMCACHE_MAX_BATCH_SIZE], namespace=namespace)
+            if cached_data := conf.db_memcache_client.get_multi(keys[:MEMCACHE_MAX_BATCH_SIZE], namespace=namespace):
+                cached_data_result |= cached_data
             keys = keys[MEMCACHE_MAX_BATCH_SIZE:]
     except Exception as e:
         logging.error(f"""Failed to get keys form the memcache with {e=}""")
     for key, value in cached_data.items():
-        entity = Entity(Key(key))
+        entity = Entity(Key.from_legacy_urlsafe(key))
```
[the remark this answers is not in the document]
ArneGudermann: Yes, because there are converted here.
https://github.com/viur-framework/viur-core/blob/42f5afc4f23b1eb70d448a730db3a20e0e442f81/src/viur/core/db/cache.py#L96

phorward (approved):

Issue #1390, opened 2025-01-25
`DeferredTask` called with `SkeletonInstance`, but got `dict`


sveneberth:
A provided `SkeletonInstance` argument for a `DeferredTask` is converted into a dict and never converted back to a `SkeletonInstance`.

PR #1518, opened 2025-07-10
fix: `CallDeferred` with `SkeletonInstance`

ArneGudermann:
Fix for https://github.com/viur-framework/viur-core/issues/1390


src/viur/core/utils/json.py
```diff
@@ -47,7 +47,14 @@ def preprocess(obj: t.Any) -> t.Any:
             return tuple(ViURJsonEncoder.preprocess(value) for value in obj)
 
         elif hasattr(obj, "__class__") and obj.__class__.__name__ == "SkeletonInstance":  # SkeletonInstance
-            return {bone_name: ViURJsonEncoder.preprocess(obj[bone_name]) for bone_name in obj}
+            data = obj.dbEntity
+            if data is None:
+                data ={}
```
phorward, line 52:
```suggestion
            data = obj.dbEntity or {}
```

src/viur/core/utils/json.py
```diff
@@ -47,7 +47,14 @@ def preprocess(obj: t.Any) -> t.Any:
...
+            return {
+                ".__entity__": ViURJsonEncoder.preprocess(dict(data)),
```
phorward, line 54: `db.Entity` inherits from `dict`. I think this is obsolete, but please test.
```suggestion
                ".__entity__": ViURJsonEncoder.preprocess(data),
```

src/viur/core/utils/json.py
```diff
@@ -80,6 +87,15 @@ def _decode_object_hook(obj: t.Any):
         entity = db.Entity(db.Key.from_legacy_urlsafe(obj[".__key__"]) if obj[".__key__"] else None)
         entity.update(obj[".__entity__"])
         return entity
+    elif len(obj) == 3 and all(k in obj for k in (".__entity__", ".__key__", ".__skel__")):
+        from viur.core.skeleton.utils import skeletonByKind
+        skel = skeletonByKind(obj[".__skel__"])()
```
phorward, line 92: What if a subskel of a given kind was provided? This won't work.

phorward (changes requested):
I also stumbled across this about a year ago and decided at the time that this problem could not be solved.

It's precisely about these edge cases: What if, for example, a subskel or even ActionSkel with a different bone order is put in here? Then there is no child and a skeletonByKind also fails.

I would therefore like to open this topic for discussion.

PR #1534, opened 2025-07-31
fix: Missing `_atomic_dump()` for KeyBone and DateBone

ArneGudermann:
Add `_atomic_dump` for `DateBone` and `KeyBone` and `fix _create_diff` in History Module

### DIFF
--- a/src/viur/core/bones/date.py
+++ b/src/viur/core/bones/date.py
@@ -361,6 +361,9 @@ class DateBone(BaseBone):
         }
 
     def _atomic_dump(self, value):
-        if isinstance(value, datetime):
-            return value.isoformat()
-        return super()._atomic_dump(value)
+        if not value:
+            return None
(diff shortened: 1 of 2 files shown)


phorward (changes requested):

src/viur/core/bones/date.py
```diff
@@ -359,3 +359,8 @@ def structure(self) -> dict:
             "time": self.time,
             "naive": self.naive
         }
+
+    def _atomic_dump(self, value):
+        if isinstance(value, datetime):
+            return value.isoformat()
+        return super()._atomic_dump(value)
```
phorward, line 366:
```suggestion
        if not value:
            return None
        if not isinstance(value, datetime):
            raise ValueError("Expecting datetime object")

        return value.isoformat()
```

phorward (changes requested):

PR #1536, opened 2025-08-04
fix+refactor: `(Raw|String|Text)Bone\.getSearchTags`

phorward:
Implements getSearchTags() for `RawBone` as well, and lets `TextBone` and `StringBone` inherit from `RawBone` to avoid redundant code. Tag matching implemented via regex now.

### DIFF
--- a/src/viur/core/bones/raw.py
+++ b/src/viur/core/bones/raw.py
@@ -1,4 +1,7 @@
+import re
 from viur.core.bones.base import BaseBone, ReadFromClientError, ReadFromClientErrorSeverity
+
+SEARCH_TAGS = re.compile(r"\w+")
 
 
 class RawBone(BaseBone):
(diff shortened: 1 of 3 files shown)


phorward:
> That looks very promising. But I found a problem. When you store a string that looks like `(foo)` the searchtags are `{'41','foo','40'}` due to the HTML escape function. Now when you search for example for another item like `41` all items with the `(` in it pops up. I I think we should discuss that first.

Yes, but this is also the case in the current implementation.

I would like to ignore this "problem" for now, and fix it in ViUR4. IMHO, StringBone should contain any single-line string and TextBone any multi-line string. HtmlBone could be a bone (configurable for single- or multi-line) to escape HTML consequently.

phorward:
Relates to #1539

PR #1543, opened 2025-08-06
fix: `bounce` check order in `clone`

ArneGudermann:
If we frist check the `fromClient` and the the `bounce` we send all all erorrs back to the client.


phorward:
> If we frist check the `fromClient` and the the `bounce` we send all all erorrs back to the client.

Yes, that is a wanted behavior, and you cannot change this now for clone only and keep it for add or edit. Therefore, this pull request is not accepted.

If this has to be done with https://github.com/viur-framework/vi-admin/issues/49#issuecomment-3156402674, I'm in contact with @akelch and we generally have to find another way around this situation or problem.

PR #1544, opened 2025-08-06
refactor+fix: `bounce`-flag and amend-handling

phorward:
Refactors the `bounce`-flag to the following:

- `bounce` is now a bool parameter to all endpoints
- `BaseSkeleton.fromClient(amend=True)` now only outputs errors for provided fields with wrong data
- The behavior is almost the same then before (100% backward compatible)
- add/edit/clone-endpoints won't accept "throw-away-*args" anymore (never used in the wild)

Relates to https://github.com/viur-framework/vi-admin/issues/49#issuecomment-3156402674, replacement for #1543.

### DIFF
--- a/src/viur/core/skeleton/meta.py
+++ b/src/viur/core/skeleton/meta.py
@@ -411,6 +411,8 @@ class BaseSkeleton(object, metaclass=MetaBaseSkel):
                                 f"""{getattr(cls, "kindName", cls.__name__)}: {".".join(error.fieldPath)}: """
                                 f"""({error.severity}) {error.errorMessage}"""
                             )
+                    else:
+                        errors.clear()
 
                 skel.errors += errors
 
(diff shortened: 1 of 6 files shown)


src/viur/core/skeleton/meta.py
```diff
@@ -411,6 +411,8 @@ def fromClient(
                                 f"""{getattr(cls, "kindName", cls.__name__)}: {".".join(error.fieldPath)}: """
                                 f"""({error.severity}) {error.errorMessage}"""
                             )
+                    else:
+                        errors.clear()
```
phorward, line 415: This is the only confusing line. It just ensures that bones which are not provided simply don't output any errors, according to the `amend`-rules. A bone with a concrete errors (invalid value submit) will always be reported.

ArneGudermann (dismissed):

PR #1552, opened 2025-08-25
fix: `DatabaseAdapter` and `HistoryAdapter`

phorward:
- `DatabaseAdapter` now accepts the `update_relations`-flag to determine for the
- `HistoryAdapter` if entries shall be written or not during a refresh or update_relations, which currently is the case. Anyway, this solution maybe does not satisfy all cases, which could only be determined using a special "write_history"-flag provided to `skel.write()`...

This PR is breaking, because custom DatabaseAdapters must accept the `update_relations`-flag or `**kwargs`.

### DIFF
--- a/src/viur/core/bones/base.py
+++ b/src/viur/core/bones/base.py
@@ -916,7 +916,7 @@ class BaseBone(object):
             elif self.multiple:
                 res = []
 
-                assert newVal is None or isinstance(newVal, (list, tuple)), \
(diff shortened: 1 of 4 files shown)


phorward:
yes, I might discuss that again on Friday... because, as I said, I'm not quite sure if that's right. because sometimes we have the case that a dataset is deliberately rewritten by an action triggered by a user and if the patch or write is then executed with update_relations=False. that wouldn't show up in the history entries, although it should

PR #1561, opened 2025-09-08
fix: ActionSkel paradigm for `user.UserPassword.pwrecover`

phorward:
- Fixes the password recovery process via `UserPassword.pwrecover`
- Fixes HTML renderer's render.render to accept `None` skels
- Cleans-up some unused template variables


ArneGudermann (approved):
No changes in this https://github.com/viur-framework/viur-base/blob/main/deploy/html/user_passwordrecover.html template ?

phorward:
> No changes in this https://github.com/viur-framework/viur-base/blob/main/deploy/html/user_passwordrecover.html template ?

Yes. Best way to use it is to use `renderEditSkel` facilities, instead of a static template. This is the sense of the ActionSkel paradigm.

PR #1564, opened 2025-09-11
fix: Not execute `_route` on preflight requests

ArneGudermann:
When you call a list with the `x-viur-bonelist` header, the browser starts a preflight request. This request is sent without cookies, so if you have an access check in the list filter, the core response is a 401 error, and the browser thinks it cannot send the header.

### DIFF
--- a/src/viur/core/request.py
+++ b/src/viur/core/request.py
@@ -364,8 +364,10 @@ class Router:
 
             if conf.request_preprocessor:
                 path = conf.request_preprocessor(path)
-
-            self._route(path)
+            if self.method != "options":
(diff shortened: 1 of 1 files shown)


phorward (changes requested):
Hello @ArneGudermann,

I think this solution is wrong, because inside of `self._route()`, there are several more tests on `self.method == "options"` which won't be accessible anymore by your change.

Please make a deeper determination of the problem, maybe exceptions must be handled differently in pre-flight requests.

phorward:
@ArneGudermann is this pull request still an issue? I'm moving it to v3.9 for now.

Issue #1585, opened 2025-10-16
Bring back `keys_only=True` feature


phorward:
Here is an example for a function making use of the "keys_only"-feature from datastore.
This could probably be generalized for viur-core again.
```py
    @exposed
    @access("root")
    @as_json_response()
    def export_keys(self, cursor=None, group=None):
        query = db.transport.__client__.query(kind="variant")
        if group:
            f = datastore.query.PropertyFilter("group", "=", group)
            query.add_filter(filter=f)
        query.keys_only()
        query_res = query.fetch(limit=100, start_cursor=cursor)
        entities = list(query_res)
        result = list()
        for entity in list(entities):
            result.append(entity.key.to_legacy_urlsafe().decode("utf8"))
        if query_res.next_page_token:
            logging.debug(f"export_keys: {result=}, {query_res.next_page_token.decode("utf8")=}")
            return result, query_res.next_page_token.decode("utf8")
        else:
            return result, None
```

PR #1586, opened 2025-10-17
feat: Add `keys_only`-tooling for `Query`

ArneGudermann:
Resolves: https://github.com/viur-framework/viur-core/issues/1585

### DIFF
--- a/src/viur/core/db/query.py
+++ b/src/viur/core/db/query.py
@@ -652,7 +652,7 @@ class Query(object):
         if res:
             if keys_only:
                 res = [
-                    obj if isinstance(obj, Key) else obj.key 
+                    obj if isinstance(obj, Key) else obj.key
                     for obj in res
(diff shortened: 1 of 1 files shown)


src/viur/core/db/query.py
```diff
@@ -641,9 +644,18 @@ def run(self, limit: int = -1) -> t.List[Entity]:
...
+                keys_only
+            ))
 
         if res:
+            if keys_only:
+                res = [
+                    obj if isinstance(obj, Key) else obj.key 
+                    for obj in res
+                    if isinstance(obj, (Entity, Key))
+                ]
```
phorward:
```suggestion
            if keys_only:
                res = [
                    obj if isinstance(obj, Key) else obj.key
                    for obj in res
                    if isinstance(obj, (Entity, Key))
                ]
```

Issue #29, opened 2020-06-16
differentiate between Unauthorized and Forbidden


sveneberth:
In several placed in core (especially in the edit-methods) we are currently raising an Unauthorized exception.
But unauthorized means you have submitted no or invalid credentials - If you are logged in and have not the move-/edit-access it's simply just forbidden to you.
So we have to between not logged in (Unauthorized) and no access (Forbidden).

sveneberth:
Just a draft of how it could be handled in the future:

```
def can(self, action: str) -> bool:
    if not (user := current.user.get()):
        return Reason.Deny("not loggin'ed", 401)

    if user and user["access"]:
        if not ("root" in user["access"] or f"{self.moduleName}-{action]" in user["access"]):
            return Reason.Deny("missing permissions: root, action-edit", 403)

    if user and not user["has_2fa_enabled"]:
        return Reason.Deny("you have to setup 2FA!")

    return Reason.Allow("You're nice and welcome!")

```

PR #1594, opened 2025-11-06
fix: list: Distinguish between 401 and 403

phorward:
Currently, any listFilter that returns None raises error 401, althought a user is authorized. This fix raises error 403 forbidden when an authenticated user is not allowed to proceed.

Belongs to #29


sveneberth:
This solves partly #29. But it's more a hack than a solution. Can we generalize this and use in on ever permission check? It should be consistent in edit/move/etc.

PR #1604, opened 2025-11-25
fix: Check if the index is None in RecordBone postDeletedHandler

ArneGudermann:


phorward (changes requested):
Hi, the fix is wrong, I've added a suggestion.
We did a similar fix already in #1591.
In case of None, index 0 shall be kept. This is also useful when switching single to multiple bones.

Issue #1622, opened 2026-01-26
`AttributeError: 'SkeletonInstance' object has no attribute 'shop_price_'` fails with viur-core==3.8.15


phorward:
Hello,
due internal changes in viur-core 3.8.14 and 3.8.15, a computed bone in a RefSkel does not automatically enforce loading the underlying entire skeleton, as before, which now yields into this error
```
[2026-01-26 22:26:04,255] /.../viur/core/request.py:400 [ERROR] ViUR has caught an unhandled exception!
[2026-01-26 22:26:04,256] /.../viur/core/request.py:401 [ERROR] 'SkeletonInstance' object has no attribute 'shop_price_'
Traceback (most recent call last):
  File "~/myproject/.venv/lib/python3.13/site-packages/viur/core/skeleton/instance.py", line 254, in __getattr__
    return self.boneMap[item]
           ~~~~~~~~~~~~^^^^^^
KeyError: 'shop_price_'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "~/myproject/.venv/lib/python3.13/site-packages/viur/core/request.py", line 372, in _process
    self._route(path)
    ~~~~~~~~~~~^^^^^^
  File "~/myproject/.venv/lib/python3.13/site-packages/viur/core/request.py", line 657, in _route
    res = caller(*self.args, **kwargs)
  File "~/myproject/.venv/lib/python3.13/site-packages/viur/core/module.py", line 245, in __call__
    return self._func(self._instance, *args, **kwargs)
           ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "~/myproject/.venv/lib/python3.13/site-packages/viur/shop/modules/api.py", line 340, in cart_list
    child = self.json_renderer.renderSkelValues(child_skel)
  File "~/myproject/.venv/lib/python3.13/site-packages/deprecated/classic.py", line 199, in wrapper_function
    return wrapped_(*args_, **kwargs_)
  File "~/myproject/.venv/lib/python3.13/site-packages/viur/core/render/json/default.py", line 81, in renderSkelValues
    return skel.dump()
           ~~~~~~~~~^^
  File "~/myproject/.venv/lib/python3.13/site-packages/viur/core/skeleton/instance.py", line 356, in dump
    bone_name: bone.dump(self, bone_name) for bone_name, bone in self.items()
               ~~~~~~~~~^^^^^^^^^^^^^^^^^
[...]

PR #1619, opened 2026-01-28
fix: `shortkey` computation needs `Key` instance

sveneberth:
Does not work with renderPreparation.

Fixes #1618

PR #1623, opened 2026-02-05
fix: Propagate `@property` from Skeleton to `SkeletonInstance` that belongs to a `RefSkel` + reset `renderPreparation`

sveneberth:
- A `@property` member can now be accessed on a `SkeletonInstance` belongs to a `RefSkel`. Before loading `class_value` the type of `skeletonCls` will be evaluated.
- Before calling the compute or `@property` function with the help of `without_render_preparation`, it will be ensured that the `SkeletonInstance` is called without any `renderPreparation`. This avoids for example providing the key string-encoding instead of the `db.Key` instance.
- Introduce `without_render_preparation` (known from [viur-toolkit](https://github.com/viur-framework/viur-toolkit/blob/1d95b1a8ee5fb40e0847094821ff6e424891376d/src/viur/toolkit/viur.py#L38-L65))
- Introduce `is_skeletoninstance_of` for handy checks (instead of this https://github.com/viur-framework/viur-shop/blob/2a55ef1aa2fdb00d6a6af3985cfe151c5d6aedb6/src/viur/shop/types/dc_scope.py#L243-L257)

---

Fixes https://github.com/viur-framework/viur-core/issues/1622
Alternative solution to #1617
Also replaces #1619

### DIFF
--- a/src/viur/core/bones/base.py
+++ b/src/viur/core/bones/base.py
@@ -1555,7 +1555,7 @@ class BaseBone(object):
 
     def _compute(self, skel: 'viur.core.skeleton.SkeletonInstance', bone_name: str):
         """Performs the evaluation of a bone configured as compute"""
-        from ..skeleton.utils import without_render_preparation, remove_render_preparation_deep
+        from ..skeleton.utils import without_render_preparation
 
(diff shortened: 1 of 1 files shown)


phorward (changes requested):
Minifix...

PR #1631, opened 2026-02-18
fix: Call `Tree.canAdd` with `SkeletonInstance` in `Tree.clone`

### LINKED ISSUE #1602: clone: `canAdd` must be called with a SkeletonInstance
The second parameter `parentNodeSkel` of `canAdd` must be a `SkeletonInstance`
https://github.com/viur-framework/viur-core/blob/4c006c8777fae7bd6893f0e0f1f396f4fa498132/src/viur/core/prototypes/tree.py#L836

But in `clone` a string is provided: 
https://github.com/viur-framework/viur-core/blob/4c006c8777fae7bd6893f0e0f1f396f4fa498132/src/viur/core/prototypes/tree.py#L758

sveneberth:
Must be a `SkeletonInstance`, not a string-encoded key.

Fixes #1602

### DIFF
--- a/src/viur/core/prototypes/tree.py
+++ b/src/viur/core/prototypes/tree.py
@@ -1,15 +1,16 @@
-import time
 import logging
+import time
 import typing as t
+
 from deprecated.sphinx import deprecated
-from viur.core import errors, db, current
-from viur.core.decorators import *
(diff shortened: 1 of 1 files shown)


phorward (changes requested):
Improve clone-function signature and use keyword-arguments to can*.

PR #1633, opened 2026-02-20
fix: `utils.__getattr__` doesn't return callables

phorward:
Alternative replacement for #1632


phorward:
Closing this PR again in favor of #1632.

PR #1637, opened 2026-02-20
fix: Call postSavedHandler on UsingSkel (RelSkel)

sveneberth:
The bug is not new, existed already in at least viur-core 3.7.

Calling the `postSavedHandler` ensures derives are generated (or their generation is been triggered) in `ImageBones` inside a `RelSkel` of a `RelationalBone`.


src/viur/core/bones/relational.py
```diff
@@ -1180,8 +1185,11 @@ def tuple_check(in_value: tuple | None = None) -> bool:
                 raise ValueError(f"You must supply a tuple of (Database-Key, relSkel) to {boneName}")
             parsed_value = value
         elif self.multiple and not self.using:
-            if not isinstance(value, (str, int, db.Key)) and not (isinstance(value, list)) \
-                    and all([isinstance(val, (str, int, db.Key)) for val in value]):
+            if (
+                not isinstance(value, (str, int, db.Key))
```
phorward: same here

PR #1647, opened 2026-03-11
refactor: `db.fix_unindexable_properties`

sveneberth:
### 1. The length of a non-ASCII-only string says nothing about its size

Example: 
```py
import sys, string

x = (string.hexdigits * 100)[:1500]
print(len(x), sys.getsizeof(x))
# 1500 1521

y = ((string.hexdigits + "äöüèô\u2013") * 100)[:1500]
print(len(y), sys.getsizeof(y))
# 1500 3030
```
Same length, different size (bytes).
See https://python-fiddle.com/saved/a9d60757-a1e4-47a3-9a6d-d93dc781b396

That's why we should use `sys.getsizeof` here.

### 2. Replace vs. update

Why are the properties replaced and not just supplemented? It may have been a conscious decision to exclude properties, but if they are less than 1500 bytes, this method re-indexes the property. I find that strange. I have therefore added the parameter `keep_exclusions`. 
We can **discuss** here whether it should be `True` by default — I would prefer this!

### DIFF
--- a/src/viur/core/db/transport.py
+++ b/src/viur/core/db/transport.py
@@ -44,6 +44,8 @@ def get(keys: t.Union[Key, t.List[Key]]) -> t.Union[t.List[Entity], Entity, None
     :param keys: A datastore key (or a list thereof) to lookup
     :return: The entity (or None if it has not been found), or a list of entities.
     """
+    if conf.debug.trace_queries:
+        logging.info(f"Calling db.get({keys=})")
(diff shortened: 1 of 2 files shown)


phorward (dismissed):
This is a good improvement, and I would prefer keep the `keep_exclusions` default to True. I think there is no discussion required, but we can talk about it shortly.

phorward (approved):

PR #1659, opened 2026-03-30
refactor: Provide and use `_can_access`-function in vi-render

phorward:
The new function is the only "place of truth" who may access content via the vi-render. It is a wanted change, that there is an explicit check for "admin"-access, rather than "root" or "admin".

This PR is an add-on to #1620.


phorward:
> > It is a wanted change, that there is an explicit check for "admin"-access, rather than "root" or "admin"
> 
> Why? My understanding of a _root_ or _superuser_ is that they can do anything — regardless of whether they have additional individual permissions.

I've switched it back with 0e5ded7a9536b5635fd6231ae9b68241f709de53 to the old behavior, is this ok for you?

sveneberth (approved):

PR #1666, opened 2026-04-16
feat: `NumericBone` with decimal-support

### LINKED ISSUE #1099: Implement decimal mode in `NumericBone`
for prices and values with digits a fixed-point arithmetic would be nice. https://docs.python.org/3/library/decimal.html#module-decimal

KadirBalku:
Fixes #1099 

- Add Decimal mode to NumericBone
- Decimal support for JSON renderer
- Integration tests

### DIFF
--- a/src/viur/core/bones/numeric.py
+++ b/src/viur/core/bones/numeric.py
@@ -3,6 +3,7 @@ import numbers
 import sys
 import typing as t
 import warnings
+from decimal import Decimal, ROUND_HALF_UP, InvalidOperation
 
 from viur.core import db, i18n
 from viur.core.bones.base import BaseBone, ReadFromClientError, ReadFromClientErrorSeverity
@@ -33,6 +34,7 @@ class NumericBone(BaseBone):
         min: int | float = MIN,
         max: int | float = MAX,
         precision: int = 0,
+        decimal: bool = False,
         mode=None,  # deprecated!
         **kwargs
     ):
@@ -42,6 +44,7 @@ class NumericBone(BaseBone):
         :param min: Minimum accepted value (including).
         :param max: Maximum accepted value (including).
         :param precision: How may decimal places should be saved. Zero casts the value to int instead of float.
+        :param decimal: If True, use decimal.Decimal internally for exact arithmetic.
         """
         super().__init__(**kwargs)
 
@@ -61,6 +64,9 @@ class NumericBone(BaseBone):
         self.precision = precision
         self.min = min
         self.max = max
+        self.decimal = decimal
+        if decimal:
+            self._quantize_exp = Decimal(10) ** -precision
 
     def __setattr__(self, key, value):
         """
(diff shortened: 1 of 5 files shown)


src/viur/core/bones/numeric.py
```diff
@@ -3,6 +3,7 @@
 import sys
 import typing as t
 import warnings
+from decimal import Decimal, ROUND_HALF_UP, InvalidOperation
```
phorward:
```suggestion
import decimal
```
Can you please import just decimal, and use full-qualified names? `decimal.Decimal`, `decimal.ROUND_HALF_UP` and `decimal.InvalidOperation`.

src/viur/core/bones/numeric.py
```diff
@@ -80,18 +86,76 @@ def __setattr__(self, key, value):
...
+        decimal_name = f"{name}_decimal"
+        if self.multiple and isinstance(raw_values, list):
+            skel.dbEntity[decimal_name] = [
+                str(self._to_decimal(v)) if v is not None else None
+                for v in raw_values
+            ]
+        elif raw_values is not None:
+            skel.dbEntity[decimal_name] = str(self._to_decimal(raw_values))
+        else:
+            skel.dbEntity[decimal_name] = None
```
phorward: This code path completely ignores `languages`. It is generally bad-practice to work around the usual way of serialize/deserialize. Please use a utility function like [iter_bone_value](https://core.docs.viur.dev/en/stable/viur/core/bones/base/index.html#core.bones.base.BaseBone.iter_bone_value) to iterate trough all combination of a bone value effectively.

src/viur/core/bones/numeric.py
```diff
@@ -80,18 +86,76 @@ def __setattr__(self, key, value):
...
+        skel.dbEntity.exclude_from_indexes.add(decimal_name)
+
+        return True
+
+    def unserialize(self, skel: "SkeletonInstance", name: str) -> bool:
+        if self.decimal:
+            decimal_name = f"{name}_decimal"
+            if decimal_name in skel.dbEntity:
+                # Prefer exact string representation over float
+                skel.dbEntity[name] = skel.dbEntity[decimal_name]
```
phorward: This code path ignores `languages` as well.

phorward (changes requested):
@ArneGudermann Could you please take my previous comments on this PR into account to make the code more readable and clearer?

src/viur/core/bones/numeric.py
```diff
@@ -42,6 +44,7 @@ def __init__(
         :param min: Minimum accepted value (including).
         :param max: Maximum accepted value (including).
         :param precision: How may decimal places should be saved. Zero casts the value to int instead of float.
+        :param decimal_mode: If True, use decimal.Decimal internally for exact arithmetic.
```
phorward:
```suggestion
        :param decimal: If True, use decimal.Decimal internally for exact arithmetic.
```

src/viur/core/bones/numeric.py
```diff
@@ -33,6 +34,7 @@ def __init__(
         min: int | float = MIN,
         max: int | float = MAX,
         precision: int = 0,
+        decimal_mode: bool = False,
```
phorward: Can we just call it "decimal" here, same as the property and how it's described below?

```suggestion
        decimal: bool = False,
```

src/viur/core/bones/numeric.py
```diff
@@ -61,6 +64,9 @@ def __init__(
         self.precision = precision
         self.min = min
         self.max = max
+        self.decimal = decimal_mode
+        if decimal_mode:
+            self._quantize_exp = decimal.Decimal(10) ** -precision
```
phorward:
```suggestion
        self.decimal = decimal
        self._quantize_exp = decimal.Decimal(10) ** -precision if decimal else None
```

phorward (changes requested):

src/viur/core/bones/numeric.py
```diff
@@ -80,16 +86,42 @@ def __setattr__(self, key, value):
...
+            if isinstance(value, (int, float)):
+                return deci.Decimal(str(value)).quantize(self._quantize_exp)
+        raise ValueError(f"Cannot convert {type(value).__name__} to Decimal")
+
     def singleValueUnserialize(self, val):
         if val is not None:
             try:
+                if self.decimal:
+                    if isinstance(val, dict) and "decimal" in val:
+                        return self._convert_to_decimal(val["decimal"])
```
phorward:
```suggestion
                        val = val["decimal"]
```

src/viur/core/render/json/default.py
```diff
@@ -20,7 +22,18 @@ class CustomJsonEncoder(json.JSONEncoder):
 
     def default(self, o: t.Any) -> t.Any:
 
-        if isinstance(o, translate):
+        if isinstance(o, Decimal):
+            if "json.decimal.as_string" in conf.compatibility:
```
phorward: Should we use str-representation generally for decimal-values? The viur-meeting decided about that? This is more consistent.

phorward (changes requested):

phorward (approved):

PR #1679, opened 2026-04-28
fix: also unescape short-form HTML numeric entities in download filename

sveneberth:
Filenames can contain `&#40;` (short form, no leading zero) in addition to `&#040;`, both represent `(`. Add the short-form variants for `(`, `)` and `=`.


src/viur/core/modules/file.py
```diff
@@ -659,6 +659,7 @@ def create_download_url(
 
         # Undo escaping on ()= performed on fileNames
         filename = filename.replace("&#040;", "(").replace("&#041;", ")").replace("&#061;", "=")
+        filename = filename.replace("&#40;", "(").replace("&#41;", ")").replace("&#61;", "=")
```
phorward: Can't we simply use `utils.string.unescape()` for that?

```suggestion
        filename = utils.string.unescape(filename)
```

It should match both the long- and short-form entities.

phorward (changes requested):

src/viur/core/modules/file.py
sveneberth: Good point.
However, since https://github.com/viur-framework/viur-core/blob/12e496f10fbd8d14b0df6c6b7cf1ce62b17df3dd/src/viur/core/modules/file.py#L106
uses `html.unescape`, we should use it here as well to be consistent

ArneGudermann (approved):

phorward (approved):

PR #1685, opened 2026-05-05
chore: Update minimum dependency to Python 3.13

phorward:
Also removes Python 3.12 compatibility code.

### DIFF
--- a/src/viur/core/config.py
+++ b/src/viur/core/config.py
@@ -21,11 +21,7 @@ if t.TYPE_CHECKING:  # pragma: no cover
     from viur.core.tasks import CustomEnvironmentHandler
     from viur.core import i18n
 
-# Construct an alias with a generic type to be able to write Multiple[str]
-# TODO: Backward compatible implementation, refactor when viur-core
-#       becomes >= Python 3.12 with a type statement (PEP 695)
-_T = t.TypeVar("_T")
-Multiple: t.TypeAlias = list[_T] | tuple[_T] | set[_T] | frozenset[_T]  # TODO: Refactor for Python 3.12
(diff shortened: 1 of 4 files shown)


src/viur/core/config.py
```diff
@@ -21,11 +21,7 @@
     from viur.core.tasks import CustomEnvironmentHandler
     from viur.core import i18n
 
-# Construct an alias with a generic type to be able to write Multiple[str]
-# TODO: Backward compatible implementation, refactor when viur-core
-#       becomes >= Python 3.12 with a type statement (PEP 695)
```
sveneberth, line 26: This code run with python 3.12 or higher. It could have been used already.

Is there a reason to ban 3.12?

sveneberth (changes requested):

src/viur/core/config.py
phorward: For what reason should we still support Python 3.12?

phorward: For what reason should we still support Python 3.12?

sveneberth: In general, a minimum version in a library should be introduced by a concrete technical requirement — otherwise it just unnecessarily restricts users who run different environments where 3.12 is still the standard.
                                                                                                                                                                                                                             
PEP 695 (type statement) is available since Python 3.12, so there's no technical reason to require 3.13 for this refactoring. The original TODO even said "refactor when viur-core becomes >= Python 3.12" — that threshold was already met.                                                                                                                                                                                                          
                  
Raising the minimum to 3.13 without a specific 3.13-only feature driving it makes viur-core incompatible with setups that are perfectly capable of running it.

PR #1689, opened 2026-05-11
feat: Add `vi/routes`-endpoint

phorward:
This endpoint can be used to retrieve a list of all available exposed endpoints in a ViUR system. Its mandatory for debug purposes and additional security check tools, like the viur-crawler project.

The endpoint can only be used by users who may access the vi-renderer.


sveneberth:
Does this work recursive (for nested modules)?

phorward:
> Does this work recursive (for nested modules)?

yes. I've checked it up with a project using viur-shop.

ArneGudermann (approved):

PR #1700, opened 2026-06-02
feat: Replace client-side MultiQuery with native Datastore operators

### LINKED ISSUE #1569: `default_order` is not applied to MultiQueries
Having a listFilter() that modifies the query to become a MultiQuery, any default_order is not applied. Maybe this a general bug with MultiQueries and orders.

ArneGudermann:
This PR replace client-side MultiQuery with native Datastore operators add or_filter()

  IN, != and NOT_IN filters are no longer split into multiple sub-queries
  and merged client-side. They are now passed as native PropertyFilter
  operators directly to Cloud Datastore, reducing RPC calls from N to 1
  for IN queries and fixing default_order not being applied to these
  queries (fixes #1569).

  New Query.or_filter() method allows expressing OR composite filters
  that are sent to Datastore as Or([PropertyFilter, ...]) groups.
  Multiple or_filter() calls are AND-ed together with each other and
  with any regular filter() calls.

  - query.py: remove MultiQuery split logic for IN/!=/NOT_IN; add or_filter() method; extend _entryMatchesQuery for IN/!=/NOT_IN/OR
  - transport.py: handle IN/!=/NOT_IN natively; send Or composites
  - types.py: add TOrFilters type alias and or_filters field to QueryDefinition
  - tasks.py: guard against serializing queries with OR filters

### DIFF
--- a/src/viur/core/db/query.py
+++ b/src/viur/core/db/query.py
@@ -16,6 +16,7 @@ from .types import (
     SortOrder,
     TFilters,
     TOrders,
+    TOrFilters,
     Key
 )
 from . import utils
@@ -29,7 +30,11 @@ TFilterHook = t.TypeVar("TFilterHook", bound=t.Callable[
 ])
 
 
-def _entryMatchesQuery(entry: Entity, singleFilter: dict) -> bool:
+def _entryMatchesQuery(
+    entry: Entity,
+    singleFilter: dict,
+    or_filters: TOrFilters | None = None,
+) -> bool:
     """
     Utility function which checks if the given entity could have been returned by a query filtering by the
     properties in singleFilter. This can be used if a list of entities have been retrieved (e.g. by a 3rd party
@@ -37,6 +42,7 @@ def _entryMatchesQuery(entry: Entity, singleFilter: dict) -> bool:
     :meth:`viur.core.prototypes.list.listFilter` method.
     :param entry: The entity which will be tested
     :param singleFilter: A dictionary containing all the filters from the query
+    :param or_filters: Optional list of OR groups; each group is a list of (filterStr, value) pairs
     :return: True if the entity could have been returned by such an query, False otherwise
     """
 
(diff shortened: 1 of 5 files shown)


src/viur/core/db/query.py
```diff
@@ -238,55 +260,82 @@ def filter(self, prop: str, value: DATASTORE_BASE_TYPES | list[DATASTORE_BASE_TY
...
-            else:  # IN filter
-                if not isinstance(value, (list, tuple)):
-                    raise ValueError("Value must be list or tuple if using IN filter!")
-                for val in value:
-                    newFilter = copy.deepcopy(origQuery)
-                    newFilter.filters[f"{field} ="] = val
-                    self.queries.append(newFilter)
+
+        # Normalize to uppercase for native Datastore operators passed as lowercase
+        op = op.upper() if op.lower() in {"in", "!=", "not_in"} else op
```
phorward: there is no lower-case "!="...
```suggestion
        op = op.upper() if op.lower() in {"in", "not_in"} else op
```

src/viur/core/db/query.py
```diff
@@ -238,55 +260,82 @@ def filter(self, prop: str, value: DATASTORE_BASE_TYPES | list[DATASTORE_BASE_TY
...
+        if self.queries is None:
+            return self
+
+        parsed = []
+        for prop, value in conditions:
+            if " " not in prop:
+                field, op = prop, "="
+            else:
+                field, op = prop.split(" ", 1)
+            op = op.upper() if op.lower() in {"in", "!=", "not_in"} else op
```
phorward:
```suggestion
            op = op.upper() if op.lower() in {"in", "not_in"} else op
```

phorward: Is a general `op = op.upper()` not sufficient?

phorward (changes requested):
Nice, can you please review the op.upper()-stuff being quite confusing.

phorward (approved):

PR #1704, opened 2026-06-09
fix: propagate user-entity changes into active sessions on edit

ArneGudermann:
The session-update loop introduced in User.onEdited was a no-op due to
  three bugs: wrong Datastore kind ("user" instead of Session.kindName),
  filtering by raw db.Key instead of str, and never calling db.put().

  Replaces the broken loop with a new deferred function
  session.update_session_user() that correctly fetches the fresh user
  entity and writes it into all active sessions of that user, so
  access-right changes take effect immediately without requiring re-login.

### DIFF
--- a/src/viur/core/session.py
+++ b/src/viur/core/session.py
@@ -304,6 +304,29 @@ def killSessionByUser(user: t.Optional[t.Union[str, "db.Key", None]] = None):
     DeleteSessionsIter.startIterOnQuery(query)
 
 
+def update_session_user(user_key: "db.Key"):
+    """
+    Updates the cached user entity in all active sessions of the given *user*.
+
+    Whenever a user entity is modified (e.g. its access rights have changed), the copy
+    stored within the user's active sessions has to be refreshed as well; Otherwise,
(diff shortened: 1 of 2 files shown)


phorward (approved):

src/viur/core/session.py
```diff
@@ -304,6 +304,29 @@ def killSessionByUser(user: t.Optional[t.Union[str, "db.Key", None]] = None):
...
+
+    if not (user_entity := db.get(user_key)):
+        logging.warning(f"Cannot update sessions, {user_key=} not found")
+        return
+
+    for session_entity in db.Query(Session.kindName).filter("user =", str(user_key)).iter():
+        session_entity["data"]["user"] = user_entity
+        session_entity["data"] = db.fix_unindexable_properties(session_entity["data"])
+        session_entity.exclude_from_indexes = {"data"}
+        db.put(session_entity)
```
phorward: This is not running within a transaction.

phorward (changes requested):

phorward (approved):

sveneberth (approved):

PR #1705, opened 2026-06-10
feat:  `SpatialBone`  extend `setBoneValue` to accept `dict` with lat/lng aliases

ArneGudermann:
SpatialBone.setBoneValue now accepts dicts with lat and
  lng keys in addition to tuple/list. Also fixes a logic
  bug in the existing type check (and → or, == 2 → != 2).

### DIFF
--- a/src/viur/core/bones/spatial.py
+++ b/src/viur/core/bones/spatial.py
@@ -378,7 +378,8 @@ class SpatialBone(BaseBone):
 
         :param skel: Dictionary with the current values from the skeleton the bone belongs to
         :param boneName: The name of the bone that should be modified
-        :param value: The value that should be assigned. Its type depends on the type of the bone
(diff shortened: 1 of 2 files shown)


sveneberth (changes requested):
I like the dict syntax. But do we really need two or three possible key names for long/lat? Can't we just agree on a single unique identifier for each one—preferably the one that's also used by the renderer?!

ArneGudermann:
I thought that would cover all the bases.

sveneberth:
> I thought that would cover all the bases.

But isn't it easier to have a single version instead of several? That seems to contradict our goal of getting rid of all possible compatibility flags as quickly as possible.

phorward (approved):

sveneberth (approved):

PR #1711, opened 2026-06-16
Refactor/1013 securityheaders into config

### LINKED ISSUE #1013: Move `securityheaders` into `config`
With the introduction of the new config-class concept #833, the function set from securityheaders seems to become obsolete, and should be implemented as part of the [`Security`-ConfigType-class](https://github.com/viur-framework/viur-core/blob/develop/src/viur/core/config.py#L279).

skoegl:
## Summary
  - Relocate the `viur.core.securityheaders` functions onto `config.Security` as snake_case methods (add_csp_rule, extend_csp, enable_strict_transport_security,
  set_x_frame_options, …).
  - Drop the magic `"_headerCache"` dict key; build derived headers into private Security attributes via `Security.finalize()` (called from setup()), emit via
  `Security.update_response_headers()`.
  - `securityheaders.py` reduced to deprecating delegators (old camelCase signatures + `validReferrerPolicies` kept) → no breaking changes.
  - Add `conf.debug.trace_headers` (+ `trace_headers_redact`) to log request/response headers, sensitive headers redacted.
  - Docs + CHANGELOG + unit tests. 
  
  Closes #1013.
  
  ## Test Plan
  - [ ] `uv run --with ".[testing]" python -m unittest discover tests`
  - [ ] `flake8 src/`


skoegl:
## Merged `develop` (4984d21, 46fd125)

Two features landed on `develop` in `securityheaders.py` after this branch reduced the module to deprecating delegators, so the conflicts could not be resolved mechanically -- both were ported onto `config.Security`:

- **Reporting-Endpoints (#1757):** `set_reporting_endpoint()` is now a `Security` method. The `Reporting-Endpoints` header is emitted in `update_response_headers()`, the reporting validation (`_validate_reporting_config`) runs as part of `finalize()`.
- **CSP level 3 directives (#1714):** extended directive whitelist in `add_csp_rule()`; `report-to` is replaced instead of accumulated, same as `report-uri`.
- **`securityheaders.set_reporting_endpoint`** stays as a deprecating delegator like the other functions, since it is already public in v3.9.0.rc2. The private helpers (`_build_reporting_endpoints_header`, `_validate_reporting_config`) are gone from the module; they were never public.
- **`__init__.py` / `request.py`:** kept this branch's side -- `finalize()` and `update_response_headers()` already cover the develop-side validation/emission blocks.
- **`docs/adr/securityheaders.md`** (new on develop) still described `_headerCache` and the camelCase functions; updated to the `Security` methods.
- **Tests:** develop's `tests/test_securityheaders.py` was pointed at `conf.security` directly and then merged into the existing `tests/test_security_headers.py` (one file, four classes). Full suite: 687 tests, green. flake8 on touched files reports only findings that pre-exist on `develop`.

src/viur/core/securityheaders.py
```diff
@@ -1,426 +1,78 @@
...
-        conf.security.content_security_policy[enforceMode] = {}
-    if objectType in ("report-uri", "report-to"):
-        # Both directives take exactly one value; a second one would be ignored by the browser anyway
-        conf.security.content_security_policy[enforceMode][objectType] = [srcOrDirective]
-    else:
-        if objectType not in conf.security.content_security_policy[enforceMode]:
-            conf.security.content_security_policy[enforceMode][objectType] = []
-        if srcOrDirective not in conf.security.content_security_policy[enforceMode][objectType]:
-            conf.security.content_security_policy[enforceMode][objectType].append(srcOrDirective)
+def _deprecated(old: str, new: str) -> None:
```
sveneberth, line 14: Why use [`@deprecated`](https://pypi.org/project/Deprecated/) for this

src/viur/core/securityheaders.py
sveneberth, line 1: I don't understand the point of moving everything into the config.

1. Doesn't that just bloat the config?!
2. Aren't all methods just results of persistent configurations? `extend_csp`, for example, extends the current request, not the config — so it doesn't belong in the config.
3. The config serves as a value store for configurations, not as a controller or API — there are separate modules for that, such as this `securityheader` module.

CHANGELOG.md
sveneberth, line 1: Part of the release, not a PR. See https://github.com/viur-framework/viur-core/blob/main/CONTRIBUTING.md#releasing

src/viur/core/config.py
```diff
@@ -556,6 +873,14 @@ class Debug(ConfigType):
     disable_cache: bool = False
     """If set to true, the decorator @enableCache from viur.core.cache has no effect"""
 
+    trace_headers: bool = False
+    """If enabled, log the incoming request headers and the final outgoing response headers per request.
+    Sensitive headers are redacted (see :attr:`trace_headers_redact`)."""
+
+    trace_headers_redact: Multiple[str] = ("Authorization", "Proxy-Authorization", "Cookie", "Set-Cookie")
+    """Header names (matched case-insensitively) whose values are redacted in :attr:`trace_headers` output.
+    An empty collection disables redaction (full raw dump)."""
```
sveneberth, line 882: These are features. That's not part of a refactoring (btw: the PR does not follow the conventional commit syntax. The prefix should be `refactor: ...`.)

sveneberth (changes requested):

PR #1715, opened 2026-06-22
fix: Several `History` improvements

phorward:
As the `History` module is not in use yet, these improvements would help existing projects to port. It also decreases redundant storage of useless information.

### DIFF
--- a/src/viur/core/modules/history.py
+++ b/src/viur/core/modules/history.py
@@ -324,7 +324,7 @@ class HistoryAdapter(DatabaseAdapter):
             if kindname == "viur-history":
                 return None
 
-        return history_module.write_diff(
+        return history_module.write(
             action, old_skel, new_skel,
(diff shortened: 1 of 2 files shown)


sveneberth (dismissed):

src/viur/core/modules/history.py
```diff
@@ -495,10 +496,10 @@ def create_history_entry(
         that can either be written to datastore or another database.
         """
         skel = new_skel or old_skel
-        new_data = skel.dump()
+        new_data = skel.dump(bones=change_list) if skel else {}
 
         if change_list and old_skel != new_skel:
-            old_data = old_skel.dump()
+            old_data = old_skel.dump(bones=change_list)
```
phorward:
```suggestion
            old_data = old_skel.dump(bones=change_list) if old_skel else {}
```

PR #1717, opened 2026-06-29
feat: Configurable named datastore database and namespace

akelch:
Adds a supported way to point viur-core at a named Datastore database and/or a namespace via `conf.db`, instead of always binding to the project's `(default)` database at import time.

With nothing configured the behaviour is unchanged: the client is still built as a bare `datastore.Client()` and keys carry no database/namespace.

**`conf.db.name` / `conf.db.namespace` (config.py):**
Two new `conf.db` fields selecting the target database and namespace. Both default from the `VIUR_DB_NAME` / `VIUR_DB_NAMESPACE` environment variables.

**Client construction (db/transport.py):**
`__client__` is built with `database=` / `namespace=` taken from `conf.db`. When neither is set the call collapses to the historical parameterless `datastore.Client()`.

**Key alignment (db/types.py):**
`Key.__init__` defaults `database` / `namespace` from the client (an explicit caller argument still wins); a mismatch otherwise makes Datastore reject the request as a cross-database access. `Key.to_legacy_urlsafe` drops the database id before encoding — the upstream implementation refuses any key carrying one, while `str(key)` and session handling hit it constantly.

**Boot banner (__init__.py):**
The local dev-server banner gains a `database` / `namespace` line whenever a non-default target is configured, so it is never unnoticed which datastore viur-core is wired to.

### DIFF
--- a/src/viur/core/__init__.py
+++ b/src/viur/core/__init__.py
@@ -245,6 +245,16 @@ def __build_app(modules: ModuleType | object, renderers: ModuleType | object, de
     conf.main_app = index
 
 
+def _datastore_banner_lines() -> list[str]:
+    """Boot-banner lines for a non-default datastore target; empty for default."""
+    extra = []
+    if conf.db.name:
+        extra.append(f"""database = \033[1;33m{conf.db.name}\033[0m""")
+    if conf.db.namespace:
+        extra.append(f"""namespace = \033[1;33m{conf.db.namespace}\033[0m""")
+    return extra
+
+
 def setup(modules:  ModuleType | object, render:  ModuleType | object = None, default: str = "html"):
     """
         Define whats going to be served by this instance.
@@ -334,6 +344,7 @@ def setup(modules:  ModuleType | object, render:  ModuleType | object = None, de
             f"""project = \033[1;31m{conf.instance.project_id}\033[0m""",
             f"""python = \033[1;32m{".".join((str(i) for i in PYTHON_VERSION))}\033[0m""",
             f"""viur = \033[1;32m{".".join((str(i) for i in conf.version))}\033[0m""",
+            *_datastore_banner_lines(),  # only when a non-default db/namespace is set
             ""  # empty line
         )
 
--- a/src/viur/core/db/transport.py
+++ b/src/viur/core/db/transport.py
@@ -16,7 +16,16 @@ from viur.core.errors import HTTPException
 datastore.helpers.key_from_protobuf = key_from_protobuf
 datastore.helpers.entity_from_protobuf = entity_from_protobuf
 
-__client__ = datastore.Client()
+# Built once at import, kept for the process lifetime — so db/namespace have to
+# come from env (via conf.db); nothing can retarget the client afterwards.
+# Empty kwargs == datastore.Client(): no change for default deployments.
+__client_kwargs__: dict[str, str] = {}
+if conf.db.name:
+    __client_kwargs__["database"] = conf.db.name
+if conf.db.namespace:
+    __client_kwargs__["namespace"] = conf.db.namespace
+
+__client__ = datastore.Client(**__client_kwargs__)
 
 
 def allocate_ids(kind_name: str, num_ids: int = 1, retry=None, timeout=None) -> list[Key]:
--- a/src/viur/core/db/types.py
+++ b/src/viur/core/db/types.py
@@ -63,14 +63,34 @@ class Key(Datastore_key):
                 id_or_name = int(id_or_name)
             new_path_args.extend((kind, id_or_name))
 
+        from .transport import __client__  # noqa: E402 # import works only here because circular imports
+
         if project is None:
-            from .transport import __client__  # noqa: E402 # import works only here because circular imports
             project = __client__.project
+
+        # Keys must match the client's db/namespace or Datastore rejects the
+        # request as cross-database. Default from the client; caller wins.
+        if __client__.database:
+            kwargs.setdefault("database", __client__.database)
+        if __client__.namespace:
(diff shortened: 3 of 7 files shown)


src/viur/core/db/types.py
```diff
@@ -63,15 +63,35 @@ def __init__(self, *path_args, project: str | None = None, **kwargs):
...
 
     def __str__(self):
         return self.to_legacy_urlsafe().decode("ASCII")
 
+    def to_legacy_urlsafe(self, location_prefix=None):
+        # Upstream to_legacy_urlsafe() rejects keys carrying a database, but
+        # str(key)/session paths hit it constantly. Drop the db for encoding —
+        # unambiguous to restore since the process talks to a single database.
+        if self._database is None:
+            return super().to_legacy_urlsafe(location_prefix=location_prefix)
```
phorward: This path could be omitted; an None-set self._database is always set back to None.
```suggestion
```

phorward (approved):

sveneberth:
This PR is similar to https://github.com/viur-framework/viur-core/pull/1698 but – as far as I understood – takes a different approach.
Here are the keys database-aware. Different database means different legacy key.
In my approach the keys are database-agnostic. This allows you to migrate data in a copied database and swap it in the same project in produciton.
We need to discuss here our use-cases before merging!

sveneberth (changes requested):
Discuss requirements first

src/viur/core/db/transport.py
```diff
@@ -16,7 +16,16 @@
...
+# Built once at import, kept for the process lifetime — so db/namespace have to
+# come from env (via conf.db); nothing can retarget the client afterwards.
+# Empty kwargs == datastore.Client(): no change for default deployments.
+__client_kwargs__: dict[str, str] = {}
+if conf.db.name:
+    __client_kwargs__["database"] = conf.db.name
+if conf.db.namespace:
+    __client_kwargs__["namespace"] = conf.db.namespace
+
+__client__ = datastore.Client(**__client_kwargs__)
```
sveneberth, line 28: Dunder names are reserved for the Python interpreter and standard library (https://docs.python.org/3/reference/lexical_analysis.html#reserved-classes-of-identifiers). We shouldn't use them. It's bad enough that `__client__` is defined that way. We shouldn't introduce any more names like that.

Furthermore, there is no need to create an additional dict if the default parameter is None anyway.

```suggestion
__client__ = datastore.Client(database=conf.db.name, namespace=conf.db.namespace)
```

src/viur/core/db/types.py
```diff
@@ -63,15 +63,33 @@ def __init__(self, *path_args, project: str | None = None, **kwargs):
...
 
+    def to_legacy_urlsafe(self, location_prefix=None):
+        # Upstream to_legacy_urlsafe() rejects keys carrying a database, but
+        # str(key)/session paths hit it constantly. Drop the db for encoding —
+        # unambiguous to restore since the process talks to a single database.
+        saved, self._database = self._database, None
+        try:
+            return super().to_legacy_urlsafe(location_prefix=location_prefix)
+        finally:
+            self._database = saved
```
sveneberth, line 91: This flip mutates shared instance state and is not thread-safe! If the same `Key` object is encoded from two threads concurrently, the second thread captures the transient `None` as its saved value and restores it in finally — leaving `_database=None` on the instance permanently. Data-integrity loss, not just a transient misread.

You have to create a copy of self, and modify only this instance to pass it to super.

src/viur/core/__init__.py
```diff
@@ -245,6 +245,16 @@ def __build_app(modules: ModuleType | object, renderers: ModuleType | object, de
     conf.main_app = index
 
 
+def _datastore_banner_lines() -> list[str]:
```
sveneberth, line 248: Why must this be a method, if it's called exactly one time?

tests/test_db.py
```diff
@@ -79,3 +82,105 @@ def test_query_order_plain_tuple_compat(self) -> None:
...
+        from viur.core import db
+        from viur.core.db import transport
+        with mock.patch.object(
+            transport, "__client__", self._fake_client(database="viur-tests"),
+        ):
+            key = db.Key("viur", "foo")
+            # Without the override both calls would raise ValueError.
+            self.assertIsInstance(key.to_legacy_urlsafe(), bytes)
+            self.assertIsInstance(str(key), str)
+
```
sveneberth, line 141: Please test also the other direction: `from_legacy_urlsafe`
```suggestion

    def test_from_legacy_urlsafe_tolerates_named_database(self) -> None:
        from viur.core import db
        from viur.core.db import transport
        with mock.patch.object(
            transport, "__client__", self._fake_client(database="viur-tests"),
        ):
            key = db.Key("viur", "foo")
            key_loaded = db.Key.from_legacy_urlsafe(key.to_legacy_urlsafe())
            self.assertEqual(key, key_loaded)
            self.assertEqual(key.database, key_loaded.database)
            self.assertEqual(key.namespace, key_loaded.namespace)
            self.assertEqual(key.project, key_loaded.project)
            self.assertEqual(key.id_or_name, key_loaded.id_or_name)

```

sveneberth (changes requested):
The current implementation is not thread-safe and should therefore under no circumstances be used as is.

However, if my comments are implemented, this will be a complete alternative to my PR #1698

sveneberth:
As decided in the ViUR meeting, I'm supposed to implement my own suggestions here.
However, since @akelch apparently didn't enable the _"Allow edits and access to secrets by maintainers"_ option for this PR, I'm unable to do so within this PR/branch.
That's why I created https://github.com/viur-framework/viur-core/pull/1743.

phorward:
Closed because of replacement #1743

PR #1719, opened 2026-07-01
fix: Normalize parsed `db.Key` to current `project_id`

phorward:
This caused errors "ValueError: Keys do not match project" when passing the key of a different project to any Method annotated with `db.Key`-type.

### DIFF
--- a/src/viur/core/db/utils.py
+++ b/src/viur/core/db/utils.py
@@ -54,7 +54,7 @@ def fix_unindexable_properties(entry: Entity, *, keep_exclusions: bool = True) -
 
 def normalize_key(key: t.Union[None, Key, str]) -> t.Union[None, Key]:
     """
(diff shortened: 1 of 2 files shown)


sveneberth (dismissed):
Hmm. I think it's right that keys aren't simply normalized and that the project is ignored. That's just not the correct identifier; it belongs to a different project. I don't see a use case for this in standard situations; rather, I see validation as a safety mechanism to catch programming errors (like an incorrect host or something like that).

phorward:
> rather, I see validation as a safety mechanism to catch programming errors (like an incorrect host or something like that).

In this cases, an 500 internal server error occurred. This should not be valid. Please provide a better solution.

phorward:
viur-meeting: Accept project_ids from `valid_application_ids`. Raise 406 otherwise.

ArneGudermann (approved):

PR #1720, opened 2026-07-07
fix(request): allow same-site requests in FetchMetaDataValidator

sveneberth:
The validator only accepted `same-site` requests on the dev server, so in production a legitimate same-site navigation was rejected with `403 Request rejected due to fetch metadata`. This happens e.g. when a request is redirected within the same registrable domain (non-www to www) or upgraded from http to https, both of which downgrade `Sec-Fetch-Site` to `same-site`.

Treat `same-site` like `same-origin` and `none`, as recommended by the reference resource isolation policy on https://web.dev/fetch-metadata/.

### DIFF
--- a/src/viur/core/request.py
+++ b/src/viur/core/request.py
@@ -74,15 +74,16 @@ class FetchMetaDataValidator(RequestValidator):
         headers = request.request.headers
 
         match headers.get("sec-fetch-site"):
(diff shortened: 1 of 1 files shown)


phorward (approved):
I'm ok with this. What does @ArneGudermann think?

ArneGudermann (approved):

PR #1722, opened 2026-07-08
fix(bones): reject invalid domain labels in `EmailBone`

sveneberth:
`encodings.idna.ToASCII` does not reject domain labels with a leading or trailing hyphen (e.g. `-online`), so addresses like `foo@-online.de` passed validation. Such addresses are later rejected by mail providers, which can make the email send-task retry indefinitely.

Validate each domain label explicitly: every label must be non-empty, must not start or end with a hyphen, and must contain only alphanumerics and hyphens after IDNA encoding; the TLD must not be purely numeric. Add tests covering valid and invalid addresses.

### DIFF
--- a/src/viur/core/bones/email.py
+++ b/src/viur/core/bones/email.py
@@ -1,7 +1,12 @@
+import re
 import string
 from encodings import idna
 from viur.core.bones.string import StringBone
 from viur.core import i18n
+
+_DNS_LABEL_RE = re.compile(r"(?!-)[a-z0-9-]{1,63}(?<!-)", re.IGNORECASE)
+"""A single DNS label per RFC 1035: 1-63 alphanumerics or hyphens, no leading or trailing hyphen."""
+
 
 class EmailBone(StringBone):
     """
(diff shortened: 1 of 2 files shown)


ArneGudermann (dismissed):

phorward (changes requested):
Is there no standard regular expression for checking this?

sveneberth:
> Is there no standard regular expression for checking this?

@copilot What do you think?

phorward:
> @copilot What do you think?

Did that work?

sveneberth:
> Is there no standard regular expression for checking this?

There's no single standard regex that covers this reliably. The RFC 5322 address grammar is notoriously unwieldy as a regex, and — more importantly — it doesn't help with the actual problem here: the domain may be a Unicode/IDNA name (e.g. `münchen.de`), which has to be IDNA-encoded *before* its label structure can be checked. A regex on the raw input would validate the wrong form (`-münchen` vs. `xn--…`).

Where a regex *does* fit is the per-label check after IDNA encoding, so I've reworked it that way: each IDNA-encoded label is matched against a compiled RFC 1035 pattern (`(?!-)[a-z0-9-]{1,63}(?<!-)`). This replaces the manual hyphen/char loop, and as a bonus enforces the 63-char label length limit, which the previous version didn't. I also dropped the now-redundant `subDomain`/`tld` split so the domain is parsed in a single place.

phorward (approved):

PR #1724, opened 2026-07-17
feat(bones): strict mode rejects unknown bone attributes (typo guard)

skoegl:
## Motivation

A one-character typo — `readonly` instead of the real bone attribute `readOnly` — silently creates a dead instance attribute instead of setting the intended flag. In a recent security audit this exact mistake left a user-record lock a no-op, so any authenticated user could grant themselves admin. Python's `setattr` accepts any name, and `BaseBone.__setattr__` did too for post-construction assignments (constructor `**kwargs` already raise `TypeError`, but later attribute assignment did not).

## What

New config flag **`conf.bone_strict_mode`** (default **on**, opt-out via `conf.bone_strict_mode = False` or env `VIUR_CORE_BONE_STRICT_MODE=false`, mirroring `VIUR_CORE_CONFIG_STRICT_MODE`).

Once a bone is bound to its Skeleton (`__set_name__`), assigning an **unknown** attribute name raises `AttributeError` instead of silently creating it:

```python
skel.myBone.readonly = True   # AttributeError: ... Did you mean: 'readOnly'?
skel.myBone.readOnly = True   # fine (existing attribute)
```

## How (minimal — extends existing hooks)

- No new state: a bone is "construction complete" once `__set_name__` has bound it to its Skeleton, i.e. `self.skel_cls is not None`. That survives `copy.deepcopy`, so cloned skeleton-instance bones are guarded too.
- Guard at the top of the existing `BaseBone.__setattr__`: reject when bound, flag on, name not `_`-prefixed, and `not hasattr(self, key)`. Existing attributes (`readOnly`, `required`, `visible`, `descr`, subclass attrs like `min`/`max`) and `_`-internals stay writable — no hardcoded allow-list.
- The `AttributeError` is raised with `name=` and `obj=`, so Python's traceback machinery appends its own suggestion, e.g. `Did you mean: 'readOnly'?` — no custom distance function needed.
- Assignments inside `__init__` are not affected, the bone is not bound yet at that point. The former dead `self.readonly = True` in `DateBone` is therefore not something this guard would have caught; that line has since been removed on `develop` (#1772), and this PR does not touch it.

## Compatibility

- **viur-core:** full test suite (672) passes with the flag default-on; no in-tree code assigns unknown attributes to a bone after construction.
- **Downstream (breaking, opt-out):** projects that assign *ad-hoc* attributes to bones after construction will now get `AttributeError`. That is the point (it catches the bug class), but it can be disabled per project via `conf.bone_strict_mode = False` / `VIUR_CORE_BONE_STRICT_MODE=false`.

## Tests

`tests/bones/test_base_bone.py::TestBoneStrictMode`: unknown attr raises under strict, including the `Did you mean: 'readOnly'?` hint in the rendered traceback; known attr allowed; disabled → legacy silent-set; constructor kwarg typo still raises `TypeError`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ This file documents any relevant changes done to ViUR-core since version 3.
 
 ## [3.9.0.dev6]
 
+- feat(bones): Add `conf.bone_strict_mode` — reject setting an unknown attribute on a bone after construction (typo guard for `readonly` vs `readOnly`); default on, opt-out via `VIUR_CORE_BONE_STRICT_MODE=false`
+- fix(DateBone): Correct `readonly` typo to `readOnly` so magic date bones become read-only as intended
 - chore: Deprecate endpoint `/vi/getStructure` (#1616)
 - doc: Docstrings for compute-related enum variants
 - doc: Fix seo_language_map in docstring
--- a/src/viur/core/bones/base.py
+++ b/src/viur/core/bones/base.py
@@ -467,6 +467,9 @@ class BaseBone(object):
     def __set_name__(self, owner: "Skeleton", name: str) -> None:
         self.skel_cls = owner
         self.name = name
+        # Construction is complete once the bone is bound to its Skeleton: from here on,
+        # bone_strict_mode rejects assignment of unknown attribute names (see __setattr__).
+        self._bone_sealed = True
 
     def setSystemInitialized(self) -> None:
         """
@@ -562,6 +565,13 @@ class BaseBone(object):
         :raises AttributeError: If a protected attribute is attempted to be modified after its initial
             assignment.
         """
+        if (getattr(self, "_bone_sealed", False) and conf.bone_strict_mode
+                and not key.startswith("_") and not hasattr(self, key)):
+            raise AttributeError(
+                f"{type(self).__name__} has no attribute {key!r} -- set after construction with "
+                f"bone_strict_mode enabled (typo? e.g. 'readonly' instead of 'readOnly'). "
+                f"Disable via conf.bone_strict_mode = False."
+            )
         if not self.isClonedInstance and getSystemInitialized() and key != "isClonedInstance" and not key.startswith(
                 "_"):
             raise AttributeError("You cannot modify this Skeleton. Grab a copy using .clone() first")
--- a/src/viur/core/bones/base.py
+++ b/src/viur/core/bones/base.py
@@ -567,10 +567,11 @@ class BaseBone(object):
         """
         if (getattr(self, "_bone_sealed", False) and conf.bone_strict_mode
                 and not key.startswith("_") and not hasattr(self, key)):
+            # name/obj let Python's traceback machinery append "Did you mean: 'readOnly'?"
             raise AttributeError(
                 f"{type(self).__name__} has no attribute {key!r} -- set after construction with "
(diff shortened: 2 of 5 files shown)


src/viur/core/bones/base.py
```diff
@@ -562,6 +565,13 @@ def __setattr__(self, key, value):
         :raises AttributeError: If a protected attribute is attempted to be modified after its initial
             assignment.
         """
+        if (getattr(self, "_bone_sealed", False) and conf.bone_strict_mode
+                and not key.startswith("_") and not hasattr(self, key)):
+            raise AttributeError(
```
sveneberth, line 570: It is possible to get the user-friendly hint "Did you mean" like on `NameError`?

```py
>>> python -c "lst()"
  File "<string>", line 1, in <module>
    lst()
    ^^^
NameError: name 'lst' is not defined. Did you mean: 'list'?
```

skoegl: Yes -- Python does this itself since 3.10: an `AttributeError` raised with `name=` and `obj=` gets the "Did you mean: 'readOnly'?" suffix appended by the traceback machinery (default excepthook, `traceback.print_exception`, `logging.exception`). No custom distance function needed.

Added in 9ef0938, incl. a test asserting the hint via `traceback.format_exception_only`. Note: the hint lives in the rendered traceback, not in `str(e)`.

sveneberth: nice, good to know.

sveneberth (dismissed):

ArneGudermann (dismissed):

CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ This file documents any relevant changes done to ViUR-core since version 3.
 - feat: Re-Add trace-queries logging (#1642)
 - feat: Refactor and improve `getUniquePropertyIndexValues` (#1681)
 - feat: Replace client-side `MultiQuery` with native Datastore operators (#1700)
+- feat(bones): Add `conf.bone_strict_mode` — reject setting an unknown attribute on a bone after construction (typo guard for `readonly` vs `readOnly`); default on, opt-out via `VIUR_CORE_BONE_STRICT_MODE=false`
```
sveneberth:
```suggestion
```
That's not part of a PR, it's part of the release and done by the maintainer. 

Please familiarize yourself with https://github.com/viur-framework/viur-core/blob/main/CONTRIBUTING.md#releasing

sveneberth (changes requested):

CHANGELOG.md
skoegl: Done, the CHANGELOG entry is gone (cb3719192), same for #1725.

For the record: this PR was opened on July 17 and reviewed on July 22, the CHANGELOG line has been in the diff since day one. It then survived a re-review and two approvals on September 9 and 10, with no mention of it. If the entry is a blocker, a heads-up in one of those four passes would have saved both of us a round trip.

I'm happy to follow the releasing process as documented. It would help contributors if CONTRIBUTING.md said so explicitly in the PR section too, "don't touch CHANGELOG.md, the maintainer generates it at release time". The releasing section is written for people with release permissions, which is why it doesn't get read by the rest of us.

ArneGudermann (approved):

sveneberth (approved):

PR #1725, opened 2026-07-17
feat(file): accept retired hmac-keys for zero-downtime key rotation

skoegl:
## Motivation

`conf.file_hmac_key` signs every file download URL (and the `/file/serve` auth
and blobKey signatures — all of which verify through `File.hmac_verify`). There
is only **one** key, and `hmac_verify` recomputes the signature with exactly
that key. Rotating it — e.g. after a suspected leak — therefore invalidates
**every** signature still in the wild at once: download URLs until they expire
(`expires=0` never expires), plus long-lived serve signatures. In practice the
key is effectively unrotatable without breaking live links.

## What

New config **`conf.file_hmac_key_fallbacks: list[str]`** (default `[]`): retired
keys that `hmac_verify` still **accepts** but `hmac_sign` never uses. Enables
zero-downtime rotation:

1. Set the new key as `file_hmac_key`, move the old one into `file_hmac_key_fallbacks`.
2. New URLs are signed with the new key; URLs signed with the old key still verify.
3. Once every link signed with the old key has expired, drop it from the fallbacks.

Analogous to Django's `SECRET_KEY_FALLBACKS`.

## How (minimal — one choke point)

All signed artifacts verify through `File.hmac_verify`, so only signing and
verification change:

- `hmac_sign(data, key=None)` gets an optional `key`. With the default `None` it
  signs with the active `conf.file_hmac_key`, as before. Keys given as `str` are
  coerced to bytes like the active key. Signing and verification share this
  single method, so they **cannot diverge** on the algorithm (`sha3_384`).
- `hmac_verify` tries the active key plus each fallback by calling `hmac_sign`
  with that `key`, compares each result with `hmac.compare_digest`
  (constant-time) and returns True on the first match.
- Both methods have docstrings explaining the rotation scheme.

No startup/auto-generation change; fallbacks are operator-set during rotation.

## Compatibility

Fully backward compatible: with the default `[]`, `hmac_verify` behaves exactly
as before (single active key). The new `key` parameter of `hmac_sign` is
optional, existing callers are unaffected. No downstream change required.

## Tests

`tests/modules/test_file.py::TestFileHmacRotation`: old signature still verifies
after rotation with fallback; rejected without fallback; str-configured fallback
accepted; garbage signature rejected; explicit `key` yields the same digest as
the default. Full suite (664 tests) passes on 3.12, 3.13 and 3.14.


src/viur/core/modules/file.py
```diff
@@ -572,16 +572,29 @@ def is_valid_filename(cls, filename: str) -> bool:
...
 
     @staticmethod
-    def hmac_sign(data: t.Any) -> str:
+    def _hmac_digest(key: bytes | str, data: bytes) -> str:
+        if not isinstance(key, bytes):
+            key = key.encode("UTF-8")
+        return hmac.new(key, msg=data, digestmod=hashlib.sha3_384).hexdigest()
+
+    @classmethod
+    def hmac_sign(cls, data: t.Any) -> str:
```
sveneberth, line 581: Can you add some docstrings here?

It took me a moment to realize that one method accepts any keys and the other always uses `conf.file_hmac_key`.

But actually, couldn't we just combine them into one method where `key=None` (default) uses `conf.file_hmac_key`?!


--- 

Besides this the PR and feature LGTM.

sveneberth (changes requested):

src/viur/core/modules/file.py
skoegl: @sveneberth Thx for the good comments. Please reconsider a re-review.

src/viur/core/modules/file.py
```diff
@@ -575,16 +575,46 @@ def is_valid_filename(cls, filename: str) -> bool:
...
+        Signs `data` with an HMAC (SHA3-384) and returns the hex digest.
+
+        :param data: Payload to sign. Anything that is not `bytes` is converted via `str()` and UTF-8 encoded.
+        :param key: Key to sign with. Defaults to the active :attr:`conf.file_hmac_key`; `str` keys are UTF-8 encoded.
+            Only :meth:`hmac_verify` passes another key, to check signatures made with a retired key from
+            :attr:`conf.file_hmac_key_fallbacks`. New signatures are always created with the active key.
+        """
+        if key is None:
+            key = conf.file_hmac_key
+            assert key is not None, "No hmac-key set!"
```
sveneberth, line 589: nit: `assert`s are note the best solution for validations, since they can be disabled with `-O`.

sveneberth (approved):

PR #1727, opened 2026-07-21
feat: `edit`-function in List/Tree/Singleton using `Skeleton.patch()`

phorward:
This runs the edit-action of List/Tree/Singleton-prototypes in a transaction.
Previously, it could be possible that data would become lost due race conditions.

### DIFF
--- a/src/viur/core/bones/base.py
+++ b/src/viur/core/bones/base.py
@@ -11,6 +11,7 @@ import hashlib
 import inspect
 import logging
 import typing as t
+import warnings
 from collections.abc import Iterable
 from dataclasses import dataclass, field
 from datetime import timedelta
@@ -73,11 +74,15 @@ class ReadFromClientErrorSeverity(Enum):
 
 
 @dataclass
-class ReadFromClientError:
+class ReadFromClientError(ValueError):
     """
     The ReadFromClientError class represents an error that occurs while reading data from the client.
     This class is used to store information about the error, including its severity, an error message,
     the field path where the error occurred, and a list of invalidated fields.
+
+    It is also an Exception, so a single error can be raised directly. Use :meth:`raise_from` to raise
+    one or several errors at once (e.g. the accumulated skel.errors of a Skeleton) - if there are
+    several, the raised ReadFromClientError carries them in its `errors` attribute.
     """
     severity: ReadFromClientErrorSeverity
     """A ReadFromClientErrorSeverity enumeration value representing the severity of the error."""
@@ -87,6 +92,8 @@ class ReadFromClientError:
     """A list of strings representing the path to the field where the error occurred."""
     invalidatedFields: list[str] = None
     """A list of strings containing the names of invalidated fields, if any."""
+    errors: tuple["ReadFromClientError", ...] = None
+    """If set, this instance represents several errors at once; see :meth:`raise_from`."""
 
     def __post_init__(self):
         if not self.errorMessage:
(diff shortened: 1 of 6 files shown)


sveneberth (commented):
What's the advantage here, apart from a different syntax?

phorward:
> What's the advantage here, apart from a different syntax?

It's not different syntax, it turns the entire edit-action into a transaction-safe process.
This isn't relevant to add or delete.

I've updated the description of this PR as follows:

> This runs the edit-action of List/Tree/Singleton-prototypes in a transaction.
> Previously, it could be possible that data would become lost due race conditions.

src/viur/core/bones/base.py
```diff
@@ -73,11 +74,15 @@ class ReadFromClientErrorSeverity(Enum):
 
 
 @dataclass
-class ReadFromClientError:
+class ReadFromClientError(ValueError):
```
sveneberth, line 77: This is an incomplete subclass of `ValueError` or `Exception`. The `args` arguments are missing. 
Also, I'm not sure — or haven't been able to find out — how clean the MRO resolution of the dataclass is when it's not based on a dataclass, and whether, for example, `super` calls are handled correctly on all bases.

---

But in general, I don't understand the restructuring here. Why not keep containers (`ReadFromClientException`) and data objects (`ReadFromClientError`) separate? 
It’s also now possible to nest any number of errors within each other. How is recursion supposed to be handled here? With two types, that wasn’t possible...
To me, this whole thing seems like a solution to a problem that never existed— at least, I don’t see a suitable description or explanation for it in the PR.

src/viur/core/bones/base.py
```diff
@@ -87,6 +92,8 @@ class ReadFromClientError:
     """A list of strings representing the path to the field where the error occurred."""
     invalidatedFields: list[str] = None
     """A list of strings containing the names of invalidated fields, if any."""
+    errors: tuple["ReadFromClientError", ...] = None
```
sveneberth, line 95: That's exactly what `ReadFromClientException` has already been doing, holding instances of `ReadFromClientError`.

src/viur/core/skeleton/skeleton.py
```diff
@@ -304,7 +310,7 @@ def read(
             return None
         elif isinstance(create, dict):
             if create and not skel.fromClient(create, amend=True, ignore=()):
-                raise ReadFromClientException(skel.errors)
+                ReadFromClientError.raise_from(skel.errors)
```
sveneberth, line 313: What kind of unpythonic syntax is that?

Why hide the `raise` keyword instead of using the constructor like with every other exception? Does ViUR always have to be different?

Why not just do it this way?
```suggestion
                raise ReadFromClientError(skel.errors)
```

src/viur/core/bones/base.py
```diff
@@ -102,12 +109,45 @@ def __post_init__(self):
...
+
+        Raises the sole error directly if only one is given; otherwise raises a single
+        ReadFromClientError representing all of them (see its `errors` attribute).
+
+        :param errors: Either one or an iterable of errors.
+        """
+        if isinstance(errors, ReadFromClientError):
+            raise errors
+
+        error_tuple = tuple(error for error in errors if isinstance(error, ReadFromClientError))
```
sveneberth, line 130: Are objects of a different type simply silently discarded? If a type check is already taking place, then a `TypeError` should be thrown here.

sveneberth (changes requested):

PR #1734, opened 2026-07-28
fix: i18n `_last_reload`-debounce

KadirBalku:
```python
self._last_reload - utils.utcNow() < datetime.timedelta(minutes=10)
```
_last_reload is only ever set to utils.utcNow(), so it always holds a past
point in time. past - now is a negative timedelta, which is always less
than 10 minutes — the debounce condition is permanently true.

Impact

_last_reload starts as None, so the first call still reloads. Every call
after that returns early. systemTranslations is refreshed exactly once per
module instance, meaning translations edited in the admin are not picked up
until the instance is recycled or the app is redeployed.

### DIFF
--- a/src/viur/core/modules/translation.py
+++ b/src/viur/core/modules/translation.py
@@ -238,7 +238,7 @@ class Translation(List):
     def _reload_translations(self):
         if (
             self._last_reload is not None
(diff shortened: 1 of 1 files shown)


phorward (approved):
Looks absolutely correct.

PR #1735, opened 2026-07-28
feat: add i18n fallback languages

KadirBalku:
## Problem

There is no fallback across languages. If a translation has no value for the
requested language, `translate` goes straight to `_default_text_`, then the
caller's `defaultText`, then the key itself - it never tries another language.
`merge_alias` only copies a main language into its dialects (`at` -> `de`), it
does not chain between main languages.

Projects work around this with their own wrapper functions, which is framework
work leaking into project code.

## Change

New config option, empty by default so existing behaviour is unchanged:

```python
conf.i18n.fallback_languages = ["de", "en"]
```
Resolution order becomes: requested language (incl. alias) -> each fallback
language in order -> default text.

Both resolution paths now share one helper, translate.resolve_language(), used
by translate.__str__ and TranslationExtension._translate.

Note

This also aligns the two paths. _translate previously used
translations.get(lang, default_text), so a translation present but empty was
returned verbatim, while __str__ already treated empty as missing. Empty and
whitespace-only values now count as missing in both, consistent with
merge_alias.


src/viur/core/i18n.py
sveneberth, line 74: Please explain in this docstring the behavior of language resolution.

E.g. as in your description
> Resolution order becomes: requested language (incl. alias) -> each fallback
language in order -> default text.

sveneberth (changes requested):
LGTM, but please explain this behavior not only in the PR description, but also in the code and docstrings.

phorward (dismissed):

ArneGudermann (dismissed):

sveneberth (dismissed):

PR #1737, opened 2026-07-28
fix: Make user `Status` an `IntEnum` for cross-project comparability

### LINKED ISSUE #1671: fix: User.addSkel() hardcodes Status.UNSET, breaking projects that override the status bone
## Problem

  `User.addSkel()` in `viur/core/modules/user.py` assigns the core `Status.UNSET` enum instance directly to `skel["status"]` for non-admin users. This hardcoded assignment breaks projects that
  override the `status` bone with project-specific values (e.g. additional states, integer-keyed `values`, or a different enum).

  ## Affected code

  viur-core 3.8.22 — `viur/core/modules/user.py`, in `User.addSkel()`:

  ```python
  def addSkel(self):
      skel = super().addSkel().clone()

      if self.is_admin(current.user.get()):
          ...
      else:
          skel.status.readOnly = True
          skel["status"] = Status.UNSET          # <-- hardcoded enum
          skel.status.visible = False
          ...
  ```

  The default core bone uses `values=Status` (the enum), so `Status.UNSET` works there. Projects that override `UserSkel.status` with a different `SelectBone` (different `values` dict / different
  enum) fail validation.

  ## Reproduction

  1. Override `UserSkel.status` with a custom `SelectBone`, e.g. integer-keyed or a project enum with additional states:

     ```python
     class MyStatus(enum.Enum):
         UNSET = 0
         WAITING_FOR_EMAIL_VERIFICATION = 1
         WAITING_FOR_ADMIN_VERIFICATION = 2
         DISABLED = 5
         LOCKED_BY_EXTERNAL_SYSTEM = 6   # extra state not in core
         ACTIVE = 10

     status = SelectBone(
         values={s.value: s.name for s in MyStatus},
         defaultValue=MyStatus.ACTIVE.value,
         required=True,
     )
     ```

  2. Log in as a user with a non-admin management role (e.g. `user-management`).
  3. Try to create a new user via `/vi/user/add`.

  ## Expected

  User gets created; the overridden `status` bone's `defaultValue` (or an explicit, bone-compatible value) is used.

  ## Actual

  `fromClient` / validation fails because `Status.UNSET` (the core enum instance) is not a member of the overridden bone's `values` dict.

  ## Suggested fix

  Don't hardcode the core enum in `addSkel()`. Possible options:

  - Rely on the bone's own `defaultValue` — don't touch `skel["status"]` at all, only adjust `readOnly`/`visible`.
  - Assign `skel.status.defaultValue` instead of `Status.UNSET`.
  - At minimum, assign `Status.UNSET.value` (integer `0`) so projects using integer-keyed `SelectBone` overrides keep working.

  ## Current workaround

  ```python
  def addSkel(self):
      skel = super().addSkel()
      ...
      if current.user.get() and not self.is_admin(current.user.get()):
          skel["status"] = skel["status"].value  # coerce core enum -> int
      return skel
  ```

  ## Environment

  - viur-core 3.8.22
  - Python 3.13

phorward:
Downstream projects that redefine `Status` to add custom account states end up with two `Enum` classes that don't compare equal, even for the same value. `IntEnum` fixes this, since members compare via plain int equality across independently defined enums.

Also drops the now-obsolete `__eq__`/`__lt__` backward-compat shim and the `.value` workarounds this required.

Closes #1671

### DIFF
--- a/src/viur/core/modules/user.py
+++ b/src/viur/core/modules/user.py
@@ -2,7 +2,6 @@ import abc
 import datetime
 import enum
 import fnmatch
-import functools
 import hashlib
 import hmac
 import json
@@ -35,12 +34,27 @@ from viur.core.securityheaders import extendCsp
 from viur.core.session import Session
 
 
-@functools.total_ordering
-class Status(enum.Enum):
+class Status(enum.IntEnum):
     """Status enum for a user
 
-    Has backwards compatibility to be comparable with non-enum values.
-    Will be removed with viur-core 4.0.0
+    This is an IntEnum, so it is comparable with plain ints as well as with
+    other IntEnum classes of the same values, e.g. when a project defines its
+    own Status enum to add custom status values to a subclassed UserSkel:
+
+    class Status(enum.IntEnum):
+        UNSET = 0
+        WAITING_FOR_EMAIL_VERIFICATION = 1
+        WAITING_FOR_ADMIN_VERIFICATION = 2
+        DISABLED = 5
+        ACTIVE = 10
+        PENDING_REVIEW = 15  # custom, project-specific status
+
+    class UserSkel(user.UserSkel):
+        status = SelectBone(
+            ...,
+            values=Status,
+            defaultValue=Status.ACTIVE,
+        )
     """
 
     UNSET = 0  # Status is unset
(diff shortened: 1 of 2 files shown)


sveneberth (approved):
That's a nice and easy solution

PR #1747, opened 2026-08-06
feat: add Query.iter_skel() to iterate over SkeletonInstances

ArneGudermann:
`iter_skel()` is to `iter()` what` fetch()` is to `run()`: it walks a large result set via the query cursor 
without pulling it in advance, but yields SkeletonInstances instead of Entities. Every result is a  
 separate instance sharing the source skeleton's bone-map, so collecting or writing the results      
 inside the loop behaves as expected.                                                                
                                                                                                     
The query must have been created by skel.all(); queries without a srcSkel and multi-queries raise   
immediately on call instead of on the first next(), which is why the iteration itself lives in an   
inner generator.

Also fix Query.iter() raising StopIteration inside a generator: PEP 479 turns that into a           
RuntimeError, so an unsatisfiable query now ends the iteration cleanly.

### DIFF
--- a/src/viur/core/db/query.py
+++ b/src/viur/core/db/query.py
@@ -787,7 +787,7 @@ class Query(object):
         for a query for more than ~30 seconds.
         """
         if self.queries is None:  # Noting to pull here
-            raise StopIteration()
+            return
         elif isinstance(self.queries, list):
             raise ValueError("No iter on Multiqueries")
         while True:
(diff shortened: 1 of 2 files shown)


src/viur/core/db/query.py
```diff
@@ -796,6 +796,46 @@ def iter(self, keys_only=False) -> t.Iterator[Entity]:
...
+        :raises ValueError: If this is a multi-query, which cannot be iterated.
+        """
+        if self.srcSkel is None:
+            raise NotImplementedError("This query has not been created using skel.all()")
+        elif isinstance(self.queries, list):
+            raise ValueError("No iter_skel on Multiqueries")
+
+        from viur.core.skeleton import SkeletonInstance
+
+        # Wrapped in an inner generator, so the checks above are raised on call and not on first next()
```
sveneberth, line 830: Wow, that's crazy — I didn't know that either. Thanks for the explanation.

sveneberth (approved):

phorward (approved):

Issue #1539, opened 2025-08-05
Merge `StringBone(escape_html=True)` and `TextBone` into `HtmlBone`


phorward:
- Make `StringBone` a real `StringBone`, with optional `multiline=True`
- Make `TextBone` a `HtmlBone` with HTML-escaping for website frontend content

PR #1751, opened 2026-08-12
feat: add `escape_html`-parameter to `TextBone`

ArneGudermann:
Adds `escape_html` to `TextBone`, mirroring the `StringBone` option. With `escape_html=False` the value is stored as-is — no tags filtered, no characters escaped, line breaks preserved. Previously every value 
  went through `HtmlSerializer`, so a plain multi-line text bone was impossible even with `validHtml=None`.                                                                                                        
                                                                                                                                                                                                                   
  - Only allowed with `validHtml=None`, otherwise `ValueError` (default included, it falls back to `conf.bone_html_default_allow`)                                                                                 
  - Validation (`max_length`, `required`, type) unchanged — only sanitizing is skipped                                                                                                                             
  - `getReferencedBlobs()` returns an empty set; content is no longer parsed as HTML                                                                                                                               
  - `refresh()` unescapes stored values to migrate entries written under the old behavior (`multiple` and `languages` included)                                                                                    
  - `structure()` exposes the flag for the admin UI                                                                                                                                                                
  - Default `True` → no behavior change                                                                                                                                                                            
                                                                                                                                                                                                                   
  **Security:** disabling escaping removes all server-side sanitizing (XSS risk) and blob tracking, so linked files are no longer protected from deletion. Documented as a warning in the docstring.               
                                                                                                                                                                                                                   
  **Tests:** invalid parameter combinations, raw pass-through, validation, empty blob set, `refresh()` unescaping in all variants, `structure()`.

### DIFF
--- a/src/viur/core/bones/text.py
+++ b/src/viur/core/bones/text.py
@@ -7,7 +7,7 @@ import string
 import typing as t
 import warnings
 from html.parser import HTMLParser
-from viur.core import db, conf, i18n
+from viur.core import db, conf, i18n, utils
 from .base import ReadFromClientError, ReadFromClientErrorSeverity
 from .raw import RawBone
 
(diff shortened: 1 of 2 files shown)


sveneberth (commented):
Hmm. Without a use case, it's really hard to see what the benefit of this is right now. I'd say that if you really don't care about anything, you should use a `RawBone` or `CodeBone`. Or a `StringBone` with `escape_html=False`.

I’d call that parameter `sanitize_html` in this bone, since there’s a bit more to it than just escaping.
But couldn’t we handle this with a few tweaks in the case where `validHtml=None`?

phorward:
Good morning fellows!

> Hmm. Without a use case, it's really hard to see what the benefit of this is right now. I'd say that if you really don't care about anything, you should use a `RawBone` or `CodeBone`. Or a `StringBone` with `escape_html=False`.

I'm not sure either about this feature. IMHO TextBone should be called HtmlBone, and a TextBone should be what a StringBone is, but with line-breaks (and the HTML escaping option). But this isn't a change we can do in 3.9.

See https://github.com/viur-framework/viur-core/issues/1539 on this, too.

phorward (approved):

PR #1752, opened 2026-08-12
fix: `getDefaultValue` shares one list across all languages of a multiple bone

ClaudiaDietrichLev:
For a bone with `languages` and `multiple`, the constructor builds the default value as `{lang: default for lang in self.languages}` - with a single empty list instance as `default`. `getDefaultValue()` then only returns `self.defaultValue.copy()`, a shallow copy: the dict is new, but all languages still point to that one list, which also remains the instance stored on the bone itself.

Any code that fills the freshly initialized value in place - e.g. `skel[name][lang].append(...)`, which is how `setBoneValue(..., append=True, language=...)` works - therefore writes into all languages at once. And since the mutated list is the bone's own `defaultValue` (bones live on the skeleton class, i.e. process level), the mutation survives the request: every skeleton created afterwards starts with the accumulated entries of previous ones. Introduced with #1282, which turned the per-language `None` default into a single shared `[]`.

Fixed by copying the inner lists as well when `getDefaultValue` hands out a dict default, analogous to the existing `self.defaultValue[:]` for plain list defaults. `fromClient` and `unserialize` are not affected - they already build fresh lists per language.

### DIFF
--- a/src/viur/core/bones/base.py
+++ b/src/viur/core/bones/base.py
@@ -537,7 +537,13 @@ class BaseBone(object):
         elif isinstance(self.defaultValue, list):
             return self.defaultValue[:]
         elif isinstance(self.defaultValue, dict):
-            return self.defaultValue.copy()
+            # A shallow dict copy is not enough: the inner lists must be copied as well,
+            # otherwise all languages share one list instance living on the bone, and any
+            # mutation bleeds into the other languages and into every other skeleton.
+            return {
+                lang: value[:] if isinstance(value, list) else value
+                for lang, value in self.defaultValue.items()
+            }
         else:
             return self.defaultValue
 
--- a/src/viur/core/bones/base.py
+++ b/src/viur/core/bones/base.py
@@ -535,12 +535,15 @@ class BaseBone(object):
             return res
 
         elif isinstance(self.defaultValue, list):
-            return copy.deepcopy(self.defaultValue)
+            return self.defaultValue[:]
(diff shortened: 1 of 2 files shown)


ArneGudermann (dismissed):

src/viur/core/bones/base.py
```diff
@@ -537,7 +537,13 @@ def getDefaultValue(self, skeletonInstance):
...
             return self.defaultValue[:]
         elif isinstance(self.defaultValue, dict):
-            return self.defaultValue.copy()
+            # A shallow dict copy is not enough: the inner lists must be copied as well,
+            # otherwise all languages share one list instance living on the bone, and any
+            # mutation bleeds into the other languages and into every other skeleton.
+            return {
+                lang: value[:] if isinstance(value, list) else value
+                for lang, value in self.defaultValue.items()
+            }
```
phorward: Can you please verify if this can be resolved with [`copy.deepcopy`](https://docs.python.org/3/library/copy.html)?

phorward (changes requested):

src/viur/core/bones/base.py
ClaudiaDietrichLev: Verified - deepcopy over the whole dict keeps the aliasing: the one list shared by all languages comes back as one new list, again shared (the regression test caught it). Deep-copying each language value separately works, since every call uses its own memo. Pushed that variant, nested defaults are covered now.

phorward (approved):

PR #1753, opened 2026-08-13
fix: `ColorBone` rejects misplaced `#` and non-string values

ArneGudermann:
## What                                                                                                                                                                                                          
                                                                                                                                                                                                                   
  `ColorBone.singleValueFromClient` only counted the `#` characters and then                                                                                                                                       
  looked at lengths. The position of the `#` was never checked and the length                                                                                                                                      
  branches cascaded into each other (len 3 → prepend `#` → len 4 → expand), so                                                                                                                                     
  malformed input was silently accepted and stored in a broken form:                                                                                                                                               
                                                                                                                                                                                                                   
  | Input       | Before                       | After     |                                                                                                                                                       
  |-------------|------------------------------|-----------|                                                                                                                                                       
  | `"#ab"`     | stored as `"###aabb"`        | `Invalid` |                                                                                                                                                       
  | `"abcdefa"` | stored as-is, without a `#`  | `Invalid` |                                                                                                                                                       
  | `"ab#cde"`  | stored as `"#ab#cde"`        | `Invalid` |                                                                                                                                                       
  | `123`       | `AttributeError` → HTTP 500  | `Invalid` |                                                                                                                                                       
                                                                                                                                                                                                                   
  Validation now works the other way round: strip an optional leading `#` once,                                                                                                                                    
  check the remainder as pure hex digits, check its length against the ones valid                                                                                                                                  
  for the mode (3 or 6 for `rgb`, 8 for `rgba`), expand the 3-digit shorthand and                                                                                                                                  
  prepend exactly one `#`. Non-string values are rejected instead of crashing.                                                                                                                                     
                                                                                                                                                                                                                   
  Additionally, `ColorBone` had no `structure()` override, so `mode` never reached                                                                                                                                 
  the client — an rgb and an rgba bone looked identical. `structure()` now exposes                                                                                                                                 
  `mode`.                                                                                                                                                                                                          
                                                                                                                                                                                                                   
  Unchanged behaviour: `"aabbcc"`/`"#aabbcc"` → `"#aabbcc"`, `"abc"` → `"#aabbcc"`,                                                                                                                                
  `"AABBCC"` → `"#aabbcc"`, `"##aabbcc"` → `Invalid`, `mode="rgba"` accepts 8 hex                                                                                                                                  
  digits only.                                                                                                                                                                                                     
                                                                                                                                                                                                                   
  ## Breaking change                                                                                                                                                                                               
                                                                                                                                                                                                                   
  This is a behaviour change: input that used to be accepted (and mangled) is now                                                                                                                                  
  rejected with `ReadFromClientErrorSeverity.Invalid`. Values already stored in the                                                                                                                                
  datastore are untouched and are not re-validated on read, so existing malformed                                                                                                                                  
  values (e.g. `###aabb`) stay as they are.

### DIFF
--- a/src/viur/core/bones/color.py
+++ b/src/viur/core/bones/color.py
@@ -1,4 +1,5 @@
 import string
+import typing as t
 from viur.core import i18n
 from .base import BaseBone, ReadFromClientError, ReadFromClientErrorSeverity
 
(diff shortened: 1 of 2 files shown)


src/viur/core/bones/color.py
```diff
@@ -14,45 +15,55 @@ class ColorBone(BaseBone):
...
+    # Accepted number of hex digits (without the leading "#") per mode
+    VALID_LENGTHS: t.Final[dict[str, tuple[int, ...]]] = {
+        "rgb": (3, 6),
+        "rgba": (8,),
+    }
+
     def __init__(self, *, mode="rgb", **kwargs):  # mode rgb/rgba
         super().__init__(**kwargs)
-        assert mode in {"rgb", "rgba"}
+        assert mode in self.VALID_LENGTHS
```
phorward:
```suggestion
        if mode not in self.VALID_LENGTHS:
            raise ValueError(f"{mode=} is not in {self.VALID_LENGTHS!r}")
```

phorward (changes requested):

phorward (approved):

PR #1760, opened 2026-08-21
fix(db): repair the datastore cache read path

ArneGudermann:
The datastore cache added in #1514 has never served a single hit — two bugs, one                                                                                                                                 
  hiding the other:                                                                                                                                                                                                
                                                                                                                                                                                                                   
  1. `cache.get()` read `conf.db_memcache_client` instead of                                                                                                                                                       
     `conf.db.memcache_client`. `conf` is a `ConfigType`, so that raises                                                                                                                                           
     `AttributeError` in strict *and* non-strict mode, swallowed by the surrounding                                                                                                                                
     `try/except`. Net effect per `db.get()`: an error log line, a superfluous                                                                                                                                     
     memcache write, zero hits — worse enabled than disabled.                                                                                                                                                      
                                                                                                                                                                                                                   
  2. Fixing 1. crashes on every hit. `cache.get()` returned a bare `Entity` for a                                                                                                                                  
     single hit, and `transport.get()` normalized that with                                                                                                                                                        
     `utils.ensure_iterable()` — which passes an `Entity` through unchanged,                                                                                                                                       
     because an `Entity` is dict-like and therefore already `Iterable`:                                                                                                                                            
                                                                                                                                                                                                                   
     ```python                                                                                                                                                                                                     
     e = Entity(Key("auftrag", "A1")); e["name"] = "Test"                                                                                                                                                          
     {str(x.key): x for x in utils.ensure_iterable(e)}                                                                                                                                                             
     # AttributeError: 'str' object has no attribute 'key'                                                                                                                                                         
     ```                                                                                                                                                                                                           
                                                                                                                                                                                                                   
     Iterating an `Entity` yields property *names*. This is outside of any `try`,                                                                                                                                  
     and a single key is the common case (`Skeleton.read()`), so every cache hit                                                                                                                                   
     would have aborted the request.                                                                                                                                                                               
                                                                                                                                                                                                                   
  **Fix:** `cache.get()` always returns a list, so `transport.get()` needs no                                                                                                                                      
  normalization. `db.put()` now caches *after* the datastore accepted the write —                                                                                                                                  
  before, a failed write could leave a value that was never persisted in the cache                                                                                                                                 
  for up to `MEMCACHE_TIMEOUT` (1 day).                                                                                                                                                                            
                                                                                                                                                                                                                   
  **Tests:** `TestDbCache` adds the first 15 tests for the layer, using the                                                                                                                                        
  memcache stub of the testbed `ViURTestCase` already activates — no emulator                                                                                                                                      
  needed. 11 of them fail against `develop`.


tests/abstract.py
```diff
@@ -16,4 +16,14 @@ def setUp(self) -> None:
         google.auth.default = mock.Mock(return_value=(mock.Mock(), os.getenv("GOOGLE_CLOUD_PROJECT")))
 
     def tearDown(self) -> None:
+        # ContextVars outlive the test case that set them. A leaked current.request
```
sveneberth: I've created https://github.com/viur-framework/viur-core/pull/1759/changes to fix this. We should separate this.

phorward (approved):

PR #1762, opened 2026-08-22
fix: `read_all_modules` overwrites existing `ModuleConf` entries

sveneberth:
## The bug

`ModuleConf.read_all_modules` collects the module names that already have a `viur-module-conf` entry, then checks every routable module against them:

```python
db_module_names = (m["name"] for m in db.Query(MODULECONF_KINDNAME).run(999))
...
    if module_name not in db_module_names:
        skel = conf.main_app.vi._moduleconf.addSkel()
        skel["key"] = db.Key(MODULECONF_KINDNAME, module_name)
        skel["name"] = module_name
        skel.write()
```

`db_module_names` is a **generator expression**, and `in` consumes it.

As long as every lookup succeeds, the generator merely advances and the check appears to work. But the first module that has no entry yet drains the generator to exhaustion — and from then on it is empty. Every module checked afterwards is reported as missing and gets written with a blank `addSkel()`.

## Why this loses data

That write is not additive. `ModuleConfSkel` is explicitly meant to be subclassed by applications, and every bone an application adds to it is part of the skeleton — so it is persisted as its empty default. SEO identifiers, per-module configuration, editorial help texts: all silently replaced with `None` / `[]`.

The blast radius is *everything sorted after the first newly added module*. Since `dir()` yields names sorted and the datastore query returns them in key order, the two happen to line up, so the damage is a contiguous run from the new module to the end of the module list.

## Why it went unnoticed for so long

It needs a module to be **added**. On an unchanged deployment every lookup hits, the generator is never exhausted, and the task is correct. It only fires on the first startup after a release that introduces a module or sub-module — and then it leaves no trace: `read_all_modules` completes successfully, it just writes the wrong thing. No exception, no log line.

Worth knowing for anyone assessing exposure: a version that is **deployed but not yet serving traffic** is enough. The startup task runs on the first request that version receives and writes to the same datastore — opening a staged version's URL once can wipe the production entries.

## The fix

The lookup collection becomes a `set`, which is what a repeated membership test needs: it cannot be consumed, and each check is O(1) instead of O(n).

## Regression coverage

`tests/modules/test_moduleconf.py` runs `read_all_modules` against a faked datastore and module tree:

- `test_known_module_after_missing_one_is_left_alone` — the actual regression: a missing module must not cause the modules checked after it to be rewritten.
- `test_lookup_is_order_independent` — the lookup must not depend on the query returning names in the same order `dir()` yields them. Nothing guarantees that alignment.
- plus the two baseline cases (missing module is created, known modules are left alone).

Both regression tests fail on `main` and pass with this change; the full suite is green (79 tests).

## Origin

`read_all_modules` was rewritten into the recursive `collect_modules` form in #1073, which turned the former list into a generator expression. Releases before that are unaffected.

## Left untouched

`run(999)` still caps the lookup — an application with more than 999 modules would reach the same overwrite by a different route. Kept out of this PR to keep the fix minimal, but it is the same failure class and may deserve its own issue.

### DIFF
--- a/src/viur/core/modules/moduleconf.py
+++ b/src/viur/core/modules/moduleconf.py
@@ -151,7 +151,9 @@ class ModuleConf(List):
     @tasks.StartupTask
     @staticmethod
     def read_all_modules():
-        db_module_names = (m["name"] for m in db.Query(MODULECONF_KINDNAME).run(999))
+        # Must be a set, not a generator: the membership test below runs once per module
+        # and a generator would be consumed by it (see #1762).
+        db_module_names = {m["name"] for m in db.Query(MODULECONF_KINDNAME).run(999)}
         visited_modules = set()
 
         def collect_modules(parent, depth: int = 0, prefix: str = "") -> None:
--- a/tests/modules/test_moduleconf.py
+++ b/tests/modules/test_moduleconf.py
@@ -1,0 +1,89 @@
+from unittest import mock
+
+from abstract import ViURTestCase
+
+
+class FakeSkel(dict):
+    """Minimal stand-in for the skeleton returned by ``ModuleConf.addSkel()``."""
+
+    def __init__(self, sink: list):
+        super().__init__()
+        self._sink = sink
+
+    def write(self, *args, **kwargs):
+        self._sink.append(self["name"])
+
+
+class TestReadAllModules(ViURTestCase):
+    """Regression tests for :meth:`ModuleConf.read_all_modules`."""
+
+    def setUp(self):
+        super().setUp()
+        from viur.core.modules.moduleconf import ModuleConf
+        ModuleConf.MODULES.clear()
+
+    def _collect_written(self, in_db: list[str], in_code: list[str]) -> list[str]:
+        """Run ``read_all_modules()`` against a faked datastore and module tree.
+
+        :param in_db: module names that already have a ``viur-module-conf`` entry.
+        :param in_code: module names the application exposes below ``vi``.
+        :return: the module names a new entry was written for.
+        """
+        from viur.core import Module, conf
+        from viur.core.modules.moduleconf import ModuleConf
+
+        written = []
+
+        class Vi:
+            pass
+
+        vi = Vi()
+        for name in in_code:
+            setattr(vi, name, Module(name, name))
+
+        # not a Module instance, so collect_modules() skips it -- just like at runtime
+        vi._moduleconf = mock.Mock()
+        vi._moduleconf.addSkel = lambda: FakeSkel(written)
+
+        main_app = mock.Mock()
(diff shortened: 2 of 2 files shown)


sveneberth:
Two follow-up findings from tracing this in production. Both make the case for the fix stronger, and the second one may deserve its own issue.

## Why this stayed invisible for two years — and when it stops being invisible

The bogus write is real but usually harmless, because a skeleton only serializes bones that are in `accessedValues`:

```
frisches Skeleton               : []
nach skel['name'] = ...         : ['name']
nach reinem Lesen ['help_text'] : ['help_text', 'name']
```

`read_all_modules` builds a fresh `addSkel()` and touches only `key` and `name`, so only those two are written. Every other property on the existing entity survives. The entries get rewritten, `changedate` moves, and nothing is lost.

It turns destructive the moment an application overrides `ModuleConfSkel.write()` and **reads** one of its own bones in there — which is the natural thing to do when you want to react to a value changing:

```python
@classmethod
def write(cls, skel, **kwargs):
    new_value = skel["my_bone"] or {}   # <- plain read, but now in accessedValues
    ...
```

From that point on `my_bone` is serialized on every write, and on a blank `addSkel()` that means its empty default. The bogus writes start destroying data.

Field data from a deployment that hit this: a module added in March triggered the chain across 72 entries with all values intact. The same chain in August — after a `write()` override that reads a bone had been added in June — wiped the application's data on 41 entries. Same bug, same mechanism, different outcome; the only thing that changed was ordinary application code.

So the exposure isn't limited to whoever notices a rewritten `changedate`. Any application subclassing `ModuleConfSkel` — which is what the class is for — can turn this into silent data loss without touching anything in core.

## Side finding: the module scan invokes property getters

`collect_modules` does `getattr(parent, module_name, None)` over everything `dir()` yields. That evaluates properties. In the case we traced, a payment provider module exposes a `private_key` property that fetches from Secret Manager; the secret was missing, the getter raised, and the whole startup task died:

```
tasks.py in runStartupTasks
moduleconf.py in read_all_modules      -> collect_modules(conf.main_app.vi)
moduleconf.py in collect_modules       -> recursion
moduleconf.py in collect_modules       -> getattr(parent, module_name, None)
<payment provider>.private_key
viur/core/secret.py in get
-> 404 Secret not found
```

Two consequences worth considering independently of this PR:

- the scan performs network calls it does not need — it only wants to know which attributes are `Module` instances
- a single misconfigured module aborts `read_all_modules` entirely, and since the deferred handler turns the exception into a `408` the queue retries straight back into the same exception

Restricting the traversal so it does not evaluate arbitrary descriptors would avoid both. Split out into #1763, since it is a separate bug from the one this PR fixes.

ArneGudermann (approved):

PR #1766, opened 2026-09-01
feat: create security keys in batches via create(amount=...)

ArneGudermann:
## What changed

`securitykey.create()` gained an `amount` parameter. Instead of one `db.put()` per key it
builds all entities first and writes them in a single `db.put(entities)`.

- `amount` is capped at 500.
- `render.json.skey()` uses the batched call instead of its
  `[securitykey.create() for _ in range(amount)]` loop.

| keys | N single puts | 1 batched put | speedup | ms/key single | ms/key batched |
|-----:|--------------:|--------------:|--------:|--------------:|---------------:|
| 1 | 68 ms | 60 ms | 1.1x | 67.51 | 60.06 |
| 10 | 426 ms | 46 ms | 9.3x | 42.62 | 4.60 |
| 100 | 4171 ms | 84 ms | 49.6x | 41.71 | 0.84 |
| 200 | 8324 ms | 115 ms | 72.6x | 41.62 | 0.57 |
| 300 | 12589 ms | 143 ms | 88.0x | 41.96 | 0.48 |

The single-put path costs a near-constant ~42 ms per key - that is pure round-trip latency,
and it scales linearly. The batched path stays in one commit, so its per-key cost falls to
0.5 ms at 300 keys. A batch of 100 keys drops from 4.2 s to 84 ms.


phorward (approved):

src/viur/core/securitykey.py
```diff
@@ -50,36 +51,41 @@ def create(
...
+
+        entity["viur_until"] = utils.utcNow() + utils.parse.timedelta(duration)
+
+        if not indexed:
+            entity.exclude_from_indexes = [k for k in entity.keys() if not k.startswith("viur_")]
+        entities.append(entity)
+        keys.append(key)
+    db.put(entities)
+    if amount > 1:
+        return keys
```
phorward: Optimization for keeping only one list of entities, and extract the keys from this?

```suggestion
    entities = []
    for i in range(amount):
        key = utils.string.random(key_length)
        entity = db.Entity(db.Key(SECURITYKEY_KINDNAME, key))
        entity |= custom_data
        if session_bound:
            session = current.session.get()
            if not session.loaded:
                session.reset()
            entity["viur_session"] = session.cookie_key

        else:
            entity["viur_session"] = None

        entity["viur_until"] = utils.utcNow() + utils.parse.timedelta(duration)

        if not indexed:
            entity.exclude_from_indexes = [k for k in entity.keys() if not k.startswith("viur_")]
        entities.append(entity)

    db.put(entities)

    if amount > 1:
        return tuple(entity.key.id_or_name for entity in entities)
```

src/viur/core/securitykey.py
```diff
@@ -50,36 +51,41 @@ def create(
         :param indexed: Indexes all values stored with the security-key (default), set False to not index.
         :param key_length: Allows to modify the length of the generated randomized key
         :param custom_data: Any other data is stored with the CSRF-token, for later re-use.
+        :param amount: The amount of CSRF-tokens to generate.
 
         :returns: The new one-time key, which is a randomized string.
```
phorward:
```suggestion
        :returns: The new one-time key. This is always a randomized string. \
            In case amount > 1 is returned, it will be a tuple of strings with the keys.
```

src/viur/core/securitykey.py
```diff
@@ -50,36 +51,41 @@ def create(
...
     """
     if any(k.startswith("viur_") for k in custom_data):
         raise ValueError("custom_data keys with a 'viur_'-prefix are reserved.")
-
+    if amount < 1 or amount > 500:
```
phorward, line 61: This is partly a breaking change... hopefully nobody every used `securitykey.create` with custom_data "amount"...

src/viur/core/securitykey.py
```diff
@@ -37,8 +37,9 @@ def create(
         session_bound: bool = True,
         key_length: int = 13,
         indexed: bool = True,
+        amount: int = 1,
         **custom_data,
-) -> str:
+) -> str | list[str]:
```
phorward:
```suggestion
) -> str | tuple[str]:
```

phorward (changes requested):

phorward (approved):

PR #1768, opened 2026-09-01
perf: `RelationalBone.postSavedHandler` batched writes

ArneGudermann:
## db.get splits oversized lookups

A `Lookup` accepts at most 1,000 keys. `db.get()` now asks in chunks of
`MAX_LOOKUP_KEYS` and reassembles the result in the original key order, so a caller
handing it more keys than that gets an answer instead of a rejected request.

`put()` and `delete()` deliberately keep sending everything in **one** commit. A commit
has no comparable mutation cap - measured against the real datastore, 10,000 entities go
through in a single commit (2.4 s) - so splitting it would give up atomicity without
buying anything. That is stated in both docstrings now.

## postSavedHandler writes once

`RelationalBone.postSavedHandler` did one `db.put()` per relation and one `db.delete()`
per removed relation, on every write of a skeleton carrying a relational bone. The
entities are now collected and written in a single commit each.

Two things this buys, beyond the round-trips: all `viur-relations` entities of an entry
share its entity group (`parent=key`), so N separate puts are N writes against the
one-write-per-second-and-entity-group rate limit, where one commit counts as one. And
the relation update is now atomic, which the loop never was.

## Numbers

Writing N entities, one put each vs. one batched put, measured through two endpoints that
time only the write (median of 3, real datastore):

| entities | one put each | one batched put | speedup |
|---------:|-------------:|----------------:|--------:|
| 1 | 79 ms | 72 ms | 1.1x |
| 10 | 525 ms | 86 ms | 6.1x |
| 100 | 5028 ms | 471 ms | 10.7x |
| 500 | 24933 ms | 2299 ms | 10.8x |

Single puts cost a near-constant ~50 ms per entity: pure round-trip latency, scaling
linearly. The same comparison inside `db.run_in_transaction` comes out at 1.0x across the
board, which is expected - inside a transaction `put_multi` only appends to the running
transaction and the commit happens once at the end. That also settles it for the
transactional path: there is nothing to gain and nothing to lose there.


phorward (approved):

sveneberth:
@ArneGudermann 

> The same comparison inside `db.run_in_transaction` comes out at 1.0x across the
> board,

Compared to what? The current state, or a `put_multi` outside a transaction?

`postSavedHandler` is always executed within a transaction — that is on-batch. I don't quite see yet in which practical scenarios individual operations would actually be executed.

ArneGudermann:
> @ArneGudermann
> 
> > The same comparison inside `db.run_in_transaction` comes out at 1.0x across the
> > board,
> 
> Compared to what? The current state, or a `put_multi` outside a transaction?
> 
> `postSavedHandler` is always executed within a transaction — that is on-batch. I don't quite see yet in which practical scenarios individual operations would actually be executed.

https://github.com/viur-framework/viur-core/blob/a7927d0dd3419ea2683abff9f5d065c163573704/src/viur/core/skeleton/skeleton.py#L614-L623 

The `postSavedHandler` call is outside of the transaction

sveneberth:
> The `postSavedHandler` call is outside of the transaction

Oh, I guess I missed an indent there...

sveneberth (approved):

PR #1770, opened 2026-09-07
fix: known bugs quick fixes

ArneGudermann:
Fixes defects collected in `docs/known_bugs.md` while writing the ADRs. All of them are
cases where the code does not do what it says — a type check against a class instead of
an instance, an assignment to an attribute that does not exist, a missing comma. No
intended behaviour changes.

- **`i18n.py`** — the translation migration marked `entity["translation"]`, singular. No
  bone reads that name, and without the `_viurLanguageWrapper_` marker on `translations`
  `unserialize` collapses `{"de": "Hallo", "en": "Hello"}` into
  `{"en": "{'de': 'Hallo', 'en': 'Hello'}"}`. `translations_missing` is computed on write,
  so that was serialized back — the migration destroyed the data it exists to migrate.
  `DatastoreSource.load` also guarded with `entity["translations"]`, a `KeyError` that
  `initializeTranslations` re-raises, taking the instance down at startup.
- **`bones/uri.py`** — `accepted_protocols="https"` became `set("https")`, allowing h, t,
  p and s; the wildcard was then tested against the original argument, so `"http*"`
  accepted `file://etc/passwd`. And `urlparse(...).port` is `None` without an explicit
  port, so `accepted_ports=(443,)` rejected `https://example.com`; a malformed port raised
  `ValueError` uncaught.
- **`bones/spatial.py`** — `buildDBFilter` assigned to `dbFilter.datastoreQuery`, which
  does not exist, so an invalid geo filter ran the query *without* the spatial constraint
  and returned everything instead of nothing.
- **`email.py`** — the dev-server guard compared an instance against a class, so mail went
  to the App Engine Mail API locally; the quota check tested the deprecated
  `EmailTransportSendInBlue`, so Brevo projects never got a credit warning.
- **`securityheaders.py`** — `extendCsp` assumed `content_security_policy` is set; the
  error page calls it, turning an error response into a second error.
- **`bones/record.py`** — the write path lacked the `None` guard `getSearchTags` has.
- **`tasks.py`** — `send_email` takes `tpl` xor `stringTemplate` and `retry_n_times` passed
  both, silently losing the "task failed permanently" mail.
- **`bones/uid.py`** — `db.CollisionError` is gone since #1431, so the retry never ran and
  real errors surfaced as `AttributeError`. Conflicts belong to `db.run_in_transaction`.
- Smaller: `bones/base.py` (a single `ReadFromClientError` was dropped),
  `bones/randomslice.py` (`raise NotImplemented`), `bones/numeric.py` (missing comma merged
  the translation key into the default text), `prototypes/skelmodule.py` (docstring).

Two changes go slightly beyond the above, both noted in their commits: `UriBone` raises
`ValueError` for a non-iterable `accepted_protocols` instead of the `TypeError` from
`set(5)`, and a scheme with no known default port stays a rejection — which is what the
old code did by testing `None` against the accepted ranges.

### DIFF
--- a/docs/known_bugs.md
+++ b/docs/known_bugs.md
@@ -1,60 +1,14 @@
 # Known bugs
 
 Found while reading the code for the seam documentation in `docs/adr/`, at tag
-`v3.8.33`. Nothing in this list has been fixed. Line numbers refer to that tag.
+`v3.8.33`. Line numbers refer to that tag. Everything listed here is still
+open; fixed entries are removed from this file.
 
 The first pass covered the framework seams (skeleton, module, tasks, email,
 file module, ...), the second pass every bone type under
 `src/viur/core/bones/`.
 
 Each entry: what is wrong, what it costs, what the fix would be.
-
-## Broken comparisons and type checks
-
-### `src/viur/core/email.py:403` - dev-server guard never triggers
-
-```python
-if not conf.email.send_from_local_development_server or transport_class is EmailTransportAppengine:
-```
-
-`transport_class` is an *instance* (`conf.email.transport_class`), so
-`is EmailTransportAppengine` is always False. The intent - never deliver
-through the App Engine Mail API from a local development server - is not
-enforced: with `send_from_local_development_server = True` the call goes
-through and fails inside the API instead.
-
-Fix: `isinstance(transport_class, EmailTransportAppengine)`.
-
-### `src/viur/core/email.py:579` - Brevo quota check never runs
-
-```python
-if not isinstance(conf.email.transport_class, EmailTransportSendInBlue):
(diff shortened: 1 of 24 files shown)


docs/known_bugs.md
```diff
@@ -1,61 +1,15 @@
 # Known bugs
 
 Found while reading the code for the seam documentation in `docs/adr/`, at tag
-`v3.8.33`. Nothing in this list has been fixed. Line numbers refer to that tag.
+`v3.8.33`. Line numbers refer to that tag. Everything listed here is still
```
phorward: Can we update this to 3.9.0.rc1?

phorward (approved):

PR #1772, opened 2026-09-08
fix: resolve further known bugs

ArneGudermann:
- **`DateBone`** — `"1.5"`, `"1-2"`, `"12-"`, `"1.2-3"` raised an uncaught
  ValueError, an HTTP 500 on client input.
- **`BooleanBone.refresh()`** — crashed on a multi-language bone whose value is
  still None, and wrote the default *dict* into every language.
- **`BooleanBone.setBoneValue()`** — now honours `conf.bone_boolean_str2true`
  like every other path in the bone.
- **`StringBone`** — DIN 5007-2 folds `ß`, not just `ẞ`.
- **`UidBone`** — pads with `"0"` instead of `"*"`, and no longer counts the
  wildcard towards the prefix length.
- **`RecordBone`** — raises ValueError instead of TypeError for a missing
  `using`; drops the dead `getSearchDocumentFields`.
- **`RandomSliceBone.buildDBSort()`** — matches the base signature again.
- **`SpatialBone.setBoneValue()`** — returns the documented bool.
- **`render/json`, `skeleton`** — `__all__` holds names, so `import *` works.
- **`File.write()`** — the `weak` flag was inverted: a file in a folder got no
  blob lock and could be collected by the blob GC, a repository-less file
  stayed locked forever.
- **Blob GC** — stopped on the first already-marked blob, so cleanup
  progressed one blob per periodic call.
- **`File.create_src_set()`** — called `cls.get()`, which does not exist, for
  every FileBone with `languages`.


src/viur/core/bones/date.py
```diff
@@ -77,7 +78,9 @@ def __init__(
             if self.multiple:
                 raise ValueError("Cannot be multiple and have a creation/update-magic set!")
 
-            self.readonly = True  # todo: why???
+            # Deprecated and unused: the attribute name does not match BaseBone.readOnly, so this
+            # never locked the bone. Left as-is on purpose -- use "compute" instead of the magic.
+            self.readonly = True
```
phorward: Remove this useless shit, then.
```suggestion
```

src/viur/core/bones/randomslice.py
```diff
@@ -87,45 +89,45 @@ def applyFilterHook(dbfilter, property, value):
...
                 # In each case we kill the query here - making it to return no results
                 raise RuntimeError()
             return property, value
 
-        if "orderby" in rawFilter and rawFilter["orderby"] == name:
+        if "orderby" in params and params["orderby"] == name:
             # We select a random set of elements from that collection
-            assert not isinstance(dbFilter.queries,
+            assert not isinstance(query.queries,
                                   list), "Orderby random is not possible on a query that already uses an IN-filter!"
```
phorward:
```suggestion
            assert not isinstance(query.queries, list), \
                "orderby=random is not possible on a query that already uses an IN-filter!"
```

phorward (changes requested):

phorward (approved):

PR #1774, opened 2026-09-08
fix(utils): `Decimal` support in `viur.core.utils.json`

skoegl:
## What

`viur.core.utils.json` is the encoder/decoder behind `JsonBone` and the deferred task
payloads in `tasks.py`. It knows `db.Key`, `db.Entity`, `datetime`, `timedelta`, `bytes`
and `set`, but not `decimal.Decimal`. On `develop`:

    >>> utils.json.dumps(Decimal("1.5"))
    TypeError: Object of type Decimal is not JSON serializable

With `NumericBone(decimal=True)` in 3.9 that hits any `Decimal` stored in a `JsonBone`
or passed to a `@CallDeferred` function.

Four commits:

1. **`Decimal` round-trips exactly.** Encoded as `{".__decimal__": "12.30"}` via `str()`,
   decoded with the `Decimal` constructor. `str()` instead of `float` so `0.1` stays `0.1`
   and trailing zeros survive. Same marker-dict scheme as the other types.

2. **Falsy marker values were not decoded.** `_decode_object_hook` used walrus assignments
   in its `elif` chain and therefore tested the marker *value* for truthiness. `b""`
   encodes to `""`, `timedelta(0)` to `0.0`, `set()` to `[]`; all three fell through
   every branch and came back as the raw marker dict, e.g. `{".__timedelta__": 0.0}`
   instead of `timedelta(0)`. Membership checks on the marker key fix this for the
   existing types and for the new one.

3. **`Security.cors_origins` docstring** contains `\d`, which Python 3.12+ reports as
   `SyntaxWarning: invalid escape sequence` on every import of `viur.core.config`. Raw string.

4. Tests for the special values (below).

## Edge cases checked

All of these round-trip to an equal `Decimal` with identical `str()`: `-0`, `0E-10`,
`1E-30`, `1E+2`, a 40-digit value, `Infinity`, `-Infinity`, `NaN`, `sNaN` (stays signalling).
Decoding does not round to the active `decimal` context: the constructor is exact,
`localcontext(prec=3)` still yields `Decimal("1.23456789")`. Covered by
`TestViURJsonDecimalSpecialValues`.

## Behaviour change to be aware of

A marker dict whose value is malformed **and falsy**, e.g. `{".__timedelta__": null}` or
`{".__key__": ""}`, used to pass through `loads()` unchanged because the truthiness test
skipped it. It now raises (`TypeError` / `ValueError`), exactly as malformed non-falsy
values such as `{".__key__": "garbage"}` always did. The encoder never produces such
values. The only path where external input reaches `loads()` is
`JsonBone.singleValueFromClient`, which catches `JSONDecodeError` only; a client can
already trigger an exception there with marker syntax on `develop`, this PR does not
introduce that, it only removes the falsy exception to the rule.

## Compatibility

Additive on the wire: one new marker key, nothing existing changes shape. Payloads
written by older cores decode as before. Payloads containing `.__decimal__` read by an
**older** core come back as the marker dict, the same behaviour older cores have for any
unknown marker. Where this matters in practice: a `JsonBone` value or a deferred task
enqueued by an instance on this version and read by an instance still on the previous
version during a rolling deploy.

`render.json.default.CustomJsonEncoder` is untouched. It already renders `Decimal` as
`str` for the JSON output and has no decoder side.

## Tests

`tests/test_utils_json.py`, 10 tests: exact round-trip, wire format, nested structures,
the existing types, the falsy regression cases (`b""`, `timedelta(0)`, `set()`,
`Decimal("0")`), special values, context precision, and a plain single-key dict that must
stay untouched. `python tests/main.py`: 616 tests, the 27 errors are the environment
errors without a datastore emulator and identical to `develop`. `pycodestyle --diff`
with the `tox.ini` settings is clean.

## Context

Found while removing a project-local monkeypatch of `ViURJsonEncoder.default` in a
ViUR 3.9.0rc2 app; that patch goes away once this is released.


phorward:
Relates to #1666

phorward (approved):

phorward (changes requested):
The test suite is not working on GitHub Actions.
Please fix.

phorward (approved):

PR #1780, opened 2026-09-15
feat: `ResponseCache(timezone_sensitive=...)` caches per guessed timezone

theVAX:
## Why

`DateBone.guessTimeZone()` derives the timezone per request from the `X-Appengine-Country` header. The `ResponseCache` key has no timezone dimension. A cached page that renders localized dates therefore shows the timezone of whoever requested it first after expiry - a crawler from another continent, or a health check without the header (UTC) - to every visitor until `max_cache_time` runs out. The uncached detail view renders per request, so the symptom looks like a template bug ("list wrong, detail right").

Typical case: a blog index showing publication dates.

## What

- `viur.core.utils.guess_timezone()`: the request-based guess, extracted unchanged from `DateBone.guessTimeZone()`. The bone keeps its contract (`None` for `naive`, UTC without `localize`) and delegates, so existing overrides keep working.
- `ResponseCache(timezone_sensitive=True)` (and `DEFAULT_SETTINGS.timezone_sensitive`) adds the guessed zone as `__timezone` to the key - one entry per timezone, analogous to `language_sensitive`.
- ADRs `docs/adr/cache.md` and `docs/adr/bones/date.md` document the flag and the alternative for place-bound dates (events, opening hours): override `guessTimeZone()` to pin a zone.
- Second commit, ADRs only: the traps behind this - the first-requester effect and its "list wrong, detail right" symptom, the silent UTC fallbacks of `guess_timezone`, and `compute` functions seeing the localized value.
- Tests: `tests/test_cache.py` (new), `TestGuessTimezone` in `tests/test_utils.py`, `TestDateBone_guessTimeZone` in `tests/bones/test_date_bone.py`.

## Notes

- Behavior without the flag is unchanged; the extraction only moves code. The bare `except:` became `except Exception:` (pycodestyle E722 on the diff).
- Suite: 682 tests green locally (668 before).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


sveneberth:
In my opinion, this is a very specific, individual case. In the age of reactive JavaScript frameworks, it also seems strange to me to cache hard-coded local times. Cache the timestamp in ISO format and use JavaScript to localize — this has the advantage of accounting not only for the time zone but also for the user's format.

Otherwise, this can also be easily solved at the project level using the `conf.cache_environment_key`:
```py
def cache_environment() -> dict | BypassCache:
    return {
        "timezone": guess_timezone(),
    }

conf.cache_environment_key = cache_environment
```

---

I don't see this as a necessary addition to viur-core

